Issue #651, opened 2021-10-27
Fllowstatus bits appear to be cleared when running additional steps


nmoroze:
Strange bug I ran into. Code to reproduce (runs when saved as file in root directory, requires working off  branch noah/allowfail):

```python
import siliconcompiler

netlist = f"tests/data/oh_fifo_sync_freepdk45.vg"
def_file = f"tests/data/oh_fifo_sync.def"

jobname = 'issue_repro'
design = "oh_fifo_sync"

chip = siliconcompiler.Chip()
chip.set('design', design)
chip.set('netlist', netlist)
chip.set('asic', 'def', def_file)
chip.set('mode', 'asic')
chip.set('quiet', True)
chip.set('jobname', jobname)

chip.target('freepdk45')

# no-op import since we're not preprocessing source files
chip.set('flowgraph', 'import', '0', 'function', 'step_join')

chip.set('flowgraph', 'place', '0', 'tool', 'openroad')
chip.set('flowgraph', 'place', '0', 'input', 'import0')

chip.set('flowgraph', 'place', '1', 'tool', 'openroad')
chip.set('flowgraph', 'place', '1', 'input', 'import0')

# Illegal value, so this branch will fail!
chip.set('eda', 'openroad', 'place', '0', 'option', 'place_density', 'asdf')
# Legal value, so this branch should succeed
chip.set('eda', 'openroad', 'place', '1', 'option', 'place_density', '0.5')

# Perform minimum
chip.set('flowgraph', 'placemin', '0', 'function', 'step_minimum')
chip.set('flowgraph', 'placemin', '0', 'input', ['place0', 'place1'])

# ** TOGGLE THIS FLAG TO CONTROL BEHAVIOR **
RUN_CTS = False
if RUN_CTS:
    chip.set('flowgraph', 'cts', '0', 'tool', 'openroad')
    chip.set('flowgraph', 'cts', '0', 'input', ['placemin0'])

chip.run()

print(chip.get('flowstatus', 'place', '0', 'error')) # expect 1
print(chip.get('flowstatus', 'place', '1', 'error')) # expect 0, get 1 when RUN_CTS == True ??
```

When `RUN_CTS` is set to False, those print statements show what I'd expect:
```
1
0
```

since the first place branch fails, while the second one succeeds. However, just by setting `RUN_CTS` to True, they both print 1:
```
1
1
```

nmoroze:
Fixed by #973!

PR #716, opened 2021-11-21
Fix bug where flowstatus error values get lost

nmoroze:
This PR fixes issue #651, where it seemed like flowstatus error values weren't being propagated between steps. It turns out there's a simple fix: I don't think we should be setting 'flowstatus', 'error' for every task to 1 in the schema in `run()`. This prevents these values from being updated by the `read_manifest()` call we make on all input dependencies in step 5 of `_runtask()`.


nmoroze:
Closing in favor of  #973.

PR #780, opened 2021-12-03
Fix ReadTheDocs build

nmoroze:
* Fix .readthedocs.yaml
* Skip leflib build on RTD
* Mock up leflib import/skip sc_floorplan for now
* Fix a sneaky typo in setup.py that caused pip-licenses not to get installed

I want to look into cleaner ways to handle some of these things going forward, but I think this should be good for now!


aolofsson:
Just incredible...

PR #789, opened 2021-12-08
Fix the 'sc-show' command on Windows

WRansohoff:
This fixes a couple of issues that we encountered running 'sc-show' on Windows:

* Windows does not add KLayout's install location to the $PATH environment variable, and the app has a different name from Linux/MacOS.
* If the tool was installed via pip, submodules such as the PDK files are not included. So we shouldn't try to load layer information if it does not exist on the user's system.
* Noah fixed the existing `sc-show` test and added a new one in the [noah/klayout](https://github.com/siliconcompiler/siliconcompiler/commit/72f681e52d868c3e9001dc39608f8e3e3405ecd4) branch, so I pulled that in. We should probably talk about the new test's logic, but I tested the 'no PDK' logic on Windows and it appears to work.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3366,8 +3366,9 @@ class Chip:
             setup_tool(self, mode='show')
 
             exe = self.get('eda', tool, step, index, 'exe')
-            if shutil.which(exe) is None:
-                self.logger.error(f'Executable {exe} not found.')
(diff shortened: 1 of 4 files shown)


aolofsson (approved):

aolofsson:
Nice workaround!

PR #796, opened 2021-12-13
Make KLayout export file read from config directly

nmoroze:
This PR changes the KLayout export script to read directly from the manifest file rather than feeding it arguments through the CLI.

I also added a unit test to exercise creating a GDS from a DEF file and an additional library GDS. In order to check the result of this test, I added a variable to KLayout that disables adding timestamps to the GDS (on by default, which was its previous behavior). We need to be able to disable this to get a deterministic result.


aolofsson:
Great change!

aolofsson (approved):

PR #800, opened 2021-12-16
Bump Surelog Version

nmoroze:
I noticed that the Surelog submodule version accidentally got decreased at some point, so our wheels bundle a version with this bug: https://github.com/chipsalliance/Surelog/issues/2176.

This PR brings Surelog back to a working version, and also adds a regression test for this bug that should prevent this from happening again.


aolofsson:
Great!

PR #802, opened 2021-12-16
Handle duplicate file basenames

nmoroze:
This PR implements the change we discussed to append a hash to collected filenames to make sure they can be unambiguously identified after being placed in a single directory. The actual changes to the core logic are fairly minimal, although this did reveal a few bugs elsewhere that needed fixing.


aolofsson:
Love it! An elegant solution that should serve the project well.

PR #803, opened 2021-12-16
Add basic Chisel support

nmoroze:
This PR adds a Chisel tool setup file with a simple single-file test to demonstrate how it works.


aolofsson:
Awesome. Can you add a test that takes the gcd examplem from scala to gds?

aolofsson:
Wouldn't the basic implementation also have runtime options for source at a minimum?

nmoroze:
>Awesome. Can you add a test that takes the gcd examplem from scala to gds?

Yes, can do! What do you think of adding a 'flowarg' parameter to asicflow and fpgaflow called 'frontend'? Unset/'verilog' would be the current default behavior, 'systemverilog' could deprecate 'flowarg', 'sv', and then we could add more such as 'chisel' as we go. Alternative would probably be to have user manually edit the flowgraph. 

> Wouldn't the basic implementation also have runtime options for source at a minimum?

The default build configuration I included tells the Scala build tool to compile every Scala file in `import/<index>/inputs/`, so no need to handle the source option explicitly!

nmoroze:
Alright, I've implemented a `-frontend` parameter and used it to replace `'flowarg', 'sv'` as well. I changed the chisel test to be an end-to-end test that produces a GDS. 

Only thing to keep in mind is that frontend needs to be set before calling `target()`.

aolofsson (approved):

aolofsson:
This is a great change! The import step had already taken on a special meaning in the flow, the frontend makes it more configurable without needing a lot of custom flows. It's a nice way to reuse the asic/fpga flow with N front ends.

PR #805, opened 2021-12-17
RHEL7 OpenROAD setup script

WRansohoff:
This change adds a setup script for the OpenROAD tools on Red Hat 7.

If the `RHEL7_OPENROAD_INSTALL_DEVTOOLS` environment variable is set, recent versions of a few core tools will be built from source. Otherwise, the script assumes that the user has already installed a GNU toolchain with support for the c++17 standard, recent versions of cmake and boost, and Python 3.6+ with development libraries.


aolofsson:
Awesome!

aolofsson (approved):

PR #807, opened 2021-12-17
Autogenerate list of install scripts in docs

nmoroze:
This PR adds an autogenerated list of our setup scripts to the installation section of the docs. It looks like:

Screen Shot 2021-12-17 at 4 50 56 PM


aolofsson (approved):

aolofsson:
Thanks for doing this. In this case it might have been more efficient to create the list by hand, but it's nice to not have to update docs when we add a new setup script.

PR #817, opened 2021-12-22
Cleaner error message when attempting to get() invalid multi-entry keypaths

nmoroze:
I factored out this small error check from the big file checking PR since it's completely independent.

Before:
```
>>> chip.get('fake', 'keypath')
  File "<stdin>", line 1, in <module>
  File "/home/noah/zeroasic/siliconcompiler/siliconcompiler/core.py", line 757, in get
    return self._search(cfg, keypathstr, *keypath, field=field, mode='get')
  File "/home/noah/zeroasic/siliconcompiler/siliconcompiler/core.py", line 1112, in _search
    cfg[param] = copy.deepcopy(cfg['default'])
KeyError: 'default'
```

After:
```
>>> chip.get('fake', 'keypath')
| ERROR   | job0    | ---          | -   | Get keypath [fake,keypath] does not exist.
```

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1108,8 +1108,12 @@ class Chip:
         #if not leaf cell descend tree
         else:
             ##copying in default tree for dynamic trees
-            if not param in cfg:
+            if not param in cfg and 'default' in cfg:
                 cfg[param] = copy.deepcopy(cfg['default'])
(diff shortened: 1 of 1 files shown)


aolofsson:
Nice cleanup for something that happens to me quite a bit from typys:-)

aolofsson (approved):

PR #823, opened 2022-01-03
Test show() on Windows CI

nmoroze:
This PR adds support for testing `show()` functionality after building Windows wheels on our CI system. 

I verified this with a manual wheel build run: https://github.com/siliconcompiler/siliconcompiler/actions/runs/1616295030, although I've since rebased this to catch up to main.

### DIFF
--- a/siliconcompiler/tools/klayout/klayout.py
+++ b/siliconcompiler/tools/klayout/klayout.py
@@ -50,6 +50,7 @@ def setup_tool(chip, mode="batch"):
         if not shutil.which(klayout_exe):
             loc_dir = os.path.join(Path.home(), 'AppData', 'Roaming', 'KLayout')
             global_dir = os.path.join(os.path.splitdrive(Path.home())[0],
+                                      os.path.sep,
                                       'Program Files (x86)',
                                       'KLayout')
             if os.path.isdir(loc_dir):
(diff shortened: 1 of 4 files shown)


siliconcompiler/tools/klayout/klayout.py
```diff
@@ -50,6 +50,7 @@ def setup_tool(chip, mode="batch"):
         if not shutil.which(klayout_exe):
             loc_dir = os.path.join(Path.home(), 'AppData', 'Roaming', 'KLayout')
             global_dir = os.path.join(os.path.splitdrive(Path.home())[0],
+                                      os.path.sep,
```
WRansohoff, line 53: Nice catch - `sc-show` seems to work on my Windows system without the extra separator, but it looks like this call was creating paths like `C:Program File\(...)`.

WRansohoff (approved):
Looks good to me!

siliconcompiler/tools/klayout/klayout.py
nmoroze: Ah yeah, there was a fun bug here - apparently the drive name without the slash refers to some notion of a CWD stored for that drive in particular ([see this SO post](https://stackoverflow.com/questions/23955968/windows-path-with-no-slash-after-drive-letter-and-colon-what-does-it-point-to)). I'm not sure exactly why, but this caused it to point to the wrong place in the CI environment. There are a bunch of weird things happening there, like Github cloning the repo into a different D:/ drive, changing to temp directories for each test, etc. Something there was enough to uncover this :)

aolofsson:
Great!

PR #824, opened 2022-01-03
Implement 'eda', ..., 'path' parameter

nmoroze:
This PR implements the 'eda', ..., 'path' parameter and uses it to refactor the $PATH hacks we added for Surelog and KLayout. This seems like a slightly cleaner way to handle these hacks: for example, now the user can check the manifest to see where the executables are being found. 

I tested that the changes to these hacks work on all platforms by rebasing this PR on PR #823 and manually running a wheel build/test: https://github.com/siliconcompiler/siliconcompiler/actions/runs/1650785417.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3068,11 +3068,10 @@ class Chip:
 
         vercheck = self.get('vercheck')
         veropt = self.get('eda', tool, 'vswitch')
-        exe = self.get('eda', tool, 'exe')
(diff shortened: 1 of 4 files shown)


WRansohoff (approved):
Seems like a good solution; it's definitely nice to move the workaround out of the `Chip.run` method.

aolofsson:
Good fix! It's great that we have multi platform testing. It's so easy to fall into the trap of relying on unix filepath structures.

PR #825, opened 2022-01-03
Implement automatic record keeping

nmoroze:
This PR implements the `_make_record()` function.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,6 +15,8 @@ pandas >= 1.1.5
 Jinja2 >= 2.11.3
 cryptography >= 3.4.7
 graphviz >= 0.18.1
+netifaces >= 0.11.0
+distro >= 1.6.0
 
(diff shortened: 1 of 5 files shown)


aolofsson:
I think we need to record the sc version for potential 'replay' option later. Otherwise how could we reproduce a result from a json file? Unless we can guarantee that the task sc version is identical to the main version?

nmoroze:
That's a good point, I hadn't considered that different nodes might be running different SC versions for the same job. That situation is a little odd though, since we write the version into the schema in the chip `__init__` function. That means the manifests written to the output directory of each task will already have the SC version used for that task written into the main SC version field. However, I think that won't get picked up on merge, so I think it still makes sense to record it separately in the record section so we can see all versions used from the final manifest.

aolofsson (approved):

PR #827, opened 2022-01-04
Automate most submodule management

nmoroze:
This PR might seem a little random, but it was a refactor I started before the holidays related to a test I was thinking of adding then scrapped. I kind of like this cleanup though, so figured I'd suggest it anyways. Long description from commit message:

This automates clones of design/tool submodules by coupling them to the
places where the submodules are actually needed. This prevents us from
having to do a fine-grained selection of which submodules are needed in
CI scripts, while still avoiding a complete download of our many
submodules.

The most intricate part are the new test fixtures: these will clone
design submodules as needed by tests and return a path to the submodule.
If the submodule is already cloned there will be no effect, and the
fixture will not attempt to clone the submodule more than once per test
session, since the 'session' scope is set.


aolofsson:
Looks good.

aolofsson (approved):

PR #828, opened 2022-01-04
Add support for -I flag

### LINKED ISSUE #632: -I flag needs to be implemented
I was trying to run a small CLI experiment a couple days ago and noticed that the -I flag didn't seem to be recognized:

```
sc ibex_src/ibex_if_stage.sv -I /home/noah/zeroasic/zerosoc/opentitan/hw/dv/sv/dv_utils  -y ibex_src \ 
   -design ibex_if_stage -target surelog_freepdk45 -arg_step import -mode asic
```

gave

```
sc: error: unrecognized arguments: -I /home/noah/zeroasic/zerosoc/opentitan/hw/dv/sv/dv_utils
```

-I is meant to be one of the special flags we provide for compiler compatibility, right? I think that special functionality might still need to be added.

nmoroze:
This PR adds supports for the -I flag. 

I actually think we don't yet have any parameters with two switches (example from discussion earlier was `ydir`, but that only supports `-y`, not `-ydir`). This might be a bit much, but I decided to mock up a solution that makes the switch parameter an optional list. I'm not sure if this is worth it for the small, finite number of cases where we might want to do this, but the code change ended up being pretty minimal so I'm curious to get thoughts on this approach. 

Closes #632

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -193,30 +193,42 @@ class Chip:
             typestr = self.get(*key, field='type')
             #Switch field fully describes switch format
             switch = self.get(*key, field='switch')
-            if switch is not None:
+            if switch is None:
+                switches = []
+            elif isinstance(switch, list):
+                switches = switch
+            else:
+                switches = [switch]
+
+            switchstrs = []
+            dest = None
+            for switch in switches:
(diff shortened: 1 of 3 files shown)


aolofsson:
Awesome fix! I didn't it was that easy to support mutliple option strings in argparse. I should have read the docs more carefully!

aolofsson (approved):

PR #829, opened 2022-01-04
Dynamically check that all required file/dir-type parameters resolve correctly in check_manifest()

nmoroze:
This check ensures that as long as the tool spec is written correctly, we should never fail during TCL execution due to a filepath being resolved to None.


aolofsson (approved):

aolofsson:
Very nice!

PR #832, opened 2022-01-05
KLayout testing/wheel packaging enhancements

nmoroze:
This PR implements a few enhancements/fixes to sc-show and our wheel builds.

- Fix chip.show() test to use GDS for pdk-less test (using DEF fails on newer KLayout version)
- Adds auto path discovery for KLayout macOS install
- Clean up wheel package script and test chip.show() on all platforms after wheels built
- Fix bug I introduced in #827 that broke Surelog build caching


aolofsson (approved):

.github/workflows/bin/setup_wheel_env_linux.sh
```diff
@@ -1,7 +1,12 @@
 #!/bin/bash
 
 # Install dependencies
-yum install -y libuuid-devel zlib-devel java-1.8.0-openjdk-devel graphviz
+yum install -y libuuid-devel zlib-devel java-1.8.0-openjdk-devel graphviz xorg-x11-server-Xvfb wget
```
mithro, line 4: FWIW - You might want to call this script something like `setup_wheel_env_centos.sh` or `setup_wheel_env_linux_centos.sh` incase you ever end up also testing Debian / Ubuntu or other Linux versions.

PR #834, opened 2022-01-07
Implement 'env' schema parameter

nmoroze:
This PR adds functionality for 'env' parameter + a test for it.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3350,6 +3350,11 @@ class Chip:
             if os.path.isdir(cur_job_dir):
                 shutil.rmtree(cur_job_dir)
 
+        # Set env variables
+        for envvar in self.getkeys('env'):
+            val = self.get('env', envvar)
+            os.environ[envvar] = val
+
         # Remote workflow: Dispatch the Chip to a remote server for processing.
         if self.get('remote'):
             # Load the remote storage config into the status dictionary.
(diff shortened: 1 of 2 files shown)


aolofsson:
Not sure this is the right place for it? Shouldn't it be in the _runtask so that it gets set on the compute node?

nmoroze:
I thought a bit about this, I think this works since compute nodes will re-enter SC through run(), and the parent environment will get propagated into the child processes that run _runtask() (the test verifies this). 

Benefit of putting this in run() is that it calls check_manifest(), which could potentially resolve on env vars being set to resolve file paths in some of the checks.

nmoroze:
> I thought a bit about this, I think this works since compute nodes will re-enter SC through run(), and the parent environment will get propagated into the child processes that run _runtask() (the test verifies this).

Although I could be wrong about this if there's some sort of more complex setup I'm not foreseeing. cc @WRansohoff, does this seem alright to you?

WRansohoff:
This seems like a reasonable place for it; the `Chip.run()` method still gets called on the compute nodes. They determine which task to run from the `-arg_step`/`-arg_index` parameters, and load their `active`/`error` flags from the parent's configuration.

My only comment is that the 'env' parameter shouldn't be allowed through the remote schema sanitizer; we don't want to accidentally overwrite a $PATH variable or something like that.

aolofsson:
You are both right. The re-entrant sc got me again.:-)

aolofsson (approved):

PR #838, opened 2022-01-12
Remove 'copy' flag from 'pdk', 'doc' parameters

nmoroze:
I think this was inadvertent, but 'copy' should be false here since there's no reason to copy PDK docs into the import directory, and it adds clutter/makes the build directory bigger than necessary.


aolofsson (approved):

aolofsson:
Nice catch!

PR #843, opened 2022-01-12
Update the `-read_def/gds` sc-show options to work with `-cfg`

WRansohoff:
Our recent schema changes may have introduced a small bug in how `sc-show` parses its command-line options. I started following [the ZeroSoC tutorial](https://docs.siliconcompiler.com/en/latest/tutorials/zerosoc.html) to get a feel for the Python floorplanning API, and I got the following error when I tried to view the first .def file:

    $ sc-show -read_def "show 0 asic_core.def" -cfg core_manifest.json
    [...]
    | ERROR   | job0    | ---          | -   | Exactly one of -read_gds, -read_def, or -design must be provided.

It looks like the `sc-show` app expects that the `-design` parameter could only have been set from command line arguments [at the point where this error is thrown](https://github.com/siliconcompiler/siliconcompiler/blob/ced8e2b35127ae8f22efa2d8499fd451d5a13a51/siliconcompiler/apps/sc_show.py#L35). But if the `-cfg` parameter is passed in, the `Chip.create_cmdline(...)` method will call `Chip.read_manifest(...)` and set the `-design` parameter early.

This PR gives the `-read_def` and `-read_gds` CLI options priority over `-cfg`, and updates the `sc-show` CI test to catch this in the future.

### DIFF
--- a/siliconcompiler/apps/sc_show.py
+++ b/siliconcompiler/apps/sc_show.py
@@ -32,8 +32,11 @@ def main():
     gds_mode = chip.valid('read', 'gds', 'show', '0') and bool(chip.get('read', 'gds', 'show', '0'))
     def_mode = chip.valid('read', 'def', 'show', '0') and bool(chip.get('read', 'def', 'show', '0'))
 
-    if (def_mode + gds_mode + design) != 1:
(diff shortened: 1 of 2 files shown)


nmoroze (approved):
Nice catch and elegant fix, LGTM. Thanks!

PR #846, opened 2022-01-14
Fix collected filename mangling on FIPS compliant systems

### LINKED ISSUE #845: md5 disabled for fips
On machines where FIPS is enabled, md5 is not allowed in the hashlib. (See error below).  Changing line 3904 from hashlib.md5 to hashlib.sha1 avoids this error completely (but no clue is that is a viable solution). 
https://github.com/siliconcompiler/siliconcompiler/blob/3c32842e31fa01fe6c6d407030eb0dae2dec0736/siliconcompiler/core.py#L3561
If there is a way to check what is enabled that might be a good general solution to avoid this.

```
   sys.exit(load_entry_point('siliconcompiler', 'console_scripts', 'sc')())
  File "siliconcompiler/apps/sc.py", line 64, in main
    chip.run()
  File "siliconcompiler/core.py", line 3443, in run
    self.check_manifest()
  File "siliconcompiler/core.py", line 1633, in check_manifest
    if not self._check_flowgraph_io():
  File "siliconcompiler/core.py", line 1736, in _check_flowgraph_io
    inputs = self._gather_outputs(in_step, in_index)
  File "siliconcompiler/core.py", line 1696, in _gather_outputs
    imports = {self._get_imported_filename(p) for p in self._collect_paths()}
  File "siliconcompiler/core.py", line 1696, in <setcomp>
    imports = {self._get_imported_filename(p) for p in self._collect_paths()}
  File "siliconcompiler/core.py", line 3904, in _get_imported_filename
    pathhash = hashlib.md5(pathstr.encode('utf-8')).hexdigest()
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for fips
```

nmoroze:
This PR adds a wrapper around the hashlib md5 function that adds the `usedforsecurity=False` keyword argument when supported. Since using MD5 to deduplicate filenames isn't security related, this is safe.

Closes #845.

### DIFF
--- a/siliconcompiler/utils.py
+++ b/siliconcompiler/utils.py
@@ -25,19 +25,3 @@ def copytree(src, dst, ignore=[], dirs_exist_ok=False, link=False):
             os.link(srcfile, dstfile)
         else:
             shutil.copy2(srcfile, dstfile)
-
-def insecure_md5(data):
-    '''Wrapper around hashlib md5 function to enable use on FIPS compliant
-    platforms.
-
-    This function SHOULD NOT be used for cryptography/security-related purposes.
-    '''
-    encoded_data = data.encode('utf-8')
-
-    if sys.version_info[0] == 3 and sys.version_info[1] >= 9:
-        # usedforsecurity flag added in Python 3.9
-        return hashlib.md5(encoded_data, usedforsecurity=False).hexdigest()
-    else:
-        # Regular md5 call seems to work on FIPS systems when using Python
-        # versions older than 3.9.
-        return hashlib.md5(encoded_data).hexdigest()
--- a/siliconcompiler/utils.py
+++ b/siliconcompiler/utils.py
@@ -1,5 +1,7 @@
 import os
+import hashlib
 import shutil
+import sys
 
 def copytree(src, dst, ignore=[], dirs_exist_ok=False, link=False):
     '''Simple implementation of shutil.copytree to give us a dirs_exist_ok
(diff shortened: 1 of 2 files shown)


siliconcompiler/utils.py
```diff
@@ -23,3 +25,19 @@ def copytree(src, dst, ignore=[], dirs_exist_ok=False, link=False):
...
+    This function SHOULD NOT be used for cryptography/security-related purposes.
+    '''
+    encoded_data = data.encode('utf-8')
+
+    if sys.version_info[0] == 3 and sys.version_info[1] >= 9:
+        # usedforsecurity flag added in Python 3.9
+        return hashlib.md5(encoded_data, usedforsecurity=False).hexdigest()
+    else:
+        # Regular md5 call seems to work on FIPS systems when using Python
+        # versions older than 3.9.
```
arlpetergadfort, line 42: Thanks for taking a look at this; however, we are running python 3.6.8

nmoroze: Huh, interesting. I tested this out on a FIPS-compliant Ubuntu 18.04 VM with Pythons 3.6-3.10 and it seemed to work. Passing in this flag on the older Python versions resulted in an exception. 

Could you send me some more info about your environment? Which OS you're running and where your Python installation came from if you know (distro package manager, built from source, etc.) would be helpful. If you'd prefer to follow up via email that's fine too (noah@zeroasic.com). 

Thanks!

nmoroze: Actually, never mind - we'll go with using sha1() instead. No reason for us not to use it here, and it will avoid any pitfalls related to platform/Python version compatibility.

aolofsson (approved):

aolofsson:
Looks good. Less dependancies is better.

PR #847, opened 2022-01-14
Add an initial Dockerfile and setup script

WRansohoff:
This PR adds a simple Dockerfile for an Ubuntu 20.04 image with sc and the core open-source EDA tools installed.

Next steps: X11 forwarding isn't set up for GUI apps like KLayout yet, and it might be a good idea to add instructions for mounting local directories in the container.

### DIFF
--- a/docs/_ext/installgen.py
+++ b/docs/_ext/installgen.py
@@ -16,13 +16,15 @@ class InstallScripts(SphinxDirective):
         scripts = {}
 
         for script in os.listdir(setup_dir):
-            components = script.split('.')[0].split('-')
-            tool = components[1]
+            # Ignore directories such as 'setup/docker/'.
+            if os.path.isfile(script):
+                components = script.split('.')[0].split('-')
+                tool = components[1]
 
-            if tool not in scripts:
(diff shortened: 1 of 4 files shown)


aolofsson:
Very cool. Is this complete? What about yosys?

WRansohoff:
It is complete in the sense that it can run the entire ASIC flow and generate a GDS file, but `sc-show` won't open the design in KLayout because the container isn't configured to forward X11 windows to the host. As a workaround, we can copy the build directory out of the container with a command like:

    docker cp my_sc_container:/siliconcompiler/build/ .

And the OpenROAD-flow-scripts repository includes yosys as a submodule, so that gets installed by `install-openroad.sh`.

aolofsson:
Got it, thanks! I forgot that openroad installs yosys as well. Apoligies for my confusion.:-)

aolofsson (approved):

umarcor:
With regard to X11, I suggest to have a look at [hdl.github.io/containers: Tools with GUI](https://hdl.github.io/containers/ug/GUI.html).

PR #853, opened 2022-01-21
Add 'tool' parameter to floorplan constructor

nmoroze:
Now that we've split up PDK tech information by tool, this parameter lets a designer specify which tool will consume the floorplan, so that the floorplan is generated using the correctly-associated tech information. The default is OpenROAD, maintaining backwards compatibility with the previous API.


aolofsson:
Nice!

aolofsson (approved):

PR #854, opened 2022-01-25
First pass at allowing multiple 'version switch' parameters for tools

WRansohoff:
This change updates the schema's `vswitch` parameter to allow multiple command-line options for version checks.

Some tools (such as KLayout) require more than one CLI flag to produce output without trying to open GUI windows. Like we discussed earlier, `vswitch` is conceptually similar to the generic CLI `option` parameter, so it makes sense to store it as a list of strings.


nmoroze (approved):
LGTM, thanks Will!

PR #857, opened 2022-01-31
Adding grep like regex parameter to schema

aolofsson:
New logfile grep feature: 
-  It's doubtful there has been a digital tapeout done in the last 30 years that didn't rely on some kind of grep check to make sure the log files were clean.
- This feature tries to replicate that grep approach within the schema.
- Instead of keeping little one line shell scripts around in a tar ball, we embed patterns in the schema for safe keeping.
- Support for a selection of grep switches in conjunction with pipe enables generalized and/or/not pattern search.

Also:
 - Enable running complete flow as nop, which is useful when testing new flows and new tools. If the tool is missing in the path, the version check would fail.
 - Changing the write manifest message from info to debug to reduce noise. It wasn't really adding value to the casual user.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -2540,7 +2540,7 @@ class Chip:
             Searches for regex matches in the place logfile.
         '''
 
-        # Using manifest to get defults
+        # Using manifest to get defaults
         if step is None:
             step = self.get('arg', 'step')
         if index is None:
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3070,6 +3070,7 @@ class Chip:
         # Shared parameters (long function!)
         design = self.get('design')
         tool = self.get('flowgraph', step, index, 'tool')
+        quiet = self.get('quiet') and (step not in self.get('bkpt'))
 
         ##################
         # 1. Wait loop
@@ -3286,7 +3287,6 @@ class Chip:
                 # Use separate reader/writer file objects as hack to display
                 # live output in non-blocking way, so we can monitor the
                 # timeout. Based on https://stackoverflow.com/a/18422264.
-                quiet = self.get('quiet') and (step not in self.get('bkpt'))
                 cmd_start_time = time.time()
                 proc = subprocess.Popen(cmdlist,
                                         stdout=log_writer,
@@ -3323,7 +3323,8 @@ class Chip:
 
         ##################
         # 18. Check log file
-        self.check_logfile(display=not quiet)
+        if (tool not in self.builtin) and (not self.get('checkonly')) :
+            self.check_logfile(display=not quiet)
 
         ##################
         # 19. Hash files
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -2459,7 +2459,8 @@ class Chip:
 
 
     ###########################################################################
-    def check_logfile(self, step, logfile=None, jobname='job0', index='0', display=True):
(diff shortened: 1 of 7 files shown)


siliconcompiler/core.py
```diff
@@ -2456,6 +2457,77 @@ def calc_dpw(self):
...
+        if self.valid('eda', tool, 'regex', step, index, 'default'):
+            regex_list = self.getkeys('eda', tool, 'regex', step, index)
+
+        for suffix in regex_list:
+            checks[suffix] = {}
+            checks[suffix]['report'] = open(f"{step}.{suffix}", "w")
+            checks[suffix]['patterns'] = []
+            patterns = self.get('eda', tool, 'regex', step, index, suffix)
+            for item in patterns:
+                negate  = re.search(r'^-v (\w+)', item)
```
WRansohoff, line 2509: Is there any chance that a negation regex could start with a non-word character such as `[]$%().`, etc? It looks like this will match a string like: `"-v error123"`, but not `"-v [error123]"`.

WRansohoff (commented):
Looks good to me!

I just have one comment about the regex search string characters - count this review as an approval if you think the search string is correct.

siliconcompiler/schema.py
```diff
@@ -2077,6 +2080,40 @@ def schema_eda(cfg, tool='default', step='default', index='default'):
...
+        output file <design>.<suffix>. A 'not' operation is supported by prepending
+        the the '-v ' to the regex pattern (like in the Unix grep command). The
+        example below illustrates how to implement a simple grep type
+        filter with the 'regex' paramter:
+
+        GREP:
+        >> grep WARNING syn.log | grep -v "blackbox" > syn.warnings
+
+        SC:
+        chip.set('eda','yosys','regex','syn',0','warnings',['WARNING','-v blackbox'])
```
nmoroze, line 2112: A few typos:

```suggestion
        A list of regular expressions patterns that together emulate a
        single line grep/pipe Unix pattern. The first regex is compared
        to the log file, and any matches are then compared to the second
        patterns, etc. The complete regex comparison is placed in the
        output file <design>.<suffix>. A 'not' operation is supported by prepending
        the '-v ' to the regex pattern (like in the Unix grep command). The
        example below illustrates how to implement a simple grep type
        filter with the 'regex' parameter:

        GREP:
        >> grep WARNING syn.log | grep -v "blackbox" > syn.warnings

        SC:
        chip.set('eda','yosys','regex','syn','0','warnings',['WARNING','-v blackbox'])
```

nmoroze (commented):
Changes overall also LGTM, +1 to Will's note and marked a couple typos.

Another thing to consider would be writing the results of the filter to an entry in the schema, but we can add this functionality in a future PR.

siliconcompiler/core.py
aolofsson: > Is there any chance that a negation regex could start with a non-word character such as `[]$%().`, etc? It looks like this will match a string like: `"-v error123"`, but not `"-v [error123]"`.

Great catch!

aolofsson: I went back and forth, I didn't like the hackiness of having a "not" operation, but clearly the -v will get us into tourble at some point, and is hacky! A less hacky compromise would be a tuple with (opt,regex), where opt  would be the switch given to grep (a subset). For example , -i, -x, -v, -w.,,,That feels right! Thoughts?

nmoroze: That seems like a good compromise, especially if we plan to extend the functionality with more options. I'd be for that!

PR #858, opened 2022-02-01
Fix steplist bug in check_manifest()

nmoroze:
Steplist defvalue is [], so we shouldn't compare to None to check if it's unset.

This change uncovered a few other hidden bugs which I've fixed:

1. Need to set the steplist in test_check_manifest(): since we're only setting up two steps here, we need to set the steplist accordingly.

2. Setting a steplist before calling the local _runtask() in remote_preprocess() - this one's a bit more complex. From the commit message body:
Another bug hidden by the check_manifest() steplist issue. In this case,
only the local step is set up before remote_preprocess() calls
_runtask(), so we need to override the steplist to force it to only
check the local step. This feels a bit hacky, but it shouldn't be
dangerous since we already overwrite the steplist for the remote run
immediately after.
I also snuck in a separate bugfix - it seems like the _runtask() call
needs to be outside of the if statement, otherwise the local step will
not run if the local step runs a builtin function.

### DIFF
--- a/siliconcompiler/client.py
+++ b/siliconcompiler/client.py
@@ -62,8 +62,13 @@ def remote_preprocess(chip):
             chip.set('arg', 'index', index)
             func = chip.find_function(tool, 'tool', 'setup_tool')
             func(chip)
-            # Run the actual import step locally.
-            chip._runtask(local_step, index, active, error)
+
+        # Need to override steplist here to make sure check_manifest() doesn't
+        # check steps that haven't been setup.
+        chip.set('steplist', local_step)
+
+        # Run the actual import step locally.
+        chip._runtask(local_step, index, active, error)
 
     # Set 'steplist' to only the remote steps, for the future server-side run.
     remote_steplist = remote_steplist[1:]
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1582,7 +1582,7 @@ class Chip:
         '''
 
         steplist = self.get('steplist')
-        if steplist is None:
+        if not steplist:
             steplist = self.list_steps()
 
         #1. Checking that flowgraph is legal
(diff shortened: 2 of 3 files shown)


aolofsson:
Looks good!

siliconcompiler/client.py
```diff
@@ -62,8 +62,13 @@ def remote_preprocess(chip):
...
             func(chip)
-            # Run the actual import step locally.
-            chip._runtask(local_step, index, active, error)
+
+        # Need to override steplist here to make sure check_manifest() doesn't
+        # check steps that haven't been setup.
+        chip.set('steplist', local_step)
+
+        # Run the actual import step locally.
+        chip._runtask(local_step, index, active, error)
```
WRansohoff, line 71: Do these steps need to be indented under 'if tool:' to avoid calling them for builtin functions?

WRansohoff (approved):
Looks okay to me, just double-check to make sure this won't interfere with steps that don't have a tool, like the 'min' tasks in parallel flows.

siliconcompiler/client.py
WRansohoff: Now that I think about it, this may be moot since we're unlikely to have multiple pre-process/import steps. We originally wrote this to possibly support a whole pre-remote flowgraph, but it seems like we're settling on running one 'import' task.

nmoroze: This was actually originally indented under the 'if tool:', but I purposely de-indented them to make sure they'd run regardless. I think even if the tool were a built-in, we would want to run it -- this seemed like a latent bug in the original code.

For example, one-shot tool runs use the 'join' builtin as a no-op import step that collects required files. If we wanted to eventually allow users to just single-step a tool on the server and we didn't call a local _runtask() for the join, I don't think SC would collect the input files and have them packaged up and sent to the server.

WRansohoff: Ah, that makes sense. I think the code used to call `setup_tool` directly, which was why it was like that, but the `_runstep` method should handle everything correctly.

Nice catch!

PR #859, opened 2022-02-03
Copy field fix ups in schema

aolofsson:
- Some of the defaults did not make sense or were wrong.


nmoroze (approved):
Agreed, good catch.

PR #860, opened 2022-02-03
Get rid of hard requirement to provide sources on CLI

nmoroze:
This PR makes it so providing sources is no longer a hard requirement when using the CLI. Instead, individual tools that ingest sources are responsible for marking it as a tool-specific requirement. I also had to add a bit of logic to sc.py to gracefully handle an error that occurs if we try to infer 'design' when no sources are provided. I think the error messages for these invalid invocations are pretty clear:

```
$ sc
| INFO    | job0    | ---          | -   | Setting commandline arguments
| ERROR   | job0    | ---          | -   | Invalid arguments: either specify -design or provide sources.
```

```
$ sc -design gcd
| INFO    | job0    | ---          | -   | Setting commandline arguments
| INFO    | job0    | ---          | -   | Loading target 'asicflow_freepdk45'
| INFO    | job0    | ---          | -   | Loading function 'setup_flow' from module 'asicflow'
| INFO    | job0    | ---          | -   | Loading function 'setup_pdk' from module 'freepdk45'
| INFO    | job0    | ---          | -   | Operating in 'asic' mode
| ERROR   | job0    | ---          | -   | Value empty for [['source']].
| ERROR   | job0    | ---          | -   | Check failed. See previous errors.
```

### DIFF
--- a/siliconcompiler/apps/sc.py
+++ b/siliconcompiler/apps/sc.py
@@ -46,7 +46,13 @@ def main():
 
     # Set design if none specified
     if not chip.get('design'):
-        topfile = chip.get('source')[0]
+        sources = chip.get('source')
+        if len(sources) > 0:
+            topfile = chip.get('source')[0]
+        else:
+            chip.logger.error('Invalid arguments: either specify -design or provide sources.')
+            sys.exit(1)
+
         topmodule = os.path.splitext(os.path.basename(topfile))[0]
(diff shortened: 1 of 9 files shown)


aolofsson:
A very nice and simple solution that helps expand sc beyond the current compiler focus!

aolofsson (approved):

PR #862, opened 2022-02-06
Expanding flowgraph and target schema to manage complexity

aolofsson:
This PR addresses limitations within SC in handling modern complex flows. The previous approach of using a "tuple" for specifying a target was borrowed from the SW world (gcc/llvm), but we are starting to see that the combinations in HW makes it impractical. Going forward, the project will have a single target string (pointing to a Python module), that sets up everything associated with a target. This will generally includes flows, libs, pdks, and any other methodology settings needed.

Another major change is introduced a 'flow' keyword to the flowgraph. In setting up a complex target, there tends to be multiple independent flows. For example, simulation, RTL2GDS,  signoff, etc. These are complex enough and independent and should be set up at target creation time. We could have managed this as independent chip objects runs, but this makes for a looser relationship between the target and the design. In general, modern PDK will have signoff reference flows for a bunch of different tools. These should all be baked into the target for the designer to choose from at run time. The new flowgraph allows for this.

Minor fixups include:

- Changing asic targetlib to asic logiclib for clarity
- Adding units to enable node agnostic constraints
- Changing target to load_target (see above)
- Reorganizing target files (modularity)
- Moving driver into cells dictionary (consistency)
- Driving tracks for asap7 from tcl file (too complex for grid approach)
- Standardizing cell name types (enforce standardization!)


siliconcompiler/core.py
```diff
@@ -492,122 +471,104 @@ def find_function(self, modulename, functype, funcname):
                 self.logger.error(f"Module setup failed for '{modulename}'")
                 self.error = 1
 
-    ###########################################################################
-    def target(self, name=None):
+    ##########################################################################
+    def load_target(self, name):
         """
-        Configures the compilation manifest based on pre-defined target modules.
+        Loads a project module and runs the setup() function.
```
nmoroze, line 477: project->target

siliconcompiler/core.py
```diff
@@ -492,122 +471,104 @@ def find_function(self, modulename, functype, funcname):
...
+            self.logger.error(f'Module {name} not found.')
             sys.exit(1)
 
-        target = self.get('target')
-        self.logger.info(f"Loading target '{target}'")
-
-        # search for module matches
-        targetlist = target.split('_')
-
-        for i, item in enumerate(targetlist):
```
nmoroze, line 549: Great that all this logic is going away, it was getting complicated!

siliconcompiler/core.py
```diff
@@ -2540,18 +2522,17 @@ def check_logfile(self, step=None, jobname=None, index=None,
         '''
 
         # Using manifest to get defaults
-        if step is None:
-            step = self.get('arg', 'step')
-        if index is None:
-            index = self.getkeys('flowgraph', step)[0]
```
nmoroze, line 2546: Any reason for getting rid of this?

Changes
```diff
@@ -0,0 +1,100 @@
+.. for github, vim: syntax=reStructuredText
+
+=================================
+Revision History and Change Log
+=================================
+
+The changes in each SiliconCompiler release version are described below.   Commit
```
nmoroze, line 7: nit: looks like extra spaces between "below." and "Commit"

siliconcompiler/flows/asicflow.py
```diff
@@ -52,14 +52,15 @@ def make_docs():
...
+    chip.set('flow', 'asicflow')
+    setup(chip)
 
     return chip
 
 ###########################################################################
 # Flowgraph Setup
 ############################################################################
-def setup_flow(chip):
+def setup(chip, flowname='asicflow'):
```
nmoroze, line 63: Maybe a change for a future PR, but if we make this `flowname` keyword arg a standard for flow setup() functions, we could let the user optionally override it through a kwarg to `load_flow`. Could be a nifty feature.

nmoroze: As an example use case, I'm imagining a user writing a target that sets up multiple variations of one flow, e.g.:

```python
chip.set('flowarg', 'proc_np', '1')
chip.load_flow('asicflow', name='serial_asicflow')
chip.set('flowarg', 'proc_np', '10')
chip.load_flow('asicflow', name='parallel_asicflow')
```

siliconcompiler/schema.py
```diff
@@ -1808,6 +1807,23 @@ def schema_flowgraph(cfg, step='default', index='default'):
         """
     }
 
+    cfg['flowgraph'][flow][step][index]['flowin'] = {
```
nmoroze, line 1810: should we mark WIP or unsupported somehow? I know we haven't done this for some other unsupported things, but I get a little worried about creating confusion by adding in placeholders to schema.

siliconcompiler/tools/openroad/sc_floorplan.tcl
```diff
@@ -48,18 +48,24 @@ if {[expr ! [dict exists $sc_cfg "read" def $sc_step $sc_index]]} {
...
-	set yoffset [dict get $sc_cfg pdk grid $sc_stackup $metal yoffset]
-
-	make_tracks $metal \
-	    -x_offset $xoffset \
-	    -x_pitch $xpitch \
-	    -y_offset $yoffset \
-	    -y_pitch $ypitch
+    # source tracks from file if found, else else use schema entries
+    if [dict exists $sc_cfg pdk aprtech openroad $sc_stackup $sc_libtype tracks] {
+	puts "READING FILE"
```
nmoroze, line 53: Seems like extra from debug, remove? (or make more descriptive)

nmoroze (commented):
I think I managed a relatively thorough look! Found mostly nitpicks/typos, I may have caught one thing that's a small bug but should be an easy fix. Comments inline.

A few higher-level points:
- I'm not sure how I feel about the default target names ending in `*_demo`. Makes them sound a bit "toy" or incomplete to me, when I figure that users would probably be okay just using them as-is if they're going with open tools (and note they also have to use these with the cloud service). But if the goal is to encourage people to fork these and write their own, then I think it makes sense. Also, sorry for not voicing that thought sooner, this doesn't seem fun to change...
- We need to update [targets page](https://docs.siliconcompiler.com/en/latest/user_guide/targets.html) in docs, and should probably expand the docs generators to cover libs and targets. I can take both on as an action in a future PR if you'd like.

siliconcompiler/core.py
```diff
@@ -394,36 +394,7 @@ def create_cmdline(self, progname, description=None, switchlist=[]):
                             self.set(*args, clobber=True)
 
     #########################################################################
-    def create_env(self):
```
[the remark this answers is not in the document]
nmoroze: If so, we might need to do something about the sc-env app: https://github.com/siliconcompiler/siliconcompiler/blob/main/siliconcompiler/apps/sc_env.py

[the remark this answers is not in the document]
WRansohoff: I think the sc-env app may be defunct; it uses the 'project' schema parameter which no longer exists in schema.py

siliconcompiler/core.py
aolofsson: Yes!

siliconcompiler/core.py
```diff
@@ -438,42 +409,50 @@ def find_function(self, modulename, functype, funcname):
         * flows/modulename.py
         * pdks/modulname.py
 
+        If the modtype is None, the module paths are search in the
```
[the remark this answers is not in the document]
aolofsson: check

siliconcompiler/core.py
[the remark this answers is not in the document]
aolofsson: Yes, defunct or broken.

siliconcompiler/core.py
aolofsson: Adding it back, should be safe (with the flow added)

siliconcompiler/flows/asicflow.py
aolofsson: Interesting idea!

siliconcompiler/schema.py
aolofsson: Removed, you are right!

siliconcompiler/tools/openroad/sc_floorplan.tcl
aolofsson: removed

siliconcompiler/core.py
```diff
@@ -438,42 +409,50 @@ def find_function(self, modulename, functype, funcname):
...
-            fullpath = self._find_sc_file(f"tools/{modulename}/{modulename}.py", missing_ok=True)
-        elif functype == 'flow':
-            fullpath = self._find_sc_file(f"flows/{modulename}.py", missing_ok=True)
-        elif functype == 'pdk':
-            fullpath = self._find_sc_file(f"pdks/{modulename}.py", missing_ok=True)
-        elif functype == 'project':
-            fullpath = self._find_sc_file(f"projects/{modulename}.py", missing_ok=True)
+        # module search path depends on modtype
+        if moduletype is None:
+            for item in ('targets', 'flows', 'pdks', 'libs'):
```
WRansohoff, line 439: Should this also include 'tools' between 'flows' and 'pdks', or should there be a special case for 'tools' like with the statements below?

siliconcompiler/core.py
```diff
@@ -328,8 +328,10 @@ def create_cmdline(self, progname, description=None, switchlist=[]):
                 sys.exit()
             if 'arg_step' in cmdargs.keys():
                 self.set('arg', 'step', cmdargs['arg_step'], clobber=True)
+            if 'fpga_partname' in cmdargs.keys():
```
WRansohoff, line 331: Are these parameters set inside of the `if 'target' in cmdargs.keys():` block because they could be overwritten by the `load_target` call if they aren't already set when it is called?

siliconcompiler/core.py
```diff
@@ -1579,22 +1540,28 @@ def check_manifest(self):
...
             steplist = self.list_steps()
 
         #1. Checking that flowgraph is legal
-        if not self.getkeys('flowgraph'):
+        if flow not in self.getkeys('flowgraph'):
             self.error = 1
-            self.logger.error(f"No flowgraph defined.")
-        legal_steps = self.getkeys('flowgraph')
+            self.logger.error(f"flowgraph {flow} not defined.")
+        legal_steps = self.getkeys('flowgraph',flow)
```
WRansohoff, line 1552: Nit:

```suggestion
        legal_steps = self.getkeys('flowgraph', flow)
```

siliconcompiler/client.py
```diff
@@ -51,16 +51,16 @@ def remote_preprocess(chip):
     active = manager.dict()
 
     # Setup up tools for all local functions
-    remote_steplist = chip.getkeys('flowgraph')
+    remote_steplist = chip.getkeys('flowgraph',chip.get('flow'))
```
WRansohoff, line 54: Normally I wouldn't comment on missing spaces, but the 'suggest changes' feature makes it easy to address these nitpicks quickly.

```suggestion
    remote_steplist = chip.getkeys('flowgraph', chip.get('flow'))
```

siliconcompiler/core.py
```diff
@@ -1901,14 +1873,14 @@ def check_checklist(self, standard, item=None):
             report_ok = False
             criteria_ok = True
             # manual
-            if step not in self.getkeys('flowgraph'):
+            if step not in self.getkeys('flowgraph',flow):
```
WRansohoff, line 1876: Nit:

```suggestion
            if step not in self.getkeys('flowgraph', flow):
```

aolofsson:
> I think I managed a relatively thorough look! Found mostly nitpicks/typos, I may have caught one thing that's a small bug but should be an easy fix. Comments inline.
> 
> A few higher-level points:
> 
> * I'm not sure how I feel about the default target names ending in `*_demo`. Makes them sound a bit "toy" or incomplete to me, when I figure that users would probably be okay just using them as-is if they're going with open tools (and note they also have to use these with the cloud service). But if the goal is to encourage people to fork these and write their own, then I think it makes sense. Also, sorry for not voicing that thought sooner, this doesn't seem fun to change...
> * We need to update [targets page](https://docs.siliconcompiler.com/en/latest/user_guide/targets.html) in docs, and should probably expand the docs generators to cover libs and targets. I can take both on as an action in a future PR if you'd like.

Yes, if you can generate a PR for that it would be great!

In terms of the "* _demo",  these flows are in fact "toys" as of now. Perhaps "*_beta" would be better. Let's discuss and also think about minimizing the pain. These types of changes are not fun.:-)

siliconcompiler/core.py
```diff
@@ -2082,15 +2063,15 @@ def write_flowgraph(self, filename, fillcolor='#ffffff',
         dot = graphviz.Digraph(format=fileformat)
         dot.graph_attr['rankdir'] = rankdir
         dot.attr(bgcolor='transparent')
-        for step in self.getkeys('flowgraph'):
+        for step in self.getkeys('flowgraph',flow):
```
WRansohoff, line 2066: Nit:

```suggestion
        for step in self.getkeys('flowgraph', flow):
```

siliconcompiler/core.py
```diff
@@ -2540,18 +2522,17 @@ def check_logfile(self, step=None, jobname=None, index=None,
...
             jobname = self.get('jobname')
         if logfile is None:
             logfile = f"{step}.log"
 
-        tool = self.get('flowgraph', step, index, 'tool')
+        flow = self.get('flow')
+        tool = self.get('flowgraph', flow, step, index, 'tool')
         design = self.get('design')
 
+
```
WRansohoff, line 2539: Extraneous newline?

```suggestion
```

siliconcompiler/core.py
```diff
@@ -2601,12 +2582,13 @@ def summary(self, steplist=None, show_all_indices=False):
...
 
         # display whole flowgraph if no steplist specified
+        flow = self.get('flow')
         if not steplist:
             steplist = self.list_steps()
 
         #only report tool based steps functions
         for step in steplist:
-            if self.get('flowgraph',step,'0','tool') in self.builtin:
+            if self.get('flowgraph',flow, step,'0','tool') in self.builtin:
```
WRansohoff, line 2595: Nit:

```suggestion
            if self.get('flowgraph', flow, step,'0','tool') in self.builtin:
```

siliconcompiler/core.py
```diff
@@ -2632,7 +2614,7 @@ def summary(self, steplist=None, show_all_indices=False):
         if self.get('mode') == 'asic':
             info_list.extend(["foundry : " + self.get('pdk', 'foundry'),
                               "process : " + self.get('pdk', 'process'),
-                              "targetlibs : "+" ".join(self.get('asic', 'targetlib'))])
+                              "targetlibs : "+" ".join(self.get('asic', 'logiclib'))])
```
WRansohoff, line 2621: Nit: make it clear that the `+` character is an operator, not a string?

```suggestion
                              "targetlibs : " + (" ".join(self.get('asic', 'logiclib')))])
```

siliconcompiler/libs/asap7sc7p5t.py
```diff
@@ -0,0 +1,91 @@
...
+
+    for libname in all_libs.keys():
+        libdir = os.path.join('..', 'third_party', 'pdks', foundry, process, 'libs', libname, rev)
+        suffix = all_libs[libname]
+
+        # rev
+        chip.set('library',libname, 'package', 'version',rev)
+
+        # timing
+        chip.add('library', libname, 'nldm', corner, 'lib',
```
WRansohoff, line 42: Should we be using `.add` here, or `.set`? Same for other calls later in this file.

siliconcompiler/libs/nangate45.py
```diff
@@ -0,0 +1,104 @@
...
+                          rev)
+
+    # standard cell typ
+    chip.set('library',libname,'type','stdcell')
+
+    # rev
+    chip.set('library',libname, 'package', 'version',rev)
+
+    # timing
+    chip.add('library',libname, 'nldm', corner, 'lib',
```
WRansohoff, line 46: Ditto the `asap7.py` commetn; `.add` or `.set`? (And ditto throughout this file)

siliconcompiler/libs/sky130.py
```diff
@@ -0,0 +1,127 @@
...
+    libdir = os.path.join('..', 'third_party', 'pdks', foundry, process, 'libs', libname, rev)
+
+
+    chip.set('library', libname, 'type', 'stdcell')
+
+    # rev
+    chip.set('library', libname, 'package', 'version', rev)
+
+    # timing
+    chip.add('library', libname, 'nldm', corner, 'lib',
```
WRansohoff, line 39: Same question as other PDKs; `.add` or `.set` for these calls?

WRansohoff (approved):
Looks good to me, the only comment that really needs looking at is the question of whether to use `.add` or `.set` in the `libs/...` setup methods. Good idea adding a file to track changes across versions.

Sorry if all of the 'missing space' suggestions are distracting; I refrained from making them in the `libs/...` files.

WRansohoff:
Oh, looks like the 'suggest changes' feature might not work on PRs that are already merged...sorry, I tried to read through this as quickly as I could once the tests started passing.

PR #863, opened 2022-02-07
Add support for KLayout .lyp files to sc-show

WRansohoff:
This change allows us to use either `.lyt` or `.lyp` files to configure display properties in `sc-show`.

The `.lyp` file contains layer display properties such as color/pattern/etc, and we currently read its location from the `.lyt` file which is passed into the `klayout_show.py` script. This change allows us to specify either a `.lyp` or `.lyt` file in the `['pdk']['layermap']['klayout'][...]` parameter.

### DIFF
--- a/siliconcompiler/tools/klayout/klayout_show.py
+++ b/siliconcompiler/tools/klayout/klayout_show.py
@@ -19,12 +19,14 @@ sc_libtype = sc_cfg['library'][sc_mainlib]['arch']['value']
 # or in the standalone format (.lyp).
 tech_filename = sc_cfg['pdk']['layermap']['klayout'][sc_stackup]['def']['gds']['value'][0]
 tech_file = None
-if tech_filename[-4:] == '.lyt':
-    # 'lyp_path' will be read out of the .lyt file later.
-    tech_file = tech_filename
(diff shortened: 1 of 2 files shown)


nmoroze:
This looks good to me! 

Idea for the future: next time we tweak the PDK schema, it could be worth thinking about if it makes sense to come up with a separate spot for the lyt file. That way we could make the layermap parameter always store the path to the lyp, make the lyt an optional entry in a different parameter, and get rid of the default behavior of loading the lyp based on the lyt (I just realized it's a little hacky).

WRansohoff:
Good idea. Allowing the `.lyp` extension seemed to make sense in this case because the parameter is called 'layermap', but maybe we should consider a separate parameter like 'displaytech'.

Or maybe we could remove the .lyt usage if `sc-show` doesn't use the extra information that the file provides. But we do configure the KLayout view with a `layoutOptions` object which gets retrieved from the tech file if possible, so we might not want to entirely remove that functionality.

aolofsson:
If there is a need for both files, then it makes sense to have two different keys for them. 
PR looks good!

aolofsson (approved):

PR #864, opened 2022-02-08
Use 'flow' parameter to specify standalone tool run

nmoroze:
This PR builds on top of #862 and fixes its failing tests. The main change is adjusting the way that standalone tool runs are set up. It gets rid of `load_tool()` -- we don't need this function since each tool's `setup()` function is called automatically inside of `run()`, and giving the user the ability to call it earlier can be problematic since it introduces ordering issues.

I chose overriding the 'flow' schema parameter as the new way to specify a standalone tool run. Now, if the user specifies a flow that doesn't have a configured flowgraph, `run()` will automatically set up a basic flowgraph to run that tool. I considered making a separate helper function for loading a standalone tool, but I realized using a schema parameter is necessary to support running tools in standalone mode from the CLI. Instead of adding an additional parameter, reusing `flow` seemed logical since the standalone tool is effectively the flow you want to run.

@aolofsson I'm opening this PR as relative to your PR's branch rather than committing directly to make it easier for you to review these changes. Once you're ready, you can merge this PR and these commits should show up in your original PR, which you can then merge into main.

### DIFF
--- a/docs/reference_manual/core_api.rst
+++ b/docs/reference_manual/core_api.rst
@@ -58,7 +58,6 @@ This chapter describes all public methods in the SiliconCompiler core Python API
     ~siliconcompiler.core.Chip.load_lib
     ~siliconcompiler.core.Chip.load_pdk
     ~siliconcompiler.core.Chip.load_target
-    ~siliconcompiler.core.Chip.load_tool
     ~siliconcompiler.core.Chip.merge_manifest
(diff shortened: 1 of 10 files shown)


aolofsson (approved):

aolofsson:
Love it. That's some fancy git footwork.;-)
Also love that you fixed the bugs by removing code. Very elegant.:-)

PR #867, opened 2022-02-11
Using metric as keyword into tool report parameter

aolofsson:
- Enables linking of metric summary to the source that generated that table.

### DIFF
--- a/Changes
+++ b/Changes
@@ -38,7 +38,8 @@ SiliconCompiler 0.6.0 (2022-02-11)
 * Schema: Changed tool version switch to a list
 * Schema: Changed 'asic', 'targetlib' to 'asic', 'logiclib' for clarity.
 * Schema: Changed 'eda, 'report' parameter guideline to always use 'metric' as keyword
-* Schema: Added -skipcheck option to speed up new target bringup
+* Schema: Added -skip_check option to speed up new target bringup
+* Schema: Added -skip_step option to enable skipping specific steps
 
 SiliconCompiler 0.4.1 (2022-01-06)
 ===================================
--- a/Changes
+++ b/Changes
@@ -38,6 +38,7 @@ SiliconCompiler 0.6.0 (2022-02-11)
 * Schema: Changed tool version switch to a list
 * Schema: Changed 'asic', 'targetlib' to 'asic', 'logiclib' for clarity.
 * Schema: Changed 'eda, 'report' parameter guideline to always use 'metric' as keyword
+* Schema: Added -skipcheck option to speed up new target bringup
 
 SiliconCompiler 0.4.1 (2022-01-06)
 ===================================
--- a/Changes
+++ b/Changes
@@ -33,11 +33,11 @@ SiliconCompiler 0.6.0 (2022-02-11)
 
 **Minor:**
 
-* Schema: Moving lib 'driver' into cells (consistency)
-* Schema: Adding site symmetry to avoid full lef parser.
-* Schema: Make tool version switch a list
-* Schema: Changing 'asic', 'targetlib' to 'asic', 'logiclib' for clarity.
-
+* Schema: Changed lib 'driver' to move into cells (consistency)
+* Schema: Added site symmetry to avoid full lef parser.
+* Schema: Changed tool version switch to a list
+* Schema: Changed 'asic', 'targetlib' to 'asic', 'logiclib' for clarity.
+* Schema: Changed 'eda, 'report' parameter guideline to always use 'metric' as keyword
 
 SiliconCompiler 0.4.1 (2022-01-06)
 ===================================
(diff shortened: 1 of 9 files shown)


siliconcompiler/tools/surelog/surelog.py
```diff
@@ -133,10 +135,20 @@ def post_process(chip):
...
     if step != 'import':
         return 0
 
+    # Log file parsing
+    chip.set('eda', tool, 'regex', step, index, 'warnings', "WARNING")
+    chip.set('eda', tool, 'regex', step, index, 'errors', "ERROR")
+
+    # Output reprts for deep dive
+    for metric in ('errors', 'warnings'):
+        chip.set('eda', tool, 'report', step, index, metric, f"{step}.log")
```
WRansohoff, line 150: I think these paths cannot be found in the `check_manifest` test because they are relative to the step directory, not the top-level build directory.

```suggestion
    # Output reprts for deep dive
    logfile = os.path.join(os.getcwd(), f"{step}.log")
    local_logfile = logfile[logfile.rfind(chip.get('dir')):]
    for metric in ('errors', 'warnings'):
        chip.set('eda', tool, 'report', step, index, metric, local_logfile)
```

WRansohoff (commented):
Looks good to me so far. Two main things:

* This branch seems to be missing empty files at `targets/__init__.py` and `libs/__init__.py`.

* The 'check_manifest' test failure seems to be caused because the call is made from one directory above the build directory, not from inside the step directory.

See the suggestion for one option to fix the failing test - yosys/openroad/etc may also need similar path modifications.

siliconcompiler/tools/surelog/surelog.py
nmoroze: Nice debugging, and good suggestion. One other thing we could consider is handling these report keypaths as a special case in `chip.find_files()`, which I generally wouldn't be a fan of except there is precedent for doing it for the 'input'/'output' keypaths (which are very similar): https://github.com/siliconcompiler/siliconcompiler/blob/main/siliconcompiler/core.py#L1279-L1290.

Changes
```diff
@@ -33,11 +33,13 @@ SiliconCompiler 0.6.0 (2022-02-11)
...
-* Schema: Make tool version switch a list
-* Schema: Changing 'asic', 'targetlib' to 'asic', 'logiclib' for clarity.
-
+* Schema: Changed lib 'driver' to move into cells (consistency)
+* Schema: Added site symmetry to avoid full lef parser.
+* Schema: Changed tool version switch to a list
+* Schema: Changed 'asic', 'targetlib' to 'asic', 'logiclib' for clarity.
+* Schema: Changed 'eda, 'report' parameter guideline to always use 'metric' as keyword
+* Schema: Added -skip_check option to speed up new target bringup
+* Schema: Added -skip_step option to enable skipping specific steps
```
WRansohoff, line 42: Should we mention the new tool-specific pdk schema parameters here? ('file', 'directory', 'variable')?
```suggestion
* Schema: Added -skip_step option to enable skipping specific steps
* Schema: Added 'pdk', ['file', 'directory', 'variable'] parameters to enable tool-specific PDK setups.
```

siliconcompiler/core.py
```diff
@@ -1612,10 +1615,11 @@ def check_manifest(self):
 
         # Dynamic checks
         # We only perform these if arg, step and arg, index are set.
-        # We don't check inputs for checkonly pass through
+        # We don't check inputs for skip all
+        # TODO: Need to add skip step
         step = self.get('arg', 'step')
         index = self.get('arg', 'index')
-        if step and index and not self.get('checkonly'):
+        if step and index and not self.get('skip','all'):
```
WRansohoff, line 1622: Nit: space

```suggestion
        if step and index and not self.get('skip', 'all'):
```

WRansohoff (approved):
Looks good to me, just a couple of minor doc/style suggestions

siliconcompiler/core.py
```diff
@@ -1612,10 +1615,11 @@ def check_manifest(self):
 
         # Dynamic checks
         # We only perform these if arg, step and arg, index are set.
-        # We don't check inputs for checkonly pass through
+        # We don't check inputs for skip all
+        # TODO: Need to add skip step
```
nmoroze, line 1619: Mark WIP in schema if not implemented yet?

nmoroze (approved):
LGTM as well. I think the only thing that gives me pause is the fact that steps marked as 'skip'-ed still run the executable for version check. That doesn't seem very clear from the name, and I could imagine someone trying to use this to skip e.g. a DRC check if they don't have magic installed. I'm not sure it's a big deal since they can just skip the step using steplist, but we should probably be clear that skipped steps will run the executable for a version check in the docs.

siliconcompiler/core.py
aolofsson: Until we start getting bug reports stating that someone has used an unimplemented feature and failed, I don't think we should spend too much time worrying about this. The schema defines the spec, not the implementation.

PR #868, opened 2022-02-14
Small fixes for the new v0.6 schema

WRansohoff:
We need `__init__.py` files in the `targets/` and `libs/` directories so that Python treats the files in those directories as modules. There's also a missing `chip.get('flow')` key in `scheduler.py`.


aolofsson (approved):

aolofsson:
Sorry about that. Thanks for catching and fixing those!

PR #870, opened 2022-02-15
Clean up tech-specific variables in OpenROAD

nmoroze:
This PR removes the tech-specific variable default values from openroad.py. It now drives the defaults through the PDK. 

Also included some misc cleanup of the file.


aolofsson (approved):

aolofsson:
LGTM! Nice to get the IP leakage out of the openroad tool setup.

PR #871, opened 2022-02-15
Clean up Yosys latch techmapping

nmoroze:
This PR adds a schema entry for general techmap files, and uses it to drive the latch techmap files into Yosys. This lets us get rid of the hardcoded tech-specific stuff from the Yosys tcl driver.


nmoroze:
Makes sense, always good to keep things simple. Fixed!

aolofsson:
Looks good!

aolofsson (approved):

PR #872, opened 2022-02-17
Fix bug in _allpaths()

nmoroze:
This PR fixes a bug I caught in the `_allpaths()` function today. The issue is that it had a `return` statement inside the for loop that loops over all inputs to the current task, meaning that it wouldn't actually generate all paths (just one of them).

Depending on the order in which you set up your flowgraph, this meant that `list_steps()` could give you an incorrect ordering, resulting in a confusing summary table. I demonstrated this with a new test, `tests/core/test_list_steps.py`. The simple flowgraph in the test looks like this:

test_list_steps

Without this fix, `list_steps()` returns `['A', 'B', 'D', 'C']`, because it only considers the path from A->D. With this fix, you get the correct order of `['A', 'B', 'C', 'D']`.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -2740,23 +2740,28 @@ class Chip:
         return list(sorted_dict.keys())
 
     ###########################################################################
-    def _allpaths(self, cfg, flow, step, index, path=None, allpaths=None):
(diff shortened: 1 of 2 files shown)


aolofsson:
Great, so it was completely broken....sorry about that! Great find and fix!

aolofsson (approved):

PR #875, opened 2022-02-19
Fix jobinput/record bug

nmoroze:
This PR fixes a subtle bug with the way the 'jobinput' parameter is handled. For any task that has a 'jobinput' set, the `job` variable gets overwritten with the name of the input job, but this `job` variable is also used to determine where records get stored when 'track' is set to True (see [core.py line 3413](https://github.com/siliconcompiler/siliconcompiler/blob/e724bcfb290a846710a6f44368b72a97e9b0c0d9/siliconcompiler/core.py#L3413)). 

Changing the variable name fixes this problem and should make thing safer from bugs in the future, since the new name is a bit more explicit.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3170,10 +3170,11 @@ class Chip:
 
         # support for sharing data across jobs
         job = self.get('jobname')
+        in_job = job
         if job in self.getkeys('jobinput'):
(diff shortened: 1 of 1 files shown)


aolofsson (approved):

aolofsson:
Nice catch!

PR #877, opened 2022-02-23
Fix bug causing [arg, index] to not be respected

nmoroze:
Since we clobber [arg, index] when running the tool setup loop, using it
later to determine whether or not to only run a single index for a step
doesn't work. Precomputing a per-step indexlist avoids this problem and
also makes the code cleaner.

This also fixes an additional uncaught bug where the indexlist parameter
wouldn't be respected by run().

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3657,25 +3657,28 @@ class Chip:
         self.set('arg', 'step', None, clobber=True)
         self.set('arg', 'index', None, clobber=True)
 
-        # Merge cfg back from last executed runstep.
-        # (Only if the index-0 run's results exist.)
+        # Merge cfg back from last executed runsteps.
+        # Note: any information generated in steps that do not merge into the
+        # last step will not be picked up in this chip object.
         laststep = steplist[-1]
-        lastindex = '0'
-        lastdir = self._getworkdir(step=laststep, index=lastindex)
-        lastcfg = f"{lastdir}/outputs/{self.get('design')}.pkg.json"
-        if os.path.isfile(lastcfg):
-            local_dir = self.get('dir')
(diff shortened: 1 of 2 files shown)


aolofsson:
Are you sure about setting the arg index default to None rather than 0? Perhaps not a big deal, but it means that the arg index would always need to be set whenever the arg step is set. Some tools would always have index=0 (eg. export/import). Here's one example of a "bug" that relies on arg index being zero:

https://github.com/siliconcompiler/siliconcompiler/blob/3d750a2392e57b983c97165a2c81b234c4f81adc/siliconcompiler/core.py#L337

nmoroze:
Good question, but with the other changes in this PR I think you shouldn't even need to specify an arg, index just because you set an arg, step. The logic for choosing what indices to run:

https://github.com/siliconcompiler/siliconcompiler/blob/1a6c88c794bdb9a2e398adf7cba6cf8dafeaf1ac/siliconcompiler/core.py#L3516-L3523

If there is no arg, index specified, it will fall back to selecting all the indices associated with the step in the flowgraph (in most cases, just '0'). However, if we have a defvalue for arg, index, then the first call in that if statement returns '0', and we only run that index regardless of what's in the flowgraph. 

This wasn't a problem before since there was another masking bug (the main object of this PR) that was causing the arg, index to basically get ignored completely.

nmoroze:
(BTW, marked as draft for now since I'm not 100% sure if the final commit is working the way I want, still doing some testing)

nmoroze:
Alright, this should be good to go on my end!

aolofsson:
This should be good to merge right?

nmoroze:
Yes!

aolofsson (approved):

PR #878, opened 2022-02-23
Some small cleanup things in core.py

nmoroze:
A few changes I had in my local branch:

- Docstring fix
- Fix bug in `_check_flowgraph_io()` that would occasionally cause checks to fail when using 'jobinput'
- Remove return value from `read_manifest()`. This is a very small change but might be the most controversial - however, I don't think we rely on the return value of this function anywhere (it's not even documented), and it can be annoying when using the interactive Python shell. The problem is that the REPL prints out the return value of any function when it's not assigned to something, so if you do `chip.read_manifest(...)` it will print out a massive dictionary that obscures previous command history.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -433,10 +433,10 @@ class Chip:
         Args:
             modulename (str): Name of module to import.
             funcname (str): Name of the function to find within the module.
-            moduletype (str): Type of module (flows, pdks,libs,targets).
+            moduletype (str): Type of module (flows, pdks, libs, targets).
 
         Examples:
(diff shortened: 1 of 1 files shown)


aolofsson:
Yes, I am ok with removing the return localcfg. The function is a trivial json/yaml load without the merge, and clearly most users should be loading the manifest into the in memory cfg dict.  Can you update the dostring (remove return statement) and remove the redundant Update=True argument. (function turns into a nop with that set to false).

nmoroze:
Nice catch, done! Somehow I completely missed the return value in the docstring...

Also added a docstring line for the 'job' argument (took description from merge_manifest()), since I saw that one was missing.

aolofsson (approved):
Looks good, thanks!

PR #880, opened 2022-02-23
Two small fixes

nmoroze:
- Fix installgen: `os.listdir()` doesn't give full path, so the `os.path.isfile` check always fails. Turns out the list of setup scripts has been missing from the docs for a while!
- Fix skipped test_tool_option: Problem was that with the new target scheme the test was hacking invalid changes into asicflow, so we needed to set up a test flow in its own "lane".


aolofsson (approved):

aolofsson:
Looks good!

PR #882, opened 2022-02-25
Add no-op lisdir to fix network filesystem issue

nmoroze:
I ran into a weird bug with a Slurm-based run using NFS to host the shared build directory. The main `run()` thread consistently failed to find the final results file, even though manual inspection after the completed run showed that it exists.

I did some digging around, and it seems likely that it's related to some sort of NFS cache coherency issue. This person had a very similar scenario: https://stackoverflow.com/questions/21292365/python-os-path-exists-failing-for-nfs-mounted-directory-file-that-exists. 

One of the answers suggested that using an `os.system('ls')` call fixes the issue by forcing a cache flush, and I found that the Python `listdir()` method seems to have the same effect (which I actually noticed by accident while using it to debug).

I figure we can merge this since this is a common Slurm use case and I think this line is otherwise harmless (despite it being an incredible hack). Another option seems to be to configure the filesystem not to cache file attributes, but this may have a performance impact. This is something we could look into, and it would require no changes to SC, just carefully documented methodology.

@WRansohoff, have you noticed any issues like this? I think you might have mentioned something related the other day, but I can't remember. Would also love to get your feedback on the problem/solution in general!

(Note: this PR is based off PR #877, if that one is merged first I might have to rebase this one)

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3664,6 +3664,14 @@ class Chip:
         last_step_failed = True
         for index in indexlist[laststep]:
             lastdir = self._getworkdir(step=laststep, index=index)
+
+            # This no-op listdir operation is important for ensuring we have a
+            # consistent view of the filesystem when dealing with NFS. Without
+            # this, this thread is often unable to find the final manifest of
+            # runs performed on job schedulers, even if they completed
+            # successfully. Inspired by: https://stackoverflow.com/a/70029046.
+            os.listdir(os.path.dirname(lastdir))
+
             lastcfg = f"{lastdir}/outputs/{self.get('design')}.pkg.json"
             if os.path.isfile(lastcfg):
                 last_step_failed = False


WRansohoff (approved):
Yep, I remember running into a similar issue when running load tests; I think we might have briefly touched on that when we talked yesterday. I worked around the issue by re-mounting the network drive in between jobs, but if running `ls` also works to flush the filesystem cache, that sounds like a much better solution. Good catch!

aolofsson:
Wow, that's a tricky one! Can we discuss some options Monday? Not a 100% sure the listdir is the best option, but I don't necessarily have a better one....wouldn't it be better to flush writes at the end of a runtask? Seems less risky for the future if we can make sure that all writes have finished before exiting a compute node task.

https://www.chpc.utah.edu/documentation/hpcfaq.php
https://www.avidandrew.com/understanding-nfs-caching.html

nmoroze:
Did some experiments, and looks like an `os.sync()` at the end of `_runtask()` has no effect (i.e. if I add that in and remove the `os.listdir()` hack, the code is still buggy). 

I think this confirms that the problem is not that the write doesn't commit in time, the problem is that the reader doesn't see the results of the write since it's reading from a stale cache. I couldn't find anything more explicit for invalidating an NFS cache, but I found someone else confirming that the `ls` trick works because it invalidates the cache for that directory: https://stackoverflow.com/a/30630912. This makes me pretty comfortable with using `os.listdir()` as our solution.

I could commit the `sync` as well, but I'm not sure if it's worth it given that we don't have any evidence that there's a problem it's solving yet.

One final comment is that I think a good solution could be to move away from relying on the existence of the final manifest to determine whether or not a run succeeds. We already query error bits for non-scheduler and Slurm-based local runs, so we can assume this file *should* exist at this point in the code. Based on my research it seems like NFS caching is time-based, so we could have some sort of retry with backoff that attempts to open the file a couple times (with increasing delays) before giving up. I think the only thing we'd need to add is some data returned by the remote run API to give a success status for the run (or if that exists already, we just need to use it in run()). This could also let us clean up the hacky logic that iterates through all steps to find the first failure. I'll make a note about this, but we can delay implementing it for now since I think a remote API change would be a bit more onerous than this fix.

aolofsson:
I agree that the listdir seems like a good fix for now, but let's to go back and review the general execution flow again to make sure we are not missing something that will bite us later. These are not nice bugs to debug.:-) Synchronization should be bullet proof.

PR #883, opened 2022-02-26
In-browser metrics report

WRansohoff:
This PR contains logic to generate a browser-based report for a completed build, with a basic table of metrics and low-res screenshot of the design. The report is automatically generated during the `chip.summary()` call.

A few open questions before I finalize this draft:

* Should we try to open the report file in a browser if the `-quiet` flag is not set, or should we just generate the report?

* Does the KLayout screenshot logic look okay? The page looked a bit empty with only a few buttons and tables, so I separated it into two columns. The metrics tables go on the right, and the signoff checklist area goes on the left under a small GDS preview image.

* Is there a better way to find the job's final task? It's probably not ideal to assume that the manifest will be located under `/export/0/`.


WRansohoff:
Updated with feedback from our discussion this morning:

* The report file is only generated when the 'mode' parameter is set to 'asic'.
* Added a `-nodisplay` flag to use instead of `-quiet` for suppressing the final report page (and other GUI windows which we might add later).
* Used the `find_results` method instead of assuming `export/0/outputs/[design].gds`.
* Added an extra schema parameter for task-specific environment variables. This lets us, for example, set an environment variable in `klayout.py` instead of using the `-z` flag to take a screenshot (although the latter approach seems cleaner):

      chip.set('eda', tool, 'environment', step, index, 'QT_QPA_PLATFORM', 'offscreen')

aolofsson:
Sorry for being unclear. I meant that the gds should only be shown if it exists. The html report should always be generated. It's generic enough.  The list of metrics to be reported is based on the flowgraph setup.

WRansohoff:
Got it, I made a change to generate the report regardless of the 'mode' parameter, and to omit the screenshot preview if a GDS results file does not exist. I also added a few fixes in the template so that a report can be compiled for the blinky/ice40 FPGA example.

aolofsson (approved):
This is awesome!

PR #884, opened 2022-02-27
Update klayout version in install-klayout.sh

sconvent:
The old URLs are dead. The new ones point to the current 0.26.XX version.


nmoroze (approved):
Good catch, thanks for the PR!

PR #885, opened 2022-03-01
Some documentation enhancements

nmoroze:
- Use Sphinx refs for linking between dynamic modules so that the links work for custom modules/within a PDF
- Display list-type schema values as bulleted list
- Add 'nobuiltins' option for dynamic module autogen -- causes them to
  only document stuff on SCPATH (off by default)
- Reorder lists in targets docs
- Bugfix: make flowgraph image generation more robust to odd build directories
- Bugfix: make skip_zero_weight work again (broke due to schema update


aolofsson (approved):
Looks good!

PR #886, opened 2022-03-01
Account for `environment` parameters being stored as an array

WRansohoff:
The `environment` schema parameter is stored as an array, not a string. So it probably makes sense to use `':'.join` with these values, even though there's usually only one value for each environment variable.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3385,7 +3385,7 @@ class Chip:
         if (step in self.getkeys('eda', tool, 'environment')) and \
            (index in self.getkeys('eda', tool, 'environment', step)):
             for item in self.getkeys('eda', tool, 'environment', step, index):
-                os.environ[item] = self.get('eda', tool, 'environment', step, index, item)
+                os.environ[item] = ':'.join(self.get('eda', tool, 'environment', step, index, item))
 
         ##################
         # 14. Check exe version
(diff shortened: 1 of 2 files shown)


nmoroze:
Do you think it would make more sense to switch this schema parameter to a scalar? I'm not sure if the ":" separator is general enough to make it standardized for any arbitrary env variable.

WRansohoff:
Good idea, that does seem like a better solution. The user can enter the variables with whatever separators the application expects.

aolofsson:
What would be the alternative to ":"? 
https://tldp.org/LDP/Bash-Beginners-Guide/html/sect_03_02.html

WRansohoff:
It seems like the most common separator for shell variables, but I don't think there's any technical reason why an individual tool couldn't use other characters like `,` or `|`.

nmoroze:
As a concrete example, Windows uses `;` for the $PATH (which is what [os.pathsep](https://docs.python.org/3/library/os.html#os.pathsep) is for), but I didn't really think having a standard way to pass "list-type" env variables besides path-esque things was common. If it's common for EDA tools I'd be okay with reverting though.

nmoroze:
If we want to be really nitpicky, it could also be worth considering consistency with our global environment variable parameter: https://github.com/siliconcompiler/siliconcompiler/blob/main/siliconcompiler/schema.py#L3740 (it's a scalar right now, but of course if we decide a list is better we could just change this one as a fixup item).

aolofsson:
@nmoroze Yeah, I think you are right. The colon is a shell hack that doesn't translate across platforms. Better to just feed in a string. Closed.

aolofsson (approved):

PR #888, opened 2022-03-02
Simplify scheduler log output

WRansohoff:
Slurm seems to have a bit of trouble writing log files to non-existant or newly-created directories.

I've been writing the slurm log files to the build's root directory and relocating them in the server's `get_results` call as a workaround; it seems like we should formalize that change since it's been in effect for a couple of releases.

### DIFF
--- a/siliconcompiler/scheduler.py
+++ b/siliconcompiler/scheduler.py
@@ -31,19 +31,10 @@ def _deferstep(chip, step, index, active, error):
             slurm_constraint = chip.status['slurm_constraint']
         else:
             slurm_constraint = 'SHARED'
-        # The task directory may not exist before the 'sc' command is run;
-        # create it to ensure that the log file can be written to.
-        os.makedirs(os.path.join(chip.get('dir'),
-                                 chip.get('design'),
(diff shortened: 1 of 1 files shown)


aolofsson:
In regards to the "newlyc reated" directories, what does that mean. Is that another NFS  race problem?

WRansohoff:
I think so; the directory gets created on the node that runs the `sbatch` command, and the compute node which receives the job may have a stale cache of the build directory when it tries to create the log file. And it tries to create the log file before it runs any script logic, so it would be difficult to force a refresh by adding logic to the shell script or `sc` logic.

We may be able to configure a "prefix" function in slurm, but I've read that that can cause performance issues when the scheduler has high throughput, and this workaround seems simpler.

nmoroze:
For what it's worth, I wanted to echo that we've been running this as a patch in production for a while, so I'm in favor of merging it for real to make things easier to manage. I think we could revisit it when we revisit the listdir hack.

aolofsson:
Sounds good!

aolofsson (approved):

PR #889, opened 2022-03-02
Fix report generation for remote runs.

WRansohoff:
The report generation logic relies on some schema parameters which get removed by the server sanitizer, so we need a bit of logic to make that part of the `summary()` function fault-tolerant, and to generate a final report locally after a remote run.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -2761,17 +2761,22 @@ class Chip:
             # Generate results page by passing the Chip manifest into the Jinja2 template.
             env = Environment(loader=FileSystemLoader(templ_dir))
             results_page = os.path.join(web_dir, 'report.html')
-            with open(results_page, 'w') as wf:
-                wf.write(env.get_template('sc_report.j2').render(
-                    manifest = self.cfg,
-                    metric_keys = metric_list,
-                    metrics = self.cfg['metric'],
-                    tasks = flow_tasks,
-                    results_fn = results_gds
-                ))
+            results_gen = False
+            try:
+                with open(results_page, 'w') as wf:
+                    wf.write(env.get_template('sc_report.j2').render(
+                        manifest = self.cfg,
+                        metric_keys = metric_list,
+                        metrics = self.cfg['metric'],
+                        tasks = flow_tasks,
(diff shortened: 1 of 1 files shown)


nmoroze:
Seems like a good idea to make sure report generation errors aren't fatal. Out of curiosity, what parameters were causing the issue?

Also, not sure if there's any need to automatically call summary() at the end of run(), hasn't that historically been the user's responsibility?

WRansohoff:
It looked like a lack of tool parameters under 'eda'; it couldn't find metrics from the surelog import step.

Good point, maybe it makes more sense to move the report generation into its own function which gets called from `summary()` and at the end of a remote run.

nmoroze:
> It looked like a lack of tool parameters under 'eda'; it couldn't find metrics from the surelog import step.

Ah, that makes sense! I was wondering why this would be a problem since tools would be producing metrics and configuring reports on the other side of the sanitizer, but I forgot about the local import. Longer-term fix might be to limit report generation based on arg,step/arg,index, which I think is how summary() itself works. It seems like a waste to do any report gen on the server though when the user can make one for the entire end at the run though, so I like this for now.

> Good point, maybe it makes more sense to move the report generation into its own function which gets called from summary() and at the end of a remote run.

I'm still confused why we need to do report generation at the end of a remote run automatically? I figured we'd just let the user call summary() to generate the report, like they'd do for a regular run. Not sure if I'm missing something, though.

WRansohoff:
Good point - I think that we used to skip the `summary()` function when the user ran `sc [...] -remote` on the command line, but that doesn't seem to be the case anymore.

WRansohoff:
Okay, I think this should be ready to merge if you're happy with it.

nmoroze:
Looks good to me, thanks!

WRansohoff:
Okay, per our discussion today I replaced the `self.list_steps()` call with the existing `steplist` variable, and removed the blanket try/except statements. With this new single-line change, the report generation appears to succeed on our test cluster.

Thanks for drilling into this, it looks like my use of `self.list_steps()` was a bug, and fixing that obviates the need for extra error handlers.

aolofsson (approved):

PR #892, opened 2022-03-03
Adding oformat parameter to schema.

aolofsson:
- Necessary to enable selection between different output formats
- gds vs oasis is one example
- gerber vs ipc2581 is another

### DIFF
--- a/siliconcompiler/schema.py
+++ b/siliconcompiler/schema.py
@@ -4836,6 +4836,24 @@ def schema_design(cfg):
         """
     }
 
+    cfg['oformat'] = {
+        'switch': "-oformat <str>",
+        'type': 'str',
+        'lock': 'false',
+        'require': None,
+        'signature': None,
+        'defvalue': [],
+        'shorthelp': 'Output format',
+        'example': ["cli: -oformat gds",
+                    "api: chip.set('oformat', 'gds')"],
+        'help': """
+        File format to use for writing the final siliconcompiler output to
+        disk. For cases, when only one output format exists, the 'oformat'
+        parameter can be omitted. Examples of ASIC layout output formats
+        include GDS and OASIS.
+        """
+    }
+
     cfg['constraint'] = {
         'switch': "-constraint <file>",
         'type': '[file]',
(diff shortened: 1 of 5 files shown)


aolofsson:
@nmoroze Can you take a look at the failing test. I added a a bit of logic to the check_manifest to let us run it without having the tool setup. Useful for checking flows pre-run....but it broke the check_manifest test. Not sure why?

nmoroze:
I think it broke a test that is checking to make sure the `check_manifest()` fails when it's supposed to. I don't think you can do `self.valid('eda', tool)` since valid checks for a full match, so now it's never checking and isn't picking up the failure. Maybe do check like: `tool in chip.getkeys('eda')`? Although this isn't foolproof since user might specify some 'eda' value like 'variable' or 'threads' before setup is run, so you could try checking something like the full exe keypath (not sure if picking one is too hacky).

nmoroze:
I figured out the library name change bug! It was an old bug in the test that you uncovered: basically, the test is meant to detect that the fake filepath provided in that library parameter is invalid. However, the test only sets up the Yosys tool, and I forgot to set a steplist in this one. That meant that check_manifest() was failing earlier due to the other tools in the flow not being set up. Since you corrected that, this uncovered the fact that the library name was wrong and so it wasn't checking the invalid filepath it's supposed to be checking.

Kind of a bad test on my part since it's just looking for any failure, not the specific one. Not sure a clean way to correct this though.

siliconcompiler/schema.py
```diff
@@ -3219,39 +3219,59 @@ def schema_metric(cfg, step='default', index='default',group='default', ):
...
-            "cli: -metric_memory 'dfm 0 goal 10e9'",
-            "api: chip.set('metric','dfm','0','memory','real,'10e6')"],
+            "cli: -metric_exetime 'dfm 0 goal 35.3'",
+            "api: chip.set('metric','dfm','0','exetime','real','35.3')"],
         'help': """
-        Metric tracking the total memory on a per step basis, specified
-        in bytes.
+        Executable time tracks the amount of time spent by the eda
+        executable 'exe'. It does not include the siliconcompiler
+        runtime overhead or time waitig for I/O operations and
```
nmoroze, line 3255:
```suggestion
        runtime overhead or time waiting for I/O operations and
```

nmoroze (approved):

aolofsson:
Thanks for finding the root cause! It's fine.

PR #895, opened 2022-03-04
Update OpenROAD to latest 

nmoroze:
This PR updates our OpenROAD-flow-scripts submodule to the latest version, and ensures that all the scripts still work.

There was actually only one small breaking API change, but the process revealed a bug in how we were handling placement padding. From the commit notes:

- We should use the $openroad_pad_detail_place value instead of the
  $openroad_pad_global_place in sc_cts.tcl, since we are doing a
  detailed placement here. This is consistent with ORFS.
- This makes the global call to set_placement_padding in sc_apr.tcl
  problematic, because it sets the placement padding value to the global
  value and then runs a placement check. However, the default detailed
  placement padding for Sky130 is smaller than the global placement
  padding, so this check can then fail overlap checks.
- This bug also revealed a bug in the GCD skywater test, where two
  conflicting constraints being provided meant that OpenROAD didn't
  receive clock constraints.

Since this might warrant some discussion, I don't want to update the version of OpenROAD on our test runner until we think we're set to merge this (since making that change will break CI for other PRs in the meantime). Let me know once you're ready to go with this, and I can update the test runner to make sure this PR is green.

### DIFF
--- a/siliconcompiler/tools/openroad/sc_apr.tcl
+++ b/siliconcompiler/tools/openroad/sc_apr.tcl
@@ -185,10 +185,6 @@ set_dont_use $sc_dontuse
 set_wire_rc -clock  -layer $sc_clkmetal
 set_wire_rc -signal -layer $sc_rcmetal
 
-set_placement_padding -global \
-    -left $openroad_pad_global_place \
-    -right $openroad_pad_global_place
-
 ###############################
 # Source Step Script
 ###############################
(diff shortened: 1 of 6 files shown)


aolofsson:
That's great! Let' discuss Monday and merge!

nmoroze:
Server has been updated, this should be good to go!

aolofsson (approved):

PR #896, opened 2022-03-07
Fixing print_tcl to output all schema values (not just non-empty)

aolofsson:
Greatly simplifies the TCL drivers, since we don't have to check for the existence of a key. Removes a lot of annoying if/else statements.


nmoroze (approved):
Looks good, seems like a nice simplification!

Issue #865, opened 2022-02-09
Netgen LVS run failed for example design because of mismatch in number of nets


jarvis-lorry:
I've run the example design, heartbeat, with DRC/LVS enabled and get a mismatch in the number of nets, meaning that LVS fails.  I see a similar problem with other designs I've run.

Circuit 1 contains 34 devices, Circuit 2 contains 34 devices.
Circuit 1 contains 53 nets,    Circuit 2 contains 104 nets. *** MISMATCH ***

Result: Netlists do not match.
Logging to file "outputs/heartbeat.lvs.out" disabled
LVS Done.

I've attached the lvs.log.
[lvs.log](https://github.com/siliconcompiler/siliconcompiler/files/8036593/lvs.log)
.

nmoroze:
Could you please attach the outputs/heartbeat.lvs.out file? I think that will be helpful for debugging. It would also be helpful to see the Python code or CLI invocation you're using to drive SC.

I was able to get LVS clean on heartbeat using Netgen 1.5.192. I've noticed issues with different errors cropping up depending on version, so could also be worth trying to downgrade to that version and see if that helps the problem.

jarvis-lorry:
The LVS mismatches appear to come from the lack of power pins in the Verilog netlist file.
[heartbeat.lvs.out.txt](https://github.com/siliconcompiler/siliconcompiler/files/8074362/heartbeat.lvs.out.txt)

jarvis-lorry:
I'm running with Netgen 1.5.219 and I'm running with the heartbeat example.  Attached is my .py file.
[heartbeat.py.txt](https://github.com/siliconcompiler/siliconcompiler/files/8074422/heartbeat.py.txt)

jarvis-lorry:
One other question: when will DRC/LVS checks be available for PDKs other than sky130A?  Is there a reason that Magic/Netgen cannot be used with other types of PDKs?

nmoroze:
Got it, thanks for sending this over! As a workaround I'd recommend trying Netgen 1.5.192 for now, I think this issue is related to #480. We still have to investigate why this behavior is different between Netgen versions. 

To answer your other question, we don't yet have Magic runsets (the file(s) that codify the DRC rules) for the other PDKs. There may be some available somewhere (or someone could create them), but we don't currently have that on the roadmap. We'd certainly appreciate this as a contribution though!

jarvis-lorry:
[...]

Issue #480, opened 2021-09-09
Netgen LVS fails on GCD with version 1.9.193 and up


nmoroze:
Noticed that the test_gcd_skywater.py test is failing on the AWS box I'm using, and turns out it's because that machine has a newer version of Netgen. I think the failure is related to power distribution not being set up in the design, so it might be worth adding a Netgen tool option to disable checking power pins, if possible.

PR #900, opened 2022-03-08
Support new versions of KLayout, Magic, and Netgen

nmoroze:
This PR is working towards ensuring we support the newest versions of our export and signoff tools, inspired by our supported version of KLayout being bumped off the downloads page and some recent issues.

I initially thought there were some coupling issues between the three, but luckily that doesn't seem to be the case. Necessary things for updating:

KLayout:

The main thing here is we need to update the *.lyt file, since I noticed layers weren't being mapped properly for Sky130 in 0.27.x. This is a duplicate of this ORFS issue: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/299. 

Magic/Netgen:

Older versions of Netgen seemed to disregard the std. cell power pins, so we were able to do a logical connection-only LVS for our basic test. This no longer seems to be the case, and I had trouble getting newer versions of Netgen to replicate this behavior (see #865, #480). I think it might be fair to say that power needs to be routed to get a clean LVS result. I wrote up a quick floorplan + PDN to use for our Skywater130 signoff test, and decided it was generic enough that we might as well make it a floorplan.py "super-function" so that people can do quick experiments with a PDN. We can discuss this however, and I can make it a one-off if preferred.

Marking as draft, todo:
- [x] Double check that KLayout 0.27.x maps FreePDK45 still
- [x] lyt own place in schema
- [ ] Update tools on test runner, once we're otherwise good to merge (this might cause some disruption)
- [x] Update setup scripts

### DIFF
--- a/examples/gcd/gcd.v
+++ b/examples/gcd/gcd.v
@@ -9,6 +9,8 @@
 //module GcdUnit
 module gcd
 (
+  input wire vdd,
+  input wire vss,
   input  wire clk,
   input  wire [  31:0] req_msg,
   output wire req_rdy,
--- a/setup/install-klayout.sh
+++ b/setup/install-klayout.sh
@@ -6,9 +6,9 @@ cd deps
 version=$(lsb_release -sr)
 
 if [ "$version" = "18.04" ]; then
-    url=https://www.klayout.org/downloads/Ubuntu-18/klayout_0.26.12-1_amd64.deb
+    url=https://www.klayout.org/downloads/Ubuntu-18/klayout_0.27.8-1_amd64.deb
 elif [ "$version" = "20.04" ]; then
-    url=https://www.klayout.org/downloads/Ubuntu-20/klayout_0.26.12-1_amd64.deb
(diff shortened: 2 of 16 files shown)


nmoroze:
Alright, this is ready for review! As suggested, I also lumped in the lyp/lyt file cleanup task into this PR. Turns out there was already a perfect spot for lyp, so didn't have to use the escape hatch :)

After an initial review, I'll update the SC runner and re-do CI to make sure tests are green before merging.

aolofsson:
Looks good in general.

A couple of comments:

1.) Some people (me for example) put explicit supply ports in the design, but most don't. We can't make vdd/vss a requirement i the design, so I would rmove it from the gcd.v

2.) I think we need to discuss the place_pdn further. It might work for older simpler blocks (limited layers, no macros), but it will break for complex nodes and/or floorplans. We definitely need an abstraction layer above place_wire, but place_pdn is probably too high.

nmoroze:
Sounds good! 

Also, I think you were right that this PR should be split up, it's too much unrelated stuff. I've put the KLayout-related stuff in its own PR:  #910. I'll address the other changes in a separate PR and close this one in favor of that.

PR #907, opened 2022-03-08
Some show() bugfixes

### LINKED ISSUE #865: Netgen LVS run failed for example design because of mismatch in number of nets

nmoroze:
- For now, hardcode last step as export. Otherwise, this breaks when running Skywater verification steps.
- Make extra_options use 'None' sentinel instead of list, using list in kwarg can cause troubles due to Python quirk.
- Make _makecmd accept extra_options so we don't have to modify schema. This results in trouble if we want to make multiple calls to show().
- Clean up logging a bit so we don't print "None"

Closes #865.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3813,7 +3813,7 @@ class Chip:
         self.cfghistory[self.get('jobname')] = copy.deepcopy(self.cfg)
 
     ###########################################################################
-    def show(self, filename=None, extra_options=[]):
+    def show(self, filename=None, extra_options=None):
         '''
         Opens a graphical viewer for the filename provided.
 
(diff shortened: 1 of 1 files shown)


aolofsson:
Hardcoding last step as 'export' seems problematic, but the whole idea of trying to find the file without specifying the file name is a bit problematic.. I think this is fine for now. Perhaps a better user interface would be to load the gds for a named step, or to take the latest one. In the rare case that multiple design.gds are found in a job, take the one with the latest date stamp.
Thoughts? I am fine merging it if you want to defer any rework until later.

nmoroze:
Thinking ahead to packaging or stitching together multiple flows, I think this question might come up again. I figure we might want a way for each flow to specify which steps produce its final outputs: e.g. GDS comes from export, netlist comes from dfm, etc. Either that, or require a methodology where all outputs end up in one step (perhaps by making the last step a join or something).

Once we solve that, I think we can use it to come up with a good solution here. This would also be a solution for the report generation, which hardcodes the screenshot to export as well. It might be good to merge this now and defer that discussion a bit.

aolofsson:
Yes, sounds good! We should make flow modularity a priority. Merging.

aolofsson (approved):

PR #909, opened 2022-03-09
add missing boost libs for openroad

suppamax:
Signed-off-by: Massimiliano Giacometti <massimiliano.giacometti@planv.tech>


nmoroze (approved):
Thanks for the contribution!

PR #911, opened 2022-03-09
Update supported Magic and Netgen

nmoroze:
This PR's main goal is to allow SC to properly support up-to-date versions of Netgen and Magic. Prior to v.1.5.193, Netgen let you get away with performing LVS on blocks without power routed. However, this is no longer the case. This PR:

- changes our basic GCD skywater LVS test to use a floorplan with PDN, generated by our API.
- includes some changes to floorplan.py to facilitate this/fix some random things I noticed in the development process.
- adds support to sc_floorplan.tcl to insert power nets based on the 'supply' schema so we can generate a PDN without requiring these pins/nets be explicitly added to the RTL.
- and finally, bumps the officially supported versions in the tool drivers and setup scripts!


aolofsson:
Looks great. A nice set of changes,

aolofsson (approved):

PR #919, opened 2022-03-09
Create distinct asictopflow and signoffflow

nmoroze:
This PR creates two new flows:

- signoffflow: this is the optional verification component of asicflow, split into its own. It has been added to the Skywater target.
- asictopflow: this is the "physflow" that is currently hardcoded into the ZeroSoC build script. It takes in a DEF and just does a GDS stream out with any configured libraries, and also generates a netlist for passing into future LVS steps.

This sets us up nicely to begin thinking about how we want to stitch flows together. Right now, I've settled on using the 'read' schema coupled with '`chip.find_result(...)` to implement the hand-off from one to another (see changes to test_gcd_skywater.py as an example), but I think there are certainly ways this could be improved.

### DIFF
--- a/docs/_ext/dynamicgen.py
+++ b/docs/_ext/dynamicgen.py
@@ -338,9 +338,10 @@ class LibGen(DynamicGen):
     def extra_content(self, chip, modname):
         # assume same pdk for all libraries configured by this module
         libname = chip.getkeys('library')[0]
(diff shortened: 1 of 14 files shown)


aolofsson:
Looks good! I have a proposal for the modular flow to discuss tomorrow.

aolofsson (approved):

PR #926, opened 2022-03-11
Refactor tutorials to pull from examples directory

nmoroze:
I was about to drop a 75+ file changed PR and decided that would be unreasonable, so I'm splitting up my examples cleanup into more reviewable chunks. This is the first!

This PR takes example code that currently exists in the documentation and moves it into the examples/ directory. This adds 5 new examples:

- `examples/heartbeat/`: code from the Quickstart guide and the Parallel Programming tutorial (decided to put both under heartbeat, since they are both based on that design)
- `examples/heartbeat_migen/`: code from Python-based design tutorial
- `examples/gcd_chisel/`: code from Chisel tutorial
- `examples/gcd_hls/`: code from Bambu tutorial
- `examples/fibone/`: code from Bluespec tutorial

I used symlinks to let us reference these in the docs. 

My next two PRs in this group will be:
- Cleaning up existing examples
- Coming up with a nice and hopefully understandable testing methodology so we can run examples in CI, and deleting redundant tests. A test for heartbeat slipped into this PR, but I'm gonna think some more about how best to do it for all of them, since I recall early feedback was that it's a little weird-looking.


aolofsson:
Looks good!!

aolofsson:
Looks good!!

aolofsson (approved):

PR #927, opened 2022-03-11
Clean up existing examples

nmoroze:
This PR cleans up the stuff we currently have in our examples/ directory. The PR is messy cause of a lot of files being shuffled around and deleted, here is the overview:

**Examples deleted**
- `and2_openfpga/`: this one is pretty badly broken, since it relies on a Yosys hack that was removed when we did the target refactor. I figure we can delete for now, since we haven't really thought about OpenFPGA support for a while.
- `mixed-source/`: this one is a duplicate of a skipped test for mixed source support, which doesn't work yet. We already have the test as a placeholder, seems like there's no reason to keep the example around.
- `hierarchy.py`: this seems to be a placeholder for some sort of handling of hierarchy, but it uses schema keys that don't yet exist. Figured there's no reason to keep this either.
- `parallel/`: this one seemed a bit incomplete. I think it's meant to be a template, but we have some other parallel programming examples, so didn't seem worth keeping.

**Examples moved**
I made a directory called `oh_experiments/` and moved the following examples into it:
- `adder_sweep.py`
- `check_area.py`
- `feedback_loop/feedback_loop.py`
- `doe/doe.py` (renamed `parallel_adder_sweep.py`) 

This seemed like a nice way to group these tests, since they're all thematically connected in that they how you can do interesting things with SC + Python to easily investigate a library.

**Examples cleaned up**
`gcd`:
- Got rid of files that seemed unecessary: `gcd_noclock.sdc`, `gcd_golden.def.gz`, `gcd_golden.pkg.json.gz`
- Got rid of `run-skywater130.sh` since it was broken, replaced it with Python example `gcd_skywater.py` and `sky130_floorplan.py`

`blinky`: 
- Made scripts run from inside the examples/ dir itself (we should come up with a concrete standard here -- I did this to make it consistent with new heartbeat example, but didn't change gcd since I didn't want to ruin the `./examples/gcd/run.sh` smoke test from the root)
- Added Python script
- Updated shell scripts for new schema


aolofsson (approved):

aolofsson:
Great cleanup!

mithro:
Just FYI, there are is a "mixed HDL hello world" example which targets the Fomu (Lattice iCE40UP5k part) in the Fomu workshop at https://workshop.fomu.im/en/latest/mixed-hdl.html and https://github.com/im-tomu/fomu-workshop/tree/master/hdl/mixed/blink

mithro:
Also FYI, @tdene has a repository for generating adders (in both Verilog and VHDL) @ https://github.com/tdene/synth_opt_adders and a number of pre-generated adders can be found at https://github.com/tdene/adder_networks

PR #929, opened 2022-03-11
Big schema rewrite

aolofsson:
Long overdue schema.py refactor/rewrite. The schema started small, but is now enormous, calling for a more structured approach to parameter entry. A function is created that fills in some of the defaults and creates keys where missing. The use of an entry function ensure better consistency and avoids copy paste bugs that have crept in over time.  In addition, more emphasis is placed on creating common patterns can be used, with for loops working on lists.

Schema changes:
-units --> unit (we use singular in all other keys)
-sc --> software (further separate schema from software)
-flatten skip dictionary --> to small to warrant a top level group, simple booleans should be flat
-metricoff (added for summary, since metric list is getting very long)
-wofff-->warningofff (clarity)
-testmodule (critical for simulation, separate path)
-Adding per author contact information in package
-Adding metric key to checklist report/waiver parameter
-adding step/index to exluce cells parameter
-removed the max/mix flostatus parameters, not needed
-shortened environment to env (Consistency)
-removing tapcell/tapmax (will "always" be a complex PDK file per tool)
-adding edif, hspice, spectre, vhdl netlist parameters to library
-adding load cell to library cells
-adding more power mode metrics


siliconcompiler/core.py
```diff
@@ -990,18 +989,30 @@ def _search(self, cfg, keypath, *args, field='value', mode='get', clobber=True):
                             else:
                                 self.logger.error(f'{field} must be set to boolean.')
                                 self.error = 1
-                        elif field in ('filehash', 'date', 'author', 'signature'):
-                            if isinstance(val, list):
+                        elif field in ('hashalgo', 'scope', 'require','type',
+                                       'shorthelp', 'switch', 'help'):
```
WRansohoff, line 993: Could these small sets of type-convertible parameters be stored somewhere in `schema.py`, or are they unlikely to be reused?

tests/core/test_read_manifest.py
```diff
@@ -21,7 +24,36 @@ def test_read_manifest_fields():
...
+    expected['version']['software']['defvalue'] = _metadata.version
+    expected['version']['software']['value'] = _metadata.version
+
+    assert actual == expected
+
 #########################
 if __name__ == "__main__":
     from tests.fixtures import datadir
-    test_read_manifest(datadir(__file__))
+    #test_read_manifest(datadir(__file__))
```
WRansohoff, line 58:
```suggestion
```

Delete?

WRansohoff (approved):
I haven't really looked at the big schema.py change yet, but we discussed the meat of that earlier.

Besides the `clobber` flag uncertainty in the `_search` function, which we should probably resolve before the next release, I think this looks good. Just a few non-blocking comments.

siliconcompiler/core.py
aolofsson: I thought about it, but there are some things that are simply too coupled to the set/get routine. Let's leave it like this.

nmoroze:
I fixed dynamic docs gen in #938. The PR is using this branch as a base, so merging that PR will add the changes to this one.

I noticed that something isn't working right with RST being parsed in help strings, but I can address that in a future PR after this is merged.

siliconcompiler/core.py
```diff
@@ -3406,7 +3416,7 @@ def _runtask(self, step, index, active, error):
         veropt = self.get('eda', tool, 'vswitch')
         exe = self._getexe(tool)
         version = None
-        if (veropt is not None) and (exe is not None):
+        if veropt and (exe is not None):
```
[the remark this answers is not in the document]
aolofsson: Intent was to fix a bug. veropt is  list, hecking for empty list.

```python
a = []
if a:
    print("not empty")
if a is not None:
    print("not EMPTY (INCORRECT)")

File: siliconcompiler/core.py [hunk 4]
```diff
@@ -972,14 +971,14 @@ def _search(self, cfg, keypath, *args, field='value', mode='get', clobber=True):
...
-                    selval = cfg[param]['value']
+                # TODO: fix clobber!!
+                selval = cfg[param]['value']
                 # updating values
                 if cfg[param]['lock'] == "true":
                     self.logger.debug("Ignoring {mode}{} to [{keypath}]. Lock bit is set.")
                 elif (mode == 'set'):
+                    #print(keypath, "**", param, field, val, isinstance(val, list))
+                    #TODO: line below is broken, should check for field
                     if (selval in empty) | clobber:
```
nmoroze, line 982: Can we resolve this TODO now, or do you think it will require more thought? I think line should just be:

`if (field != 'value') or (selval in empty) or clobber:`

nmoroze (approved):
Looks good to me!

tests/core/test_read_manifest.py
```diff
@@ -21,7 +24,36 @@ def test_read_manifest_fields():
     chip2.read_manifest('tmp.json')
     assert chip2.get('source', field='copy') is False
 
+def test_read_defaults(datadir):
+    '''Make sure read/write operaton doesn't modify manifest'''
+
+    DEBUG = False
```
nmoroze, line 30: Delete this? looks like it's unused.

tests/core/test_read_manifest.py
nmoroze: I think it's best to select test via pytest instead of uncommenting/recommenting, even if it's a bit more verbose:

`pytest tests/core/test_read_manifest.py::test_read_manifest -s` vs `pytest tests/core/test_read_manifest.py::test_read_defaults -s`

tests/core/test_read_manifest.py
aolofsson: sort of. done

siliconcompiler/core.py
aolofsson: Please check post merge.

tests/core/test_read_manifest.py
aolofsson: Agreed

siliconcompiler/core.py
```diff
@@ -3394,10 +3404,10 @@ def _runtask(self, step, index, active, error):
                 os.environ[item] = ':'.join(license_file)
 
         # Tool-specific environment variables for this task.
-        if (step in self.getkeys('eda', tool, 'environment')) and \
-           (index in self.getkeys('eda', tool, 'environment', step)):
-            for item in self.getkeys('eda', tool, 'environment', step, index):
-                os.environ[item] = self.get('eda', tool, 'environment', step, index, item)
+        if (step in self.getkeys('eda', tool, 'env')) and \
```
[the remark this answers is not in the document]
aolofsson: I am for it. 'var' is shorter and equally clear.
Can you do the change post merge.

Issue #930, opened 2022-03-11
OpenROAD sc_display.tcl is inconsistent with schema


arlpetergadfort:
https://github.com/siliconcompiler/siliconcompiler/blob/a202264ed29cbb960d85fad3be29cd071d4e66b2/siliconcompiler/tools/openroad/sc_display.tcl#L37

Should have stackup as well.

Additionally, it would be nice if the liberty files were loaded along with the sdc file.

nmoroze:
I think we might actually be able to just get rid of sc_display.tcl and reuse sc_apr.tcl setup here. I'll do some experimenting.

nmoroze:
Opened a PR implementing this. One thing to keep in mind is that show spawns a new step, meaning any SDC files fed in on a per-step basis using the 'read', 'sdc' schema won't be loaded. However, things fed in via 'constraints' will be. I'm not sure if this works for your use case, I also think this subtle distinction might trip people up.

nmoroze:
#933 is merged, closing.

PR #933, opened 2022-03-11
Use sc_apr.tcl for showing file with OpenROAD

nmoroze:
RE #930, sc_display.tcl is broken. This PR is a proposal to reuse sc_apr.tcl due to common setup. I don't know if I like the extra nesting for the logic below, we could also considering adding another level of TCL that does setup that is reused by both sc_apr.tcl and sc_display.tcl.


aolofsson (approved):
I think it's ok, but going forward we should consider having some common TCL code for all tools to reduce the copy paste. show and apr are fundamentally different so not sure how much we can share the code? Not a big issue.

Issue #906, opened 2022-03-08
Consider providing a conda environment.yml which creates an environment including all needed open source EDA tools


mithro:
Thanks to the work by @proppy tools like OpenROAD, the SkyWater PDK, Magic, Netgen, etc are all now packaged in conda @ https://github.com/hdl/conda-eda -- There are even packages for cross compilers for things like RISC-V architecture which can be found @ https://github.com/hdl/conda-compilers

This means that silicon compiler could provide a conda environment.yml file (https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#create-env-file-manually) which installs all the requirements needed to run Silicon Compiler.

proppy:
additionally maybe also consider packaging siliconcompiler w/ a [recipe](https://docs.conda.io/projects/conda-build/en/latest/concepts/recipe.html) that depends on the external tools listed here https://docs.siliconcompiler.com/en/latest/user_guide/installation.html#external-tools, so that people could do `conda install siliconcompiler` to get a full blown installation.

aolofsson:
Interesting! There are quite a lot of tricky dependancies between the tool versions that we are running into. We'll take a look to see how much work it would be. This does require peple to use conda, which is a barrier to some. (but perhaps not a big one?)

mithro:
@aolofsson -- Conda is pretty common and well accepted in the ML and scientific worlds. Like all packaging systems it has some people who hate it.

mithro:
FYI - @proppy

mithro:
Please do report the issue on the upstream repositories!

I have asked @proppy to look into what is going on here. I believe it is because these packages are still very new (less than a week old) and need a little bit more work. The only way things will get better is to have users finding and reporting these issues!

proppy:
@WRansohoff some of those package do depend on conda-forge you confirm if/this also happen when using a conda-forge based environment? (either by using https://github.com/conda-forge/miniforge#miniforge3 or folllowing https://conda-forge.org/docs/user/introduction.html#how-can-i-install-packages-from-conda-forge)

WRansohoff:
Ah, I must have missed that dependency. Using conda-forge to install the packages seems to work, e.g.:

    conda install -c conda-forge --channel "LiteX-Hub" openroad

Thank you for the clarification, and for packaging open-source EDA tools like this!

[...]

PR #934, opened 2022-03-12
Initial Conda environment setup file.

WRansohoff:
Per #906, we can use the LiteX Conda recipes to install almost all of the open-source EDA tools which are required by the default `asicflow`. It turned out to be a very simple setup file thanks to the [conda-eda](https://github.com/hdl/conda-eda) recipes.

Leaving this as a draft for now, with a few notes:

* When I used the most recent version of Surelog, it seemed to look for some input files in the directory which the `sc` command was run from, rather than the task directory where the `subprocess.run` call is made. That might bear further investigation, but for now I've tagged an older version.
* We'll need to publish a new version of `siliconcompiler` to pypi before this environment will work. Because the OpenROAD Conda recipe is fairly new, the Conda channel didn't have a version that was old enough to support the published v0.7.0 tcl scripts. The current version of OpenROAD does work with the current `main` branch of `siliconcompiler`, though.
* The `third_party/pdks/` directory is required to run the full `asicflow` with freepdk45/skywater130/etc, and we don't package those files in the pip wheels. So for this environment to work, users will also need to clone the repository and set `$SCPATH` to `[...]/siliconcompiler/siliconcompiler/`.
* KLayout must be installed manually for the export step.


aolofsson:
Looks good!

aolofsson (approved):

setup/conda/README.md
```diff
@@ -0,0 +1,21 @@
+# Environment and Tool Setup Using Conda
+
+The LiteX project [provides recipes](https://github.com/hdl/conda-eda) for installing many core open-source EDA tools using the Anaconda / Miniconda Python package management system.
```
proppy, line 3: nit: technically, this is the `conda-eda` project under the `hdl` organization., the channel is still named `Litex-Hub` for legacy quota reasons.

setup/conda/conda_env.yml
```diff
@@ -0,0 +1,14 @@
+name: sc_env
```
proppy, line 1: nit: maybe simply name this `environment.yml` to follow conda convention https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#create-env-file-manually

setup/conda/conda_env.yml
```diff
@@ -0,0 +1,14 @@
...
+channels:
+  - conda-forge
+  - LiteX-Hub
```
proppy, line 4: LiteX-Hub needs to be first otherwise you'll get the `netgen` package from `conda-forge` which is sadly totally unrelated: https://anaconda.org/conda-forge/netgen

setup/conda/conda_env.yml
```diff
@@ -0,0 +1,14 @@
...
+dependencies:
+  - pip
+  - pip:
+    - siliconcompiler
+
+  - surelog==0.0_3379_g484687d13
+  - openroad
+  - magic
+  - netgen
+  - yosys
```
proppy, line 14: what about open_pdks?

proppy (commented):
sorry for the late comments, do you want me to send a follow-up pull request?

WRansohoff:
Thank you for the feedback!

No worries, I submitted a quick change with most of your suggestions, and we can investigate the open_pdks recipe as an alternative to potentially packaging those files with siliconcompiler.

PR #935, opened 2022-03-12
Create tests for examples and remove redundant tests

nmoroze:
This PR creates tests under tests/examples/ for running examples in CI. I found a way to factor out the messy stuff from the heartbeat example into a common `setup_example_test` fixture. Now we can write these tests very concisely:

```python
def test_heartbeat(setup_example_test):
     setup_example_test('heartbeat')
     import heartbeat
     heartbeat.build()
     
     # assertions here
```

I deleted the following existing tests:
- test_asicflow: covered by test_gcd
- test_bambu: covered by test_gcd_hls
- test_bluespec: covered by test_fibone
- test_chisel: covered by test_gcd_chisel
- test_fpgaflow: covered by test_blinky
- test_gcd_skywater: covered by test_gcd
- test_gcd_cli: covered by test_gcd

By the way, do you still use examples/gcd/run.sh as a smoke test? I switched it over to use paths that are relative to its parented directory (like the other examples), but included a little bash magic to make sure it can still be run from root (or anywhere, for that matter).


aolofsson:
Neat cleanup!

aolofsson (approved):

PR #937, opened 2022-03-14
Use pseudoterminal to run tools when bkpt enabled

nmoroze:
Using our usual subprocess scheme seems to break terminal echo for some tools when run interactively through SC. This PR switches things over to using a pseudo-terminal via [pty](https://docs.python.org/3/library/pty.html) to run them instead. This seems to solve the problem.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -33,6 +33,7 @@ import getpass
 import distro
 import netifaces
 import webbrowser
+import pty
 from jinja2 import Environment, FileSystemLoader
 from pathlib import Path
 from timeit import default_timer as timer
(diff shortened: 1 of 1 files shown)


aolofsson:
Looks reasonable, but as discussed the bkpt feature is still limited. It will almost certainly break when jobs are running in parallel and it's not something we want to support. There might be value in serializing the graph for bringup.

aolofsson:
Looks reasonable, but as discussed the bkpt feature is still limited. It will almost certainly break when jobs are running in parallel and it's not something we want to support. There might be value in serializing the graph for bringup.

aolofsson (approved):

PR #938, opened 2022-03-14
Docs gen fixes for big schema refactor

nmoroze:
This PR should fix up docgen. It will merge back into the primary schema refactor PR.

Fixes:
- We used to rely on the presence of the 'value' key to determine whether we should display something, but that no longer works since we set 'value' for everything. Use truthiness of 'value' instead.
- recordgen needed updating due to 'job' key being removed


nmoroze:
I think the typo fixes broke the new read manifest test. Do you have a quick way to regenerate defaults.json?

aolofsson:
>

aolofsson (commented):
Looks good!

aolofsson (approved):

PR #939, opened 2022-03-14
Tweaks to replay script generation

nmoroze:
This PR fixes tweaks 3 things surrounding replay script generation:
- Fix issue where grouped parameters would become unquoted (i.e. `-param "a b c"` would become `-param a b c` in script, now it stays the same).
- Add code to set environment variables in script
- Add Windows support: .bat instead of .sh, no `#!/bin/bash` header, set env variables appropriately for each platform

I haven't yet tested the Windows support, but I did add a replay test based on Surelog (so that it catches all platforms), and I can kick off a wheel build to try it out.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -4124,9 +4124,20 @@ class Chip:
             for option in runtime_options(self):
                 cmdlist.extend(shlex.split(option, posix=is_posix))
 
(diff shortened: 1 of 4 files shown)


aolofsson:
Nice fixes.

aolofsson (approved):

PR #942, opened 2022-03-15
Post-schema refactor fixes

nmoroze:
This PR:
- Implements the TODO in _search() mentioned in code review
- Fixes up RST parsing of schema help strings

For the latter, I decided move to processing schelp using the Python standard "trim()" function for preprocessing docstrings that we were already using for the flows/pdks/libs/etc. documentation. The previous regex approach was stripping away newlines, which made it impossible to do fancier RST formatting, such as lists.

The new approach strips off a calculated amount of leading spaces to preserve additional indentation (for inserting things like codeblocks), and preserves newlines. A consequence of this change is that there will be newlines in the help string, but I think this is a worthwhile tradeoff, since it makes the actual docs look much nicer:

## Before
Screenshot from 2022-03-15 11-33-55

## After
Screenshot from 2022-03-15 11-39-02

### DIFF
--- a/docs/_ext/dynamicgen.py
+++ b/docs/_ext/dynamicgen.py
@@ -19,7 +19,7 @@ import subprocess
 
 from common import *
 
-import siliconcompiler
+from siliconcompiler import utils
 
 #############
 # Helpers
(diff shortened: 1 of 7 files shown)


aolofsson:
This is great. Thank you fro the clean up! Yes, this is much better. At some point, we may make a pass through the schema and expand some of the long descriptions and this is great preparation.

aolofsson (approved):

PR #944, opened 2022-03-15
[ghdl] remove assertions in generated verilog code

suppamax:
Signed-off-by: Massimiliano Giacometti <massimiliano.giacometti@planv.tech>

See #917

### DIFF
--- a/.github/workflows/on_push_tests.yml
+++ b/.github/workflows/on_push_tests.yml
@@ -29,8 +29,9 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          sudo apt-get update
-          sudo apt-get install graphviz
+          # Workaround for Github Actions issue
+          # https://github.com/actions/virtual-environments/issues/5237
+          (sudo apt-get -y update || true) && sudo apt-get install graphviz
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2


suppamax:
ghdl-yosys-plugin added

nmoroze:
Thanks a ton for the contribution! I'm assuming since you switched to the Yosys plugin that ghdl still has no option to output to a file?

I think it would be a slightly better fit for our general methodology if we could get ghdl itself to output Verilog, so I can open an issue on their repo making that feature request. If they're not interested in adding that feature, I'd be okay with going through the Yosys plugin, although in that case I think we'd want to drive it through a TCL script like how we did in `tools/yosys/`, rather than stuffing the options into the CLI and using `shell=True` in the core subprocess call.

suppamax:
I understand your point.
Yes, as far as I understand `--out` simply writes the code to stdout
https://github.com/ghdl/ghdl/blob/987c5b78ab65c897d9e5757abc807df0fe264144/src/synth/netlists-dump.adb#L352

I've already submitted a [feature request](https://github.com/ghdl/ghdl/issues/2010):  obviously, feel free to extend or correct it if you like.

suppamax:
`ghdl --synth` outputs the verilog (and the verilog only) to STDOUT, but in the subprocess you are merging STDOUT and STDERR, so you get the extra messages which you have to filter out.

I can make an alternative version which uses tcl and no `shell=True` if you want.

PR #947, opened 2022-03-16
Implement flexible version checking and make 'vercheck' default to True

nmoroze:
This PR implements a proposal for flexible version checking that allows for comparisons. I've implemented this as a baseline for discussion before I go ahead and modify more tool drivers (although I implemented it in a way that somewhat maintains backwards compatibility).

Here is the approach:
- `parse_version()` stays the same
- Tools may define a new function, `normalize_version()`. This is intended to take in a version string produced by `parse_version()`, and return a comparable value such that if `ver_a` is newer than `ver_b`, `normalize_version(ver_a) > normalize_version(ver_b)`
  - As a methodology thing, I think it makes the most sense for tools to return tuples of int. The way tuple comparison is implemented lets us do semantic versioning-style comparisons easily. For example, `(2, 0) > (1, 5)` is true, and `(1, 100) > (1, 95)` is true. 
  - If a tool doesn't have a `normalize_version()`, the result of `parse_version()` is used but the specifiers are restricted to exact match or != (backwards compatibility)
- In the 'eda', 'version' field, drivers provide a list of version specifiers that are either a bare version, or one of `==, !=, >=, >` followed by 0 or more whitespace characters, followed by a version.
  - A bare version translates to an exact match (backwards compatibility)
  - Everything in the list is a logical AND (i.e. version needs to match all specifiers for check to be ok)

The main thing that I want to discuss is whether to make each entry in the list a logical "and" or "or". I can see value in both patterns:
- and is good for excluding a single bad version: `>=1.0.0 && != 1.05` 
- or is good for specifying a list of known good versions: `==0.6 || ==0.7`

I implemented this as "and" since it matches the Python [PEP-440](https://peps.python.org/pep-0440/) standard for lists of version specifiers, but I've been thinking about ways we could potentially do both.

### DIFF
--- a/docs/user_guide/tools.rst
+++ b/docs/user_guide/tools.rst
@@ -35,17 +35,24 @@ SiliconCompiler execution depends on implementing adapter code "drivers" for eac
      - run()
      - yes
 
+   * - **normalize_version**
+     - Returns executable version
+     - tool version
+     - normalized version
+     - run()
+     - no
+
    * - **pre_process**
      - Pre-executable logic
      - chip
-     - chip
+     - n/a
      - run()
      - no
 
    * - **post_process**
      - Post-executable logic
      - chip
-     - chip
+     - exit code
      - run()
      - yes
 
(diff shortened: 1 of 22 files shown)


nmoroze:
As discussed, I added a commit that implements version checking using the standardized [packaging](https://github.com/pypa/packaging) library that corresponds with [PEP-440](https://peps.python.org/pep-0440/). 

I think I'm in favor of this approach. The main pro I see is that it turns out many of the open tools actually do provide versions as valid PEP-440 versions. This means that for these tools, we can eliminate `normalize_version()` entirely, simplifying the tool wrappers a good bit. It's also nice that we can point to an existing standard, although in some ways it does feel a bit odd to apply a standard to version #s that are defined in a way that is not aware of this standard. 

On that note, an example of a pitfall is that it's possible that tool versions may parse as-is but not in ways that are always intuitive. For example, Yosys reports versions like `0.13+15` or `0.13+3`, and the "packaging" library will evaluate `0.13+15` as lower than `0.13+3`. Therefore, someone could get away with not implementing a `normalize_version()`, but might get unintuitive behavior unless they did so and change the `+` to a `.` or `-`.

Another possible pitfall is that since "packaging" exposes the whole standard, we have to think about what happens when users try to use more complex features of version specifiers, such as wildcards and "compatible" version matching (`~=`). I think this should be okay for any tools that don't require a `normalize_version()`.  However, it's a little more complex when they do. In this case, the tool driver author would have to keep in mind that along with a real version like `x.y.z`, a user might also specify `x.y.*` -- e.g. they may have to ensure that the driver doesn't assume any particular length to the version string.

A couple other things that need discussion:
- Do we want to support the `===` escape-hatch operator from the standard? This could be useful for tools without versioning schemes (for checking commit hashes and things like that). However, support for this will be deprecated in the packaging library and removed soon, so we'd probably have to add custom logic. 
- How strict should we be in specifying version #s for the open tools? I'm thinking we can either do a loose approach of `>=` our oldest known working version, or a more conservative approach of capping using the "compatible release" specifiers (e.g. `~=0.27.6`, which supports all KLayout versions >= 0.27.6, but none that are 0.28.x).

nmoroze:
This PR should be ready for final review/merge! 

Final notes:
- We're going all in on the PEP-440-based scheme. We are not bothering to support the `===` escape hatch. For tools without valid versions you can just not add anything to 'version'. 
- For now, all our default version specifiers will be `>=`.

**Note:** this PR switches version checking to be enabled by default. We can remove that if you think it might be too disruptive, but I think it would be a good idea to make this change eventually.

aolofsson (approved):

aolofsson:
Looks fantastic! Picking an approach that serves 365,916 Python packages is a good choice. :smile:

PR #952, opened 2022-03-18
Fix wheels build

nmoroze:
This PR should get our daily wheels build back to green. I had to make a few changes to fix things I broke on Windows and macOS with my replay script change, and also some things to accommodate Github/infrastructure-related changes.

Manual run to verify: https://github.com/siliconcompiler/siliconcompiler/actions/runs/2005918393


aolofsson (commented):
Some concern with the amount of platform dependent code. Wondering if we can leverage the fact that we don't actually support interactive mode in windows?

PR #953, opened 2022-03-19
Small fixes to setup scripts

tommarin:
I encountered a few issues using the setup scripts in my fresh Ubuntu 20.04 VM, so I would like to share my fixes. I see that using conda is the most convenient solution, but these scripts can still be useful to someone who prefers a traditional approach.

### DIFF
--- a/setup/install-openroad-rhel7.sh
+++ b/setup/install-openroad-rhel7.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-scr_path=$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)
+src_path=$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)
 
 sudo yum groupinstall -y 'Development Tools'
 sudo yum install -y pkgconfig bzip2 openssl-devel pth libatomic bison flex readline-devel gawk libffi-devel git graphviz zlib-devel wget
@@ -109,7 +109,7 @@ sudo ldconfig
 cd -
 
 # Install OpenROAD tools.
-cd ${scr_path}/..
+cd ${src_path}/..
 git submodule update --init --recursive third_party/tools/openroad
 cd third_party/tools/openroad
 ./build_openroad.sh -o
--- a/setup/install-openroad-rhel7.sh
+++ b/setup/install-openroad-rhel7.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+scr_path=$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)
+
 sudo yum groupinstall -y 'Development Tools'
 sudo yum install -y pkgconfig bzip2 openssl-devel pth libatomic bison flex readline-devel gawk libffi-devel git graphviz zlib-devel wget
 sudo yum install -y qt5-qt3d-devel
@@ -106,11 +108,8 @@ sudo ln -s /usr/bin/tclsh8.6 /usr/bin/tclsh
 sudo ldconfig
 cd -
 
-
-cd ..
-
 # Install OpenROAD tools.
-cd ..
+cd ${scr_path}/..
 git submodule update --init --recursive third_party/tools/openroad
 cd third_party/tools/openroad
 ./build_openroad.sh -o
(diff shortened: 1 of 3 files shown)


setup/install-openroad-rhel7.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
+scr_path=$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)
+
```
WRansohoff, line 3:
```suggestion
```

This looks like a handy shell construction, but we would like to keep the setup shell scripts easy to read if it's not required.

setup/install-openroad-rhel7.sh
```diff
@@ -106,11 +108,8 @@ sudo ln -s /usr/bin/tclsh8.6 /usr/bin/tclsh
 sudo ldconfig
 cd -
 
-
-cd ..
-
 # Install OpenROAD tools.
-cd ..
+cd ${scr_path}/..
```
WRansohoff, line 112:
```suggestion
cd ../..
```
We should still combine the two `cd ..` steps to match the earlier `cd tcl/unix` change, though. Thank you for cleaning that up.

setup/install-openroad.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/sh
+scr_path=$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)
+
```
WRansohoff, line 3:
```suggestion
```

Same as above

setup/install-openroad.sh
```diff
@@ -23,7 +25,7 @@ cd -
 sudo ln -s /usr/bin/python3 /usr/bin/python
 sudo ln -s /usr/local/lib/libtcl8.6.so /usr/local/lib/libtcl.so
 
-cd ..
+cd ${scr_path}/..
```
WRansohoff, line 28:
```suggestion
cd ..
```

Same as above

WRansohoff (commented):
Thank you for submitting fixes for the setup issues that you encountered!

Did you run into any particular problems that required the use of the new 'source path' shell variable? We'd like to avoid complex shell constructions to keep the scripts simple and legible for as many users as possible, but we would be open to making things less dependent on relative paths if they are causing problems.

The KLayout setup changes look great, I think we can merge this after resolving the openroad setup scripts.

tommarin:
Hello,
Thank you for the review.
The only reason I went with that solution instead of using `../..` is that the script will not work if launched from outside the setup folder (e.g. `./setup/install-openroad.sh` from siliconcompiler base folder), while other scripts still work.
If that is not considered an issue because the user is supposed to run the scripts from that folder specifically, it is OK for me to revert back to the simpler solution.
Also, there are different options for identifying the script folder, I just went with the more portable, but others are definitely more readable (like using `realpath`) :)

WRansohoff:
That's a good point, ideally we should support running the scripts from any directory.

We'll discuss the trade-offs, and decide whether to merge this as-is or revert the path improvements soon. It seems like a reasonable addition, maybe with an explanatory comment, but I don't want to speak for everyone.

Thanks again for the contribution!

setup/install-openroad-rhel7.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
+scr_path=$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)
```
WRansohoff, line 2:
```suggestion
src_path=$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)
```

setup/install-openroad-rhel7.sh
WRansohoff, line 112:
```suggestion
cd ${src_path}/..
```

setup/install-openroad.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/sh
+scr_path=$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)
```
WRansohoff, line 2:
```suggestion
src_path=$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)
```

setup/install-openroad.sh
WRansohoff, line 28:
```suggestion
cd ${src_path}/..
```

WRansohoff (approved):
Okay, we talked about it and decided to merge in all of these changes. I'd just like to rename `scr_path` to `src_path`.

Thanks again for submitting these fixes!

tommarin:
Makes sense, the previous name was a bit misleading. Thank you!

PR #954, opened 2022-03-21
Optimizing runtask  

aolofsson:
- The runtask was not behaving nicely for large N. The min function was taking a tone of time due to the read_manifest step. During the runtask we don't need the heave checking that is available in the API, we can just access the dictionary directly. 
- Normalizing runtask sleep to not waste resources when tons of threads are running (hundreds).
- Removing read_manifest for critical merge steps in favor of dict copy.
- Normalizing openroad thread count based on task level parallelism

NOTE: Some of this work will get replaced by Noah's new run scheduler, but we should still merge the optimization method as a place holder. No need for read_manifest when we can just copy the dict.

### DIFF
--- a/examples/benchmark/benchmark.py
+++ b/examples/benchmark/benchmark.py
@@ -57,7 +57,7 @@ def main():
             walltime = round((wall_end - wall_start),2)
             results[design][n] = walltime
             with open(f"results.txt", 'w') as f:
-                print(results, file=f)
+                f.write(results)
 
 if __name__ == '__main__':
     main()
--- a/examples/benchmark/benchmark.py
+++ b/examples/benchmark/benchmark.py
@@ -12,9 +12,7 @@ def main():
 
     examples = [{'heartbeat' : 1000},
                 {'picorv32' : 1000},
-                {'aes' : 1500},]
-
-    examples = [{'heartbeat' : 1000}]
+                {'aes' : 1500}]
 
     results = {}
 
@@ -25,13 +23,13 @@ def main():
         rootdir = os.path.join(parent, design)
         results[design] = {}
         for n in ['1','2','4','8','16']:
-        for n in ['4']
             wall_start = time.time()
             chip = siliconcompiler.Chip(design)
             chip.set('jobname', f"job{n}")
             chip.load_target('skywater130_demo')
             chip.set('relax', True)
             chip.set('quiet', True)
+            chip.set('remote',False)
 
             # load dsign
             chip.add('source', os.path.join(rootdir, f"{design}.v"))
@@ -43,7 +41,6 @@ def main():
             chip.set('flowarg', 'syn_np', n)
             chip.set('flowarg', 'place_np', n)
             chip.set('flowarg', 'cts_np', n)
-            chip.set('flowarg', 'route_np', n)
             chip.load_flow('asicflow')
             chip.set('flow', 'asicflow')
 
@@ -59,9 +56,8 @@ def main():
             wall_end = time.time()
             walltime = round((wall_end - wall_start),2)
             results[design][n] = walltime
-            chip.write_manifest("tmp.tcl")
-
-    print(results)
+            with open(f"results.txt", 'w') as f:
+                print(results, file=f)
 
 if __name__ == '__main__':
     main()
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3898,6 +3898,7 @@ class Chip:
         # successfully. Inspired by: https://stackoverflow.com/a/70029046.
 
         os.listdir(os.path.dirname(lastdir))
+
         lastcfg = f"{lastdir}/outputs/{self.get('design')}.pkg.json"
         if os.path.isfile(lastcfg):
             last_step_failed = False
@@ -3912,6 +3913,8 @@ class Chip:
                     taskdir = self._getworkdir(step=step, index=index)
                     os.listdir(os.path.dirname(taskdir))
                     cfg = f"{taskdir}/outputs/{self.get('design')}.pkg.json"
+                    if not os.path.isfile(cfg):
+                        break
                     with open(cfg, 'r') as f:
                         localcfg = json.load(f)
                     self.cfg['metric'][step][index] = copy.deepcopy(localcfg['metric'][step][index])
@@ -3940,6 +3943,7 @@ class Chip:
                 f'See logs in {stepdir} for error details.')
             raise SiliconCompilerError(f'Run() failed on step {failed_step}! '
                 f'See logs in {stepdir} for error details.')
+
 
         # Store run in history
         self.record_history()
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3432,7 +3432,9 @@ class Chip:
             self.logger.error(f'No inputs selected after running {tool}')
             self._haltstep(step, index, active)
 
-        self.set('flowstatus', step, index, 'select', sel_inputs)
+        #TODO: This should not be needed
+        if step != 'import':
+            self.set('flowstatus', step, index, 'select', sel_inputs)
 
         ##################
         # 9. Copy (link) output data from previous steps
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3655,7 +3655,7 @@ class Chip:
         self.set('arg', 'step', None, clobber=True)
         self.set('arg', 'index', None, clobber=True)
 
-        self.write_manifest(f"outputs/{design}.pkg.json")
+        self.write_manifest(os.path.join("outputs", f"{design}.pkg.json"))
 
         ##################
         # 25. Clean up non-essential files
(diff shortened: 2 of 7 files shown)


siliconcompiler/core.py
```diff
@@ -3369,12 +3374,17 @@ def _runtask(self, step, index, active, error):
...
             for in_step, in_index in self.get('flowgraph', flow, step, index, 'input'):
                 index_error = error[in_step + in_index]
                 self.set('flowstatus', in_step, in_index, 'error', index_error)
                 if not index_error:
                     cfgfile = f"../../../{in_job}/{in_step}/{in_index}/outputs/{design}.pkg.json"
-                    self.read_manifest(cfgfile, clobber=False)
+                    with open(cfgfile, 'r') as f:
+                        localcfg = json.load(f)
+                    # quick copy of task history
+                    self.cfg['metric'][in_step][in_index] = copy.deepcopy(localcfg['metric'][in_step][in_index])
```
nmoroze, line 3387: I think this might be missing some stuff we need to carry forward:

- `'eda', ... 'report'`
- `'record', ...`
- `'flowstatus', ...` (less sure about this one, but is that the reason maybe it seemed like you needed to hack up some stuff related to flowstatus elsewhere?)

siliconcompiler/core.py
```diff
@@ -3889,24 +3901,41 @@ def run(self):
...
+                    taskdir = self._getworkdir(step=step, index=index)
+                    os.listdir(os.path.dirname(taskdir))
+                    cfg = f"{taskdir}/outputs/{self.get('design')}.pkg.json"
+                    if not os.path.isfile(cfg):
+                        break
+                    with open(cfg, 'r') as f:
+                        localcfg = json.load(f)
+                    self.cfg['metric'][step][index] = copy.deepcopy(localcfg['metric'][step][index])
+                    self.cfg['flowstatus'][step][index] = copy.deepcopy(localcfg['flowstatus'][step][index])
+
```
nmoroze, line 3938: Same comment as the `read_manifest()` replacement above, I think this misses records/eda reports (but has flowstatus).

Maybe we should refactor to some sort of `_quick_merge_manifest()` helper or something that we use in both places, and is complete?

siliconcompiler/core.py
```diff
@@ -3889,24 +3901,41 @@ def run(self):
...
-            # successfully. Inspired by: https://stackoverflow.com/a/70029046.
-            os.listdir(os.path.dirname(lastdir))
-
-            lastcfg = f"{lastdir}/outputs/{self.get('design')}.pkg.json"
-            if os.path.isfile(lastcfg):
-                last_step_failed = False
-                local_dir = self.get('dir')
-                self.read_manifest(lastcfg, clobber=True, clear=True)
-                self.set('dir', local_dir)
+        lastdir = self._getworkdir(step=laststep, index=index)
```
nmoroze, line 3908: Be careful here, I think you might want to hard code `index` to 0. I think the most recent reference to 'index' comes in the for loop that sets up multiprocessing, so the value could end up being something weird.

siliconcompiler/core.py
```diff
@@ -3369,12 +3374,17 @@ def _runtask(self, step, index, active, error):
 
         all_inputs = []
         if not self.get('remote'):
+            #copy over status from
+
```
WRansohoff, line 3378:
```suggestion
```

Is this still needed?

siliconcompiler/core.py
```diff
@@ -3889,24 +3901,41 @@ def run(self):
...
+        lastcfg = f"{lastdir}/outputs/{self.get('design')}.pkg.json"
+        if os.path.isfile(lastcfg):
+            last_step_failed = False
+            local_dir = self.get('dir')
+            self.read_manifest(lastcfg, clobber=True, clear=True)
+            self.set('dir', local_dir)
+
+        #2. Read in flowstatus and metrics from all directories
+        for step in self.getkeys('flowgraph', flow):
+            for index in self.getkeys('flowgraph', flow, step):
```
nmoroze, line 3927: Is there a need to iterate through all steps, or does iterating through each index of `laststep` work (like in the previous version of the logic)?

siliconcompiler/core.py
```diff
@@ -3889,24 +3901,41 @@ def run(self):
...
+                if step in steplist and index in indexlist[step]:
```
nmoroze, line 3928: Nitpick: I think this can be omitted by iterating over just what you need -

```python
for step in steplist:
  for index in indexlist[step]:
    ...
```

siliconcompiler/core.py
```diff
@@ -3639,11 +3651,11 @@ def read(fd):
 
         ##################
         # 24. Save a successful manifest
-        self.set('flowstatus', step, str(index), 'error', 0)
+        self.set('flowstatus', step, index, 'error', 0)
         self.set('arg', 'step', None, clobber=True)
         self.set('arg', 'index', None, clobber=True)
 
-        self.write_manifest("outputs/" + self.get('design') +'.pkg.json')
+        self.write_manifest(f"outputs/{design}.pkg.json")
```
WRansohoff, line 3658:
```suggestion
        self.write_manifest(os.path.join("outputs", f"{design}.pkg.json"))
```

While we're fixing path styling, it might be a good idea to use the platform-independent `os.path.join`.

siliconcompiler/tools/openroad/openroad.py
```diff
@@ -68,7 +70,14 @@ def setup(chip, mode='batch'):
     chip.set('eda', tool, 'option',  step, index, option, clobber=clobber)
     chip.set('eda', tool, 'refdir',  step, index, refdir, clobber=clobber)
     chip.set('eda', tool, 'script',  step, index, refdir + script, clobber=clobber)
-    chip.set('eda', tool, 'threads', step, index, os.cpu_count(), clobber=clobber)
+
+    # normalizing thread count based on parallelism and local
+    threads = os.cpu_count()
+    if not chip.get('remote') and step in chip.getkeys('flowgraph', flow):
```
nmoroze, line 76: Not sure if the `chip.get('remote')` will have the desired effect here, since I think that will be unset on the machine where this ultimately runs.

Might need to rely on something in `chip.status` to draw a distinction here, although using that seems a bit funky to use in a tool driver.

WRansohoff, line 76:
```suggestion
    if (not chip.get('remote')) and (step in chip.getkeys('flowgraph', flow)):
```

Probably not necessary, but C has left me with a habit of using parentheses to over-define the order of operations.

examples/benchmark/benchmark.py
```diff
@@ -25,13 +23,13 @@ def main():
...
         results[design] = {}
         for n in ['1','2','4','8','16']:
-        for n in ['4']
             wall_start = time.time()
             chip = siliconcompiler.Chip(design)
             chip.set('jobname', f"job{n}")
             chip.load_target('skywater130_demo')
             chip.set('relax', True)
             chip.set('quiet', True)
+            chip.set('remote',False)
```
WRansohoff, line 32:
```suggestion
            chip.set('remote', False)
```

nmoroze (commented):
Generally looks good, left a couple comments about things that seemed off. Main thing is I'm pretty sure we're not pulling in everything that could change in the replaced `read_manifest()` calls. Also it looks like we repeat that logic a couple times (and they don't currently match), so factoring it out into a separate function could be a good idea.

siliconcompiler/core.py
```diff
@@ -3931,6 +3960,7 @@ def run(self):
             raise SiliconCompilerError(f'Run() failed on step {failed_step}! '
                 f'See logs in {stepdir} for error details.')
 
+
```
WRansohoff, line 3963:
```suggestion
```

WRansohoff (approved):
LGTM - a few small comments, but none of them are blocking.

siliconcompiler/core.py
aolofsson: We should move the report file pointers back to the setup(), this removes the need to copy over the eda, report dictionary. Seems cleaner than setting it up in the post_process(). The reason I put it there was to keep all the reporting logic together in one place. From an setup/execution point, it's better to have it done once in the setup() call.

siliconcompiler/core.py
aolofsson: Agree on the 'record', that is definitely missing. Let's fix the flowstatus when you update the run schedulibg and modular flowgraphs. They all go together.

siliconcompiler/tools/openroad/openroad.py
aolofsson: Good point! We need to not overload 'remote' and find a different solution to changing the operating mode.

siliconcompiler/core.py
```diff
@@ -1885,6 +1885,7 @@ def write_manifest(self, filename, prune=True, abspath=False, job=None):
             else:
                 self.logger.error('File format not recognized %s', filepath)
                 self.error = 1
+            #flush writes
```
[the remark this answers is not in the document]
aolofsson: Left over

PR #957, opened 2022-03-21
Heartbeat padring example

WRansohoff:
This change adds a minimal example demonstrating how to use `asictopflow` to stitch macros together. There are two build scripts: one generates floorplans using the Python API, and one uses pre-built floorplan files which are committed in the example directory.

Leaving this as a draft so that we can address a couple of points:

* Should we get the `sky130` I/O and `oh` padring files from directories under `third_party/`, instead of placing them in the example directory? I couldn't find all of the required sky130 sources in our current submodules, and our `oh` submodule seems to be incompatible with the version used in [the `zerosoc` repository](https://github.com/siliconcompiler/zerosoc).
* I'd like to add a CI test for `asictopflow` using this example, but we'll need to merge #951 before that will pass.

[Commit 0010262](https://github.com/siliconcompiler/siliconcompiler/commit/001026230155a29d3190fed0ad920bc3096c2b68) contains only the new example files, if the padring/io commits are distracting.


aolofsson:
Very cool.  We should defininitely use the original versions and not copy in versions. If there are problems in oh/skywater let's fix them.

WRansohoff:
Like we discussed, I moved the I/O files out of the example directory by updating the `oh` submodule and creating a new `sky130io` directory under `third_party/pdks/`. For now, I followed [the zerosoc directory structure](https://github.com/siliconcompiler/zerosoc/tree/main/asic/sky130/io) and put the wrapper files under the same `sky130io/io` directory as the LEF/GDS/etc files. But I would be open to moving some of the Verilog sources if that would make more sense.

I also added daily CI tests for the `build.py` and `floorplan_build.py` scripts, so I think this example is ready for review now.

WRansohoff:
Like we discussed today, I dropped the `prim_sky130_clock_gating.v` file, and the build scripts still work. I think this is ready to merge now.

aolofsson:
This looks great! Thanks for going through the refactoring rounds to save duplication of files.

aolofsson (approved):

PR #958, opened 2022-03-23
Raise exceptions instead of calling 'sys.exit', to facilitate using siliconcompiler as a Python library

WRansohoff:
This change replaces most of our `sys.exit` calls with Python exceptions, because we want to make it easy to catch build errors within custom Python flows.

I left one `sys.exit` call in the `_haltstep` method, because the `_runtask` calls which it interrupts are typically run within a `multiprocessing.Process`. Raising exceptions in `_haltstep` would also make the command-line errors more verbose and less helpful, by printing an identical stack trace for every failed task in the flow.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -508,7 +508,7 @@ class Chip:
             func(self)
         else:
             self.logger.error(f'Module {name} not found.')
-            raise RuntimeError(f'Module {name} not found.')
+            raise SiliconCompilerError(f'Module {name} not found.')
 
     ##########################################################################
     def load_pdk(self, name):
@@ -533,7 +533,7 @@ class Chip:
             func(self)
         else:
             self.logger.error(f'Module {name} not found.')
-            raise RuntimeError(f'Module {name} not found.')
+            raise SiliconCompilerError(f'Module {name} not found.')
 
     ##########################################################################
     def load_flow(self, name):
@@ -558,7 +558,7 @@ class Chip:
             func(self)
         else:
             self.logger.error(f'Module {name} not found.')
-            raise RuntimeError(f'Module {name} not found.')
+            raise SiliconCompilerError(f'Module {name} not found.')
 
     ##########################################################################
     def load_lib(self, name):
@@ -583,7 +583,7 @@ class Chip:
             func(self)
         else:
             self.logger.error(f'Module {name} not found.')
-            raise RuntimeError(f'Module {name} not found.')
+            raise SiliconCompilerError(f'Module {name} not found.')
 
 
     ###########################################################################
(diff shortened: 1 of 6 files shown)


siliconcompiler/core.py
```diff
@@ -508,7 +508,7 @@ def load_target(self, name):
             func(self)
         else:
             self.logger.error(f'Module {name} not found.')
-            sys.exit(1)
+            raise SiliconCompilerError(f'Module {name} not found.')
```
arlpetergadfort, line 511: Shouldn't this be `Target {name} not found.`

siliconcompiler/core.py
```diff
@@ -533,7 +533,7 @@ def load_pdk(self, name):
...
```
arlpetergadfort, line 536: Shouldn't this be `PDK {name} not found.`

siliconcompiler/core.py
```diff
@@ -558,7 +558,7 @@ def load_flow(self, name):
...
```
arlpetergadfort, line 561: Shouldn't this be `Flow {name} not found.`

siliconcompiler/core.py
```diff
@@ -583,7 +583,7 @@ def load_lib(self, name):
...
```
arlpetergadfort, line 586: Shouldn't this be `Library {name} not found.`

siliconcompiler/core.py
```diff
@@ -334,10 +334,10 @@ def create_cmdline(self, progname, description=None, switchlist=[]):
                 self.set('mode', cmdargs['mode'], clobber=True)
             if 'techarg' in cmdargs.keys():
                 print("NOT IMPLEMENTED")
-                sys.exit()
+                raise NotImplementedError('NOT IMPLEMENTED')
```
arlpetergadfort, line 337: Shouldn't this be `techarg not implemented`

siliconcompiler/core.py
```diff
@@ -334,10 +334,10 @@ def create_cmdline(self, progname, description=None, switchlist=[]):
...
             if 'flowarg' in cmdargs.keys():
                 print("NOT IMPLEMENTED")
-                sys.exit()
+                raise NotImplementedError('NOT IMPLEMENTED')
```
arlpetergadfort, line 340: Shouldn't this be `flowarg not implemented`

siliconcompiler/core.py
WRansohoff: I believe that 'Module' is being used in the sense of a [Python module](https://docs.python.org/3/tutorial/modules.html) in these messages.

These methods dynamically search for a function called `setup` in a module with the given name. If we call `load_target(freepdk45_demo)`, the method will look for the setup function in a module called `freepdk45_demo[.py]`. The `load_target`, `load_lib`, etc methods look in different directories (`targets/`, `libs/`, etc), to allow duplicated target/lib/etc names in the future.

But you make a good point that we should make these messages more precisely-worded, thank you for the feedback!

siliconcompiler/core.py
WRansohoff: Changed these to, "NOT IMPLEMENTED: '[param]' parameter"

aolofsson:
Looks good! Great to get this cleaned up.

aolofsson (approved):

PR #964, opened 2022-03-28
Record each task's maximum memory usage in metrics

WRansohoff:
We want to record peak memory usage as part of the metrics collected for each task.

The built-in [resource](https://docs.python.org/3/library/resource.html) library looked like a promising solution, but there were a couple of issues:

* It is not included in Windows installations of Python.

* It looks like the most detailed statistic that it  can provide is the maximum "Resident Set Size". That includes shared memory such as libraries which may be used by several processes without actually "belonging" to the process whose memory usage we are interested in.

The [`psutil`](https://pypi.org/project/psutil/) module can provide more comprehensive information about processes, and it appears to be fairly widely-used. It can report the "Unique Set Size" for a process and its children, which only includes memory dedicated to that process.

This approach should work on Windows/OSX/Linux/etc, but it appears to return values of zero bytes when run within a Windows/Linux subsystem (v1). I wasn't able to find a simpler, more cross-platform approach within a small timebox, but I would be open to suggestions if there are better built-in options.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,6 +17,7 @@ cryptography >= 3.4.7
 graphviz >= 0.18.1
 netifaces >= 0.11.0
 distro >= 1.6.0
+psutil >= 5.8.0
 
 # Build dependencies
 #:build
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3555,11 +3555,7 @@ class Chip:
                         # Gather subprocess memory usage.
                         try:
                             pproc = psutil.Process(proc.pid)
-                            pchildren = list(pproc.children(recursive=True))
-                            mem_uss = pproc.memory_full_info().uss
-                            for child in pchildren:
-                                mem_uss += child.memory_full_info().uss
-                            max_mem_bytes = max(max_mem_bytes, mem_uss)
+                            max_mem_bytes = max(max_mem_bytes, pproc.memory_full_info().uss)
                         except psutil.Error:
                             # Process may have already terminated or been killed.
                             # Retain existing memory usage statistics in this case.
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3548,7 +3548,6 @@ class Chip:
                     # live output in non-blocking way, so we can monitor the
                     # timeout. Based on https://stackoverflow.com/a/18422264.
                     cmd_start_time = time.time()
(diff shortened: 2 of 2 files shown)


aolofsson:
Does this correlate with the values reported by openroad, yosys, surelog?

WRansohoff:
It's not an exact match of what the tools report, because this change polls each running process at an interval. But it seems to be very close (on the order of ~1-10MiB difference) for non-trivial designs.

On very small designs, such as the GCD or heartbeat example, each step usually takes less than a second to run. We poll every 0.1+ seconds, so we don't always catch the transient peak. We still see the correct order of magnitude though, e.g. ~35MiB reported vs. ~50MiB actual in yosys.

On longer-running designs, such as ZeroSoc, the results are much better. Taking yosys as an example, we reported ~1.3x memory usage when child processes were included in the `psutil` accounting. When we only check the main PID, however, we calculate 298.48MiB compared to the 299.88MiB reported in yosys.log. OpenROAD is similar; the main PID's post-routing memory usage is reported as 2162.74MiB, compared to the 2,163.06MiB post-routing peak recorded in route.log.

Considering those results, I've removed child process memory usage from our final tallies. If we can find a way to catch process termination with callbacks rather than polling, we can probably improve the accuracy of results for small designs.

WRansohoff:
Here are the results from a few trial runs of ZeroSoc, in megabytes. I would include more tasks, but most of them don't currently print utilization statistics to log files:

| Step | Memory (reported) | Memory (actual) | Relative error |
|------|----------------------|--------------------|----------------|
|syn0|296.29|299.16|-0.96%|
|route0|2140.61|2144.25|-0.17%|
|syn1|296.5|299.16|-0.89%|
|route1|2171.12|2171.43|-0.01%|
|syn2|296.37|299.15|-0.93%|
|route2|2201.47|2201.79|-0.02%|

It looks like this method tends to lowball the statistics by a few megabytes, but the error margin is pretty small, so it should be close enough to gauge how much memory a design requires to build. If we need more precise accuracy, maybe we could investigate wrapping the subprocess commands with a memory profiler such as `time` or `valgrind` (may be too slow).

aolofsson:
Thanks for going through the exercise of correlating with the tool reporting! This looks fantastic. Just resolve the conflicts and we are good to go.

aolofsson (approved):

PR #965, opened 2022-03-28
Proposal to refactor task management

nmoroze:
This PR implements an idea I had to refactor our parallelism/task management code. It is meant to take care of a couple things on our to-do list in one sweep, and I think it simplifies our code overall. The main thing it addresses is to make the scheduler graph-based rather than flooding the system with nodes that block until ready.

The proposed algorithm for launching tasks in `run()` is as follows:
- Create a dict `tasks_to_run` that maps each task to a list of its task dependencies
- Create an empty list `tasks_running` that lists each task that is currently active
- While either one of these lists is not empty:
  - Iterate through `tasks_to_run`
    - If any of them have any empty dependency list, launch the task, remove it from `tasks_to_run` and add it to `tasks_running`.
  - Iterate through all tasks in `tasks_running`
    - If any of them have finished, check the exit code to update their status, remove them from `tasks_running`, and remove them from dependency lists in `tasks_to_run`.

I also have an idea on how we can implement breakpoints in a parallel-safe way, and I think this refactor should help fix our latent `'flowstatus', ... 'error'` issues that have been around for a while.

Here are the code simplifications that this refactor enables:
- The main idea is to switch from shared `active`/`error` bit dictionaries to a single enumerated type for tracking task status with 3 states: `STATUS_PENDING`, `STATUS_COMPLETE`, and `STATUS_ERROR`. In addition, we can switch from storing just an error bit in the schema to storing this value under a single `'flowstatus', step, index, 'status'` parameter. This buys us a couple things:
  - Fewer variables to have to pass around
  - We now have the ability to quickly determine via the schema when a step as already been run, which I think will be helpful for future tasks (such as implementing 'resume'-type functionality)
  - Eliminates deadlock potential due to `error`/`active` being changed in the wrong order
- In addition, since we now rely on `run()` to launch each node, we no longer need to use shared dictionaries!
  - Once called, each `_runtask()` can assume it is allowed to go 
  - Each task needs to know the status of all *proceeding* tasks, but we can do that with a "frozen" copy of `status` since those statuses will all be populated before that call.
  - We can determine the outcome of the task by looking at the process exit code. This works with minimal changes since we already use `sys.exit(1)` on error in `_runtask()`
- Getting rid of the synchronization has a few benefits: 
  - Should have positive impacts on performance, since as far as I understand using a multiprocessing `Manager` launches an extra thread behind-the-scenes
  - Makes the code easier to reason about and safer from devious bugs
  - Removes need for things like `_runtask_safe()` wrapper, since processes no longer have a risk of hanging due to unhandled exceptions.

### DIFF
--- a/examples/benchmark/scalability.py
+++ b/examples/benchmark/scalability.py
@@ -1,0 +1,74 @@
+'''Benchmark to measure time taken per simple task with large flowgraphs.
+
+Running this file determines the average time taken per task on long serial
+flowgraphs of 10, 100, 250, and 500 simple "echo" tasks.
+
+Since it would take too long to measure the entire run with a flowgraph of such
+a length, we perform a weird trick to measure just the first "STEPS_TO_RUN"
+tasks. Instead of calling "run_long_serial()" directly, this file calls back into
+itself using "subprocess", using a CLI argument to change the codepath. That way,
+it can monitor stdout to start timing when it sees the initial "import" echo (so
+that it doesn't measure start-up time), and finish timing and end the process
+when it sees the "done" echo.
+
+While it would be more straightforward to do this with a steplist, using a
+steplist removes some of the overhead we're trying to measure (e.g. the flooding
+scheduler only launches processes based on what's in the steplist).
+'''
+import siliconcompiler
+
+import os
+import signal
+import subprocess
+import sys
+import time
+
+STEPS_TO_RUN = 10
+
+def run_long_serial(N):
+    chip = siliconcompiler.Chip()
+    flow = 'test_long_serial'
+    pipe = [{'import': 'echo'}]
+    pipe += [{f'measured{i}': 'echo'} for i in range(STEPS_TO_RUN - 1)]
+    pipe += [{'done': 'echo'}]
+    pipe += [{f'extra{i}': 'echo'} for i in range(N - (STEPS_TO_RUN + 1))]
+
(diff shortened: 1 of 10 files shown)


nmoroze:
Alright, I was able to collect some data showing that this does in fact give a performance benefit. However, it turns out that `check_manifest()` in `_runtask()` is currently the primary scaling bottleneck (by a lot), and so the benefit is difficult to see unless checking is disabled. I think the `check_manifest()` bottleneck could be easily removed by only running dynamic checks in `_runtask()` -- I think any checks that scale with the flowgraph only need to be done statically at the start of the run. I can explore this in a future PR.

In the meantime, I wrote up a script that can be used to repeat my scalability experiments. Here's a graph of the data I collected:

Time_task (before) and Time_task (after)

y-axis is time per task in seconds, x-axis is how many nodes are in the (serial) flowgraph. As you can see, the time still creeps up a bit even with this PR (to be expected, since the scheduler loop in run() gets longer with more nodes), but the slope is much lower now.

With this, I think I'm ready to merge this PR! I will address some of the things that came up that we've discussed offline in future PRs.

aolofsson:
This is a great improvement! Much more scalable, well done! Now that we don't flood the system with processes, can the last sys.exit in the _haltstep be changed to a raise exception?

aolofsson (approved):

PR #967, opened 2022-03-29
Fix recent daily test failures

WRansohoff:
This change should fix the recent daily test failures.

Most of them were caused by my recent change to move from `sys.exit` to Python exceptions - sorry that I only checked the quick tests for that PR.

I believe that the ASAP7 GCD test was failing because of a missing 'wafersize' parameter, which is newly required for the 'asic' mode.

[I started an extra test run on this branch](https://github.com/siliconcompiler/siliconcompiler/actions/runs/2060655730) to verify that this will fix the daily test suite.


aolofsson:
No need to add the wafersize to the test, it should go in the PDK. I added it there.

WRansohoff:
Okay, I dropped the asap7 fix.

aolofsson (approved):
Looks good!

PR #971, opened 2022-04-01
Replace direct cfg merge w/ partial read_manifest

nmoroze:
This PR tweaks the `read_manifest()` replacement from PR #954 by moving back to `read_manifest()`, with a 'partial' option that causes it to only merge the keys we care about.

It turns out the previous approach has a few functionality issues:
- It doesn't read data from steps/indices prior to a tasks immediate inputs, so the manifests don't store metrics cumulatively. 
- Since the json manifest has "default" keypaths pruned, doing a full copy in of an entry from this dictionary breaks the ability to go back and create more groups underneath dynamic keys. 

This new approach of doing a "partial" read_manifest() gets most of the performance benefit without the correctness pitfalls. The following data is collected by running `python examples/benchmarks/scalability.py parallel`. It constructs a wide flowgraph with 10, 100, and 250 parallel nodes, and measures the runtime from import completion (so setup not counted), to final merge (but not run() cleanup), and divides this by the number of parallel tasks.

**Time per task**

|# nodes| read_manifest() | optimize_runtask | read_manifest(partial=True) |
|-----------|----|----|---|
| 10  | 0.14s  | 0.07s | 0.07s|
| 100 | 7.05s | 0.66s | 0.85s|
| 250 | 32.32s | 4.56s | 5.71s|

### DIFF
--- a/examples/benchmark/scalability.py
+++ b/examples/benchmark/scalability.py
@@ -1,7 +1,7 @@
 '''Benchmark to measure time taken per simple task with large flowgraphs.
 
-Running this file determines the average time taken per task on long serial
-flowgraphs of 10, 100, 250, and 500 simple "echo" tasks.
+Running this file determines the average time taken per task on long serial or
+parallel flowgraphs of 10, 100, 250, and 500 simple "echo" tasks.
 
 Since it would take too long to measure the entire run with a flowgraph of such
 a length, we perform a weird trick to measure just the first "STEPS_TO_RUN"
@@ -17,7 +17,6 @@ scheduler only launches processes based on what's in the steplist).
 '''
 import siliconcompiler
 
-import os
 import signal
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


aolofsson:
A couple of things to discuss before merging this.

>It doesn't read data from steps/indices prior to a tasks immediate inputs, so the manifests don't store metrics cumulatively.

How does moving to read_manifest solve this. The read_manifest() reads the same json file like the current code?

>Since the json manifest has "default" keypaths pruned, doing a full copy in of an entry from this dictionary breaks the ability to go back and create more groups underneath dynamic keys.

For record/manifeset/flowstatus what is the situation inside runtask where we would want to add additional entries?

This addes another arg to the public api which is a small negative (documentation, support etc). Also, note that it's slower, which may be good enoug...

There may be casees where we have many more metrics and records, so also think about where we will end up with the benchmark. I think this is one of those times, where we do have to optimize just a bit.

nmoroze:
Per offline discussion, tweaked the PR to make the "partial" arg not part of the public interface.

aolofsson (approved):
Ok, I think that's a fair compromise. If we ever do run into performance issues, we can revisit, but for now this looks like the right solution.

PR #972, opened 2022-04-05
Make '==' default version spec operation

nmoroze:
For backwards compatibility, this PR makes it so any version specifier that's missing an operator defaults to a '==' specifier.


aolofsson (approved):

aolofsson:
Thanks!

PR #973, opened 2022-04-06
Fix flowstatus bug

nmoroze:
This PR fixes issue #651, the flowstatus issue that has been around for a while.

This PR makes two changes:
- Doesn't store flowstatus "pending" values in the manifest. Storing an explicit value for "pending" meant that our `_read_manifest()` calls wouldn't overwrite these values since they are called with clobber=False.
- Added a code block to write out the `status` dict into the schema on success. This takes care of a few edge cases where we have steps that don't merge into one.

I also added a test based on my reproduction code from the original issue. It runs a handful of graphs with 2 parallel place steps, where one fails and one succeeds. In each test case, it ensures that the flowstatuses for these place steps are recorded correctly.

It tests this graph with 3 steplists:
- `['import', 'place']`
- `['import', 'place', 'placemin']`
- `['import', 'place', 'placemin', 'cts']`

import-place-placemin-cts

It tests this other graph:
longbranch

And finally it tests the regular asicflow with 2x placement running "remotely" on the dev server.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3918,6 +3918,16 @@ class Chip:
                 if not index_succeeded:
                     raise SiliconCompilerError('Run() failed, see previous errors.')
 
+            # On success, write out status dict to 'flowstatus'. We do this
+            # since certain scenarios won't be caught by reading in manifests (a
+            # failing step doesn't dump a manifest). For example, if the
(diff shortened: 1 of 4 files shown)


aolofsson (approved):
Nice fix to a long standing issue!

PR #974, opened 2022-04-07
Refactor logic at end of run()

nmoroze:
This PR is a refactor I thought was needed for #973, but turned out to be unnecessary. I think it's a nice tweak though, so I'm splitting it out so we can review separately.

At a high level, our old run logic looked something like:

```
if remote:
  launch remote run
else:
  launch local run

merge manifests from indices of last step:
  loop over all, load first w/clobber=True (but preserving local dir), partial load the rest

if no valid manifests, read directory to find last successful step 
(for remote only, local should have finished successfully by this point)

write manifest into root
```
The new logic looks like:
```
if remote:
  launch remote run
  if root manifest returned from server:
    load full root manifest, preserving local dir
  else:
     perform hack to determine failed step
else:
  launch local run
  merge manifests from indices of last step (all partial)

write manifest
```

While it actually results in less common code, I like that this splits up the code paths logically according to the distinct needs of local vs remote runs. Now that we have a single final root manifest, we don't need to repeat the merging logic on the client after a remote run, and we can lift the code block that detects the falling step up into the remote code path, since it's only meaningful there.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3789,6 +3789,34 @@ class Chip:
 
             # Fetch results (and delete the job's data from the server).
             fetch_results(self)
+
+            # Read back configuration from final manifest.
+            cfg = os.path.join(self._getworkdir(),f"{self.get('design')}.pkg.json")
+            if os.path.isfile(cfg):
+                local_dir = self.get('dir')
+                self.read_manifest(cfg, clobber=True, clear=True)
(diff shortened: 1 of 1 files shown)


aolofsson (approved):
Looks good. I think there are some opportunities to merge the remote and local even more than we have now. While not too bad currently, the less branching we can have between remote and local int he main thread the better.

PR #975, opened 2022-04-07
Prevent defvalue lists from getting overwritten

nmoroze:
This PR fixes a little bug I noticed where defvalues for list items are updated as the value is, since the value is pointing to the same object.


aolofsson (approved):
Nice catch!!

PR #977, opened 2022-04-07
'vercheck' -> 'novercheck'

nmoroze:
Otherwise we can't disable it via CLI.


aolofsson (approved):
Thanks! Always important to have a bypass path and it's nice to have all the booleans be False by default and not have any outliers.

nmoroze:
Whoops, I think my previous flowstatus change non-deterministically broke the CI tests (if it tries to run two server tests using the same port in parallel, one of them hangs. This should be fixed up now!

Issue #932, opened 2022-03-11
Failure in .summary with failed step


arlpetergadfort:
Something fails with the "DELAY3" synthesis option
`
| ERROR   | job0    | syn          | 3   | Halting step 'syn' index '3' due to errors.
`

The flow continues, but then errors out like this when printing summary/writing the HTML report
`    chip.summary()                                # print results summary`
`  File "0.7.0/siliconcompiler/core.py", line 2760, in summary`
`    results_fn = results_gds`
`  File "jinja2/environment.py", line 1304, in render`
`    self.environment.handle_exception()`
`  File "jinja2/environment.py", line 925, in handle_exception`
`    raise rewrite_traceback_stack(source=source)`
`  File "0.7.0/siliconcompiler/templates/report/sc_report.j2", line 56, in top-level template code`
`    {% if step in manifest['eda'][manifest['flowgraph'][manifest['flow']['value']][step][index]['tool']['value']]['report'] and mk in manifest['eda'][manifest['flowgraph'][manifest['flow']['value']][step][index]['tool']['value']]['report'][step][index] %}`
`jinja2.exceptions.UndefinedError: 'dict object' has no attribute '3'`

aolofsson:
Apologies for the slow response on this one.  I think we can resolve this as part of the run scheduler changes  being implemented by @nmoroze

PR #978, opened 2022-04-07
Fix summary/report generation issues

nmoroze:
This PR fixes:
- "ERR" when running summary() on a chip with steplist. We changed this in #777 but I can't remember why, and I think we should probably go back.
- Missing report key (issue #932). This extra check might not be necessary once we swap report pointers to always be done in setup(), but it won't hurt for now.
- Display "(failed)" under metrics for tasks that error out, since they won't have metric values filled in (was causing KeyError).

I inserted some summary() calls into tests with weird steplists to try and catch these issues in CI.


aolofsson (approved):
Nice fix!

PR #980, opened 2022-04-07
Move EDA report setup to tool setup() functions

nmoroze:
RE: https://github.com/siliconcompiler/siliconcompiler/pull/954#discussion_r837860771, this PR moves all the EDA report setup from post_process() to each tool driver's setup() function. This is more consistent with the input/output schema, and it is necessary now that we don't persist any mid-run changes to schema other than metrics/record/flowstatus.


aolofsson (approved):
Makes sense.

PR #982, opened 2022-04-08
Fix failing daily tests

nmoroze:
This PR fixes up the daily tests that have been failing.

It was 2 things:
- I think the method of adding the examples directory to the Python path wasn't working for the heartbeat padring test. I'm not 100% sure why not, but I figured it might be easiest to reuse the examples testing infrastructure (the setup_example_test fixture) since that handles the Python path configuration.
- We need to clone OH for the test_oh_experiments test (of course!) 

I'll start a run of the daily tests CI on this branch a bit later tonight to make sure it doesn't hog the test runner while others are working, and then undo draft status once it's verified green.


aolofsson:
Do you want to merge this?

WRansohoff:
I think Noah might be trying to get the daily tests to pass on this branch first.

The changes to the heartbeat_padring tests look fine to me; it seems a bit odd that `chip.scroot` wouldn't work on the test runner, but thanks for catching that.

nmoroze:
Yup, should be almost there!

> The changes to the heartbeat_padring tests look fine to me; it seems a bit odd that chip.scroot wouldn't work on the test runner, but thanks for catching that.
I actually figured out the root cause of this: the daily test runner installs siliconcompiler without using the `-e` flag, so `chip.scroot` gives a path somewhere in the Python `site-packages/` directory rather than the actual place where the directory was cloned. Just using a relative path works okay since we know we're running this from `examples/`, which means the `third_party/` should be present (granted, I realize that might make this a bit less robust as a general example).

nmoroze:
This should be good now, verified green: https://github.com/siliconcompiler/siliconcompiler/actions/runs/2152650119! 

The final bug was a little tricky, even though the example tests modify `sys.path` only within the scope of each individual test, Python still maintains an import cache that maps certain module names to files. Therefore importing "build" in both `test_heartbeat_padring.py` and `test_microwatt.py` causes a problem in whichever one comes second. Put in a "good enough" fix to this, will think about if there's an easy way to make this more robust going forward.

nmoroze:
Good question - after resetting the metrics to `None` at the start of run(), `check_manifest()` began failing due to missing metric requirements. There were two things going on:

- `check_manifest()` only checks requirements for existing keypaths, which should probably get changed. If a dynamic dictionary is required, I imagine `check_manifest()` should make sure that something actually got filled in (added this to our todo list)
- Metrics shouldn't be marked as required, since they're not filled in till after a run (but requirements are checked before a run)

aolofsson (approved):

PR #983, opened 2022-04-08
Fix Surelog install script

WRansohoff:
It looks like GitHub recently changed its security policy to block some methods of cloning git:// addresses. One of Surelog's dependencies includes a 'git clone' call that points to such a repository, so our surelog install script has started to fail.

It looks like the solution is to force Git to use https:// instead of git://. It's a little tricky to force that setting without modifying the system's Git configuration, but I think the easiest approach is to override the default `~/.gitconfig` search path in `install-surelog.sh`.


WRansohoff:
It looks like [Surelog recently added a fix for this issue](https://github.com/chipsalliance/Surelog/commit/c1c8149eba4f2cd187ff6706baf1e4be3d354ecb), so we should be able to remove this workaround when we update to a new version of the tool.

aolofsson (approved):

PR #988, opened 2022-04-11
Insert valid steplist check

nmoroze:
This PR implements a `check_manifest()` check for ensuring that a steplist is valid. This will print a clean error early on if you try to run a flowgraph with a steplist that doesn't work, e.g.

```python
...
chip.set('steplist', ['syn'])
chip.run()
```

Gives you:
```
| INFO    | job0  | --         | -  | Checking manifest before running.
| ERROR   | job0  | --         | -  | syn0 relies on import0, but this task has not been run and is not in the current steplist.
| ERROR   | job0  | --         | -  | Check failed. See previous errors.
```

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1376,7 +1376,6 @@ class Chip:
         if os.path.isfile(filename):
             return filename
         else:
-            self.error = 1
             return None
 
     ###########################################################################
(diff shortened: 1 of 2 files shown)


aolofsson (approved):
Nice user-friendly change!

PR #990, opened 2022-04-13
Remove requirement for totaltime metric

nmoroze:
Looks like tests broke after the fix-daily-tests got merged, since there was a new metric 'totaltime' added, and we need to get rid of `require` for metrics now. This commit should fix things up!


aolofsson:
Thanks, sorry about that!

aolofsson (approved):

PR #991, opened 2022-04-13
Implement checklists

nmoroze:
This PR implements checklist functionality. Main changes:

- Add `load_checklist()` function to dynamically load checklist modules
- Re-implement `check_checklist()` based on the algorithm we discussed offline
   - Items that can be automatically checked must meet criteria for all associated tasks (unless waivers exist), and EDA reports must exist for all criteria
   - Items that are not automatically checked must have at least one report (manually added)
   - If `check_ok` is True, items must also have 'ok' parameter checked
- Re-architect the 'checklist' schema slightly to enable automatic checks of multiple tasks
  - I also got rid of the 'metric' subkey under 'report' to simplify things a bit, since we seemed okay with it being a catch-all bucket. However, I can add that back if you don't think that's a good change.
- Add example to `gcd_skywater.py` with simple subset of the OH tapeout checklist

Since the `check_checklist()` function has to go back into the schema history to handle multiple jobs, this change also entailed adding an optional `job` kwarg to some of our core functions, following the pattern of `set`/`get`/`add`.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -10,4 +10,5 @@ include CMakeLists.txt
 graft siliconcompiler/leflib
 graft siliconcompiler/templates
 graft siliconcompiler/tools
+graft siliconcompiler/checklists
 graft third_party/tools/openroad/tools/OpenROAD/src/odb/src/lef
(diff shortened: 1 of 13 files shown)


aolofsson (approved):
This is awesome!

PR #992, opened 2022-04-14
Update history in read_manifest()

nmoroze:
This PR fixes `read_manifest()` so that it will update the history of the current chip object if the manifest contains history information.

Previously, calling `read_manifest()` on a manifest with something in the history would return a whole bunch of errors:
```
| ERROR   | Get keypath [history,job0,steplist] does not exist.
| ERROR   | Get keypath [history,job0,steplist] does not exist.
| ERROR   | Get keypath [history,job0,steplist] does not exist.
| ERROR   | Get keypath [history,job0,steplist] does not exist.
| ERROR   | Get keypath [history,job0,target] does not exist.
| ERROR   | Get keypath [history,job0,target] does not exist.
| ERROR   | Get keypath [history,job0,target] does not exist.
| ERROR   | Get keypath [history,job0,target] does not exist.
```

and would not populate the history of the chip object.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1481,13 +1481,15 @@ class Chip:
         """
         if job is not None:
             # fill ith default schema before populating
-            self.cfghistory[job] = schema_cfg()
-            dst = self.cfghistory[job]
+            self.cfg['history'][job] = schema_cfg()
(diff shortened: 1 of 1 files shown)


aolofsson (approved):
Nice find, and nice fix!

PR #994, opened 2022-04-19
Update openroad version

WRansohoff:
It looks like we can update our version of OpenROAD to include some recent power grid generation improvements without changing our tcl scripts.

I tested the newer version on Ubuntu 20.04 and RHEL7.9 with our heartbeat and ZeroSoC examples, and everything completed successfully.

I've also installed the newer version on our CI runner so that the tests pass, but we can revert that if problems crop up.


aolofsson (commented):
Looks great!

aolofsson (approved):

PR #999, opened 2022-04-26
Decode non-ASCII stdout output in AppGen

nmoroze:
This change should fix the failing docs build in `pacman` branch. The AppGen extension expected app help output to be all ASCII, but 'sup' uses unicode characters for representing the directory structure.


aolofsson:
Great! I was wondering what was going on there.

PR #997, opened 2022-04-22
Adding support for package management

aolofsson:
```rst 

This is a proposal for a package management system for SC inspired by python-pip and rust-cargo. 

Publishing a known good chip object to a registry::

  from siliconcompiler.package import Sup
  p = sc.package.Sup()
  p.publish("daughter.pkg.json", registry)

Using packages from the repository::

  chip = sc.Chip(design='mother')
  chip.set('package', 'dependency', 'daughter', '1.2.3')
  chip.set('autoinstall', True)
  chip.update()
  chip.run()

Package management system features:

* SC package files (.sup) are direct products of the SiliconCompiler schema
* Registries can be on disk folders or IP addresses comptaible.(needs server infrastructure)
* Packages are named <design>.<semver>.sup.gz
* Dependencies are included in a project through the 'package, dependency' schema parameter
* Dependencies resolved within a design through the update() core API method
* Packages are installed in the local ~/.sc/registry cache by default
* Packages are organised according to the following directory structure

 .. code-block:: text

  <design>
         └── <version>
             ├── <jobname>
             │   ├── import (sources)
             │   └── export (results)
             ├── <jobname>
             │   ├── import
             │   └── export
             ├── <design>-<version>.html
             └── <design>-<version>.sup

A command line app 'sup' supports direct package management (like pip). Commands include

* check     : Check that manifest before publishing.
* publish   : Publish chip manifest to registry.
* install   : Install registry package in local cache.
* uninstall : Remove package from local cache.
* search    : Search for a package in registry.
* info      : Display package information.

New schema parameters included for package management:

* [depgraph]: Resolved package dependency list (automatically updated by SC)
* [registry]: List of registry filepaths and IP address.
* [autoinstall]: Enables autoinstall by the core update() method
* [package, dependency]: list of package dependencies for the current design

```

PR #1000, opened 2022-04-26
Add a sup error for circular imports

WRansohoff:
This change adds an `upstream` parameter to track prior dependencies in sup's recursive `_find_deps` function. It should catch circular imports by throwing an error if the system encounters a dependency which is currently being processed.

There's also a unit test to verify that it works in a simple "A depends on B, B depends on A" situation.

This should be applied to #997 if it looks good; I think merging it into the `pacman` branch should accomplish that.


aolofsson:
Thanks! Nice catch and fix.!

aolofsson (approved):

nmoroze:
Wow, PR #1000!

PR #1001, opened 2022-04-27
Remove single-step flow logic from run()

nmoroze:
This PR removes the single-step flow logic from the top of `run()`, one of our cleanup backlog items. I replaced it with a helper function to take care of some of the boilerplate required when creating a single-step graph (import no-op, default metric weights).

A consequence of this change, however, is that we will no longer be able to concisely run a single tool step through the CLI. Before, we supported:

`sc foo.vhdl -flow icarus -arg_step compile`

Now, we'd have to do something very verbose like:

```bash
sc foo.v \
  -flow icarus \
  -flowgraph_tool "icarus compile 0 icarus" \ 
  -flowgraph_tool "icarus import 0 nop"  \
  -flowgraph_input "icarus compile 0 (import, 0)" \
  -flowgraph_weight "icarus compile 0 errors 0 " \
 etc...
```

I wanted to get thoughts on this before finalizing this change. I could see it as a dealbreaker, since I think quickly running a tool directly is a good use of the SC CLI.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3267,6 +3267,35 @@ class Chip:
                 self.edge(flow, prevstep, step)
             prevstep = step
 
+    ###########################################################################
+    def single_step(self, step, tool, flow=None):
+        '''
+        Sets up a flow that runs a single tool step.
+
+        This function creates a flowgraph that runs a single tool step. If the
+        specified step name is not "import", this function adds a no-op "import"
+        step to satisfy the import requirement. This function also sets weights
+        for a few basic metrics to ensure that the summary() function displays
+        errors, warnings, and runtime.
+
+        Args:
+            step (str): Name of step
+            tool (str): Name of tool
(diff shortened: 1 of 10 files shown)


nmoroze:
Note to self: weights are actually set up by node(), so can remove that boilerplate from the single_step() function.

aolofsson:
I am not sure about this. Everything else is built around "flows", so I think we need to think some more about what it means to run a tool as standalone. It's definitely useful for testing purposes, but the way we were doing previously it was definitely awkward. I am not sure we should officially support launching tool in standalone mode?

nmoroze:
Okay, sounds good. I'll modify this PR to get rid of the new `single_step()` function, and refactor the tool tests to create flows inside the tests.

aolofsson (approved):
Looks good. Ready to merge?

nmoroze:
Yes!

PR #1002, opened 2022-04-28
Make summary properly select winners for parallel paths

nmoroze:
This PR fixes an old bug I noticed in the summary() function. 

Suppose you have a flowgraph like the following:

test_graph

Let's say cts1 is selected by ctsmin. Before, summary() would display cts1, but erroneously default to displaying place2 (or whatever the last index of the previous steps are):

```
                  place2      cts1  
 errors             0          0    
 warnings           0          0   
... 
```

This PR implements a fix that searches backwards from the last tasks in the steplist to find the whole chain of selected tasks, so it will properly display place1 and cts1 in this example:

```
                  place1      cts1  
 errors             0          0    
 warnings           0          0  
...  
```

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -2877,6 +2877,24 @@ class Chip:
                           self.logger.info(string.strip())
 
     ###########################################################################
+    def _find_leaves(self, steplist):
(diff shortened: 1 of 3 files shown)


aolofsson (approved):
Nice catch and fix!

PR #1004, opened 2022-04-28
Enable linking to schema reference manual throughout docs

nmoroze:
This PR makes a few enhancements to our documentation, all related to linking to the schema reference manual.

### Link to schema reference manual in set()/get()/add() docs

This is the simplest change. Added this due to feedback that if you just look at the API reference for these functions, it's unclear what valid arguments would be.

### Make keypaths in the schema value tables link to schema reference

Now, the entries in the "Keypath" column of the tables in the tools/pdks/flows/targets/libs directory pages are clickable links that take you to the relevant entry in the schema docs.

E.g. Clicking here:
Screenshot from 2022-04-28 15-47-30

takes you here:
Screenshot from 2022-04-28 15-47-50

### Add `:keypath:` inline command

Following on the previous change, I also added a new inline command (or "role" as Sphinx calls them) that lets you easily insert links to the schema reference manual in prose. I haven't gone through and changed everything to use this yet, but I included a few examples in this PR.

For example, if you include the following in a .rst source file: ``:keypath:`quiet` ``, it renders as `['quiet']`, and it automatically links to the [reference](https://docs.siliconcompiler.com/en/latest/reference_manual/schema.html#quiet) for the 'quiet' parameter.

The other cool thing about this is that it ensures consistent formatting of keypaths in the text as well. I've adopted what I think has been a de-facto but not 100% consistent standard of putting them inside square brackets, and added support for a couple different categories:
| Type | RST source | Processed result |
|--------|-------|----------|
| Fully-qualified | ``:keypath:`eda, yosys, exe` ``| `['eda', 'yosys', 'exe']`|
| Placeholder | ``:keypath:`eda, <tool>, exe` ``| `['eda', <tool>, 'exe']`|
| Prefix | ``:keypath:`eda` ``| `['eda', ...]` |

We can tweak the appearance of these if you'd like.

### DIFF
--- a/docs/_ext/common.py
+++ b/docs/_ext/common.py
@@ -2,6 +2,8 @@ from docutils import nodes
 from sphinx.util.nodes import nested_parse_with_titles
 from docutils.statemachine import ViewList
 
+from siliconcompiler.schema import schema_cfg
+
 # Docutils helpers
 def build_table(items):
     table = nodes.table()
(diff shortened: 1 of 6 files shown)


aolofsson:
Very cool. This is next level hyperlinking! Is this is ready to merge?

nmoroze:
Yep! I plan to go through and actually insert links in various places in the prose, but I'll save that for a future PR (didn't want to clutter the core functionality add PR with lots of changes to random .rst files)

aolofsson (approved):

PR #1010, opened 2022-05-02
Implement 'resume' option

nmoroze:
This PR implements a 'resume' option that allows you to re-run a job starting with the last failing step. One key decision was to make a step be considered failing (and be re-run) only if all its indices fail. This simplifies handling some edge cases, and if a user needs a different behavior, they can always manually re-run things using the steplist.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3974,12 +3974,13 @@ class Chip:
         else:
             steplist = self.list_steps()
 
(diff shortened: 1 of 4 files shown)


aolofsson (approved):
I feel a little uncomfortable about the isdir(), but I can't point to a scenario/test where it would definitely break, so I think this is fine.:-)

PR #1011, opened 2022-05-03
Implement 'clean' feature

nmoroze:
This PR implements 'clean'. There are three pieces of this change:

1. As we discussed offline, I got rid of `['eda', <tool>, 'copy']` instead of copying files into a `scripts/` in the workdir, since it didn't seem necessary to support this copying in general.
    - This change necessitated modifying scripts for a few tools to eliminate their dependence on copied files.
    - One tool (Chisel) didn't have an obvious way to run without copying stuff into the work directory, so I added `pre_process()` code to handle this. If this situation comes up more often, I think we could consider adding back an `['eda', tool, 'copy', ...]` keypath that is a per-task limited list of files to copy in from the refdir (rather than an all-or-nothing boolean)
2. Making 'refdir'/'script' orthogonal. 
    - In most tool drivers, we stuffed `refdir + script` into the 'script' schema parameter, but based on the description it seems like we want 'script' to store only the basename of the entry point, so that a user can supply a variant of the scripts just by changing 'refdir'. I think we may have done this to ensure that the 'script' parameter can be fully resolved, so I added another special case to `find_files()` to handle 'script', and fixed the setup in each tool driver where needed.
4. The actual 'clean' implementation itself, which is pretty simple. 
    - I did make a design change to delete less stuff than the spec originally states. The old description said only `outputs/`, `reports/`, and the log would remain, but I decided to keep everything required to replay the step. I'm open to changing it back to deleting everything but outputs/reports/logs though.


aolofsson:
Looks great! The find_files solves a lot of problems.

aolofsson (approved):

PR #1012, opened 2022-05-03
Fix wheels tests

nmoroze:
Add back is_posix to shlex.split() calls in _makecmd

I got a little too aggressive when clearing out the multi-platform
replay script code. We need to properly handle different platforms for
actually creating the command itself!

Verified back to green: https://github.com/siliconcompiler/siliconcompiler/actions/runs/2265960443


aolofsson (approved):
Yes, one more reason to avoid multi platform code if at all possible.

PR #1015, opened 2022-05-06
add -fsynopsys option to ghdl

suppamax:
As requested [here](https://github.com/siliconcompiler/siliconcompiler/discussions/917#discussioncomment-2701653)

I added the testcase under `examples`, because I am not sure how you want to organize it

### DIFF
--- a/examples/ghdl_fsynopsys/binary_4_bit_adder_top.vhd
+++ b/examples/ghdl_fsynopsys/binary_4_bit_adder_top.vhd
@@ -1,0 +1,16 @@
+library IEEE;
+use IEEE.STD_LOGIC_1164.ALL;
+use IEEE.STD_LOGIC_UNSIGNED.ALL;
+
+entity binary_4_bit_adder_top is
+Port ( NUM1 : in STD_LOGIC_VECTOR (4 downto 0) := "00000";
+NUM2 : in STD_LOGIC_VECTOR (4 downto 0) := "00000";
+SUM : out STD_LOGIC_VECTOR (4 downto 0));
+end binary_4_bit_adder_top;
+
+architecture Behavioral of binary_4_bit_adder_top is
+begin
+
+SUM <= NUM1 + NUM2;
+
+end Behavioral;
--- a/examples/ghdl_fsynopsys/build.py
+++ b/examples/ghdl_fsynopsys/build.py
@@ -1,0 +1,23 @@
+import siliconcompiler
+import os
+
+def main():
+    chip = siliconcompiler.Chip()
+    chip.set('design', 'binary_4_bit_adder_top')
+    chip.add('source', 'binary_4_bit_adder_top.vhd')
+
+    chip.load_target('freepdk45_demo')
+
+    # TODO: add in synthesis step once we can get an output that passes thru
+    # Yosys.
+    flow = 'vhdlsyn'
+    chip.node(flow, 'import', 'ghdl')
+    chip.node(flow, 'syn', 'yosys')
+    chip.edge(flow, 'import', 'syn')
+
+    chip.set('flow', flow)
+
+    chip.run()
+
+if __name__ == '__main__':
+    main()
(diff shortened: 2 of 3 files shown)


suppamax:
I see.
What would then be a way to set additional options in the build script?
I tried
```
chip.set('eda', 'ghdl', 'option', 'import', '0', '-fsynopsys')
```
but, set aside the fact that forcing `step` and `index` like this is really ugly, the `-fsynopsys` option is at the very first one, and for some reason GHDL doesn't like it.
```
| INFO    | job0  | import     | 0  | ghdl -fsynopsys --synth --std=08 --out=verilog --no-formal /siliconcompiler/examples/ghdl_fsynopsys/build/binary_4_bit_adder_top/job0/import/0/outputs/binary_4_bit_adder_top_cd49da6cfcadc1c2e0f72496cb6d853fba7f792f.vhd -e binary_4_bit_adder_top
/usr/local/bin/ghdl: unknown command '-fsynopsys', try 'help'
```

aolofsson:
> ..but, set aside the fact that forcing step and index like this is really ugly

You bring up a good point. The options are there to address tools like openroad that get called for many different steps and need the ability to drive the option on a per index/step basis. We can consider adding a wrapper function to make it nicer or some kind of global option.

 > -fsynopsys option is at the very first one, and for some reason GHDL doesn't like it.

Is this a ghdl bug?

nmoroze:
I think GHDL probably expects a command name (like `--synth`) to come first. I don't think that's a bug, just how it works.

I don't think many of our tool wrappers are carefully written to allow additional user-supplied options at this point. I think ghdl.py will need to be refactored to set up options in `setup()`, and it could make sense to see if the user has supplied any options first, and then append those to the end of the default options before overwriting the options schema entry.

suppamax:
Something like this?

nmoroze:
We'll give this a bit more thought and then get back to you!

nmoroze:
Sorry for the delay getting back! I think there are a couple things that need to change to get a really good general solution here. However, in order to resolve this in the meantime I'd propose:

Right now we have the [`['tool', <tool>, 'var', ...] `](https://docs.siliconcompiler.com/en/latest/reference_manual/schema.html#tool-var) keypath that lets you set arbitrary tool-specific key/value pairs. I think you could use this parameter to control whether the `-fsynopsys` parameter gets added. 

In your build script, to enable `-fsynopsys`:
`chip.set('tool', 'ghdl', 'var', step, index, 'extraopts', '-fsynopsys')`

That lets you do something like in `ghdl.py`:
```
if chip.valid('tool, 'ghdl', 'var', step, index, 'extraopts'): 
  extra_opts = chip.get('tool', 'ghdl', 'var', step, index, 'extraopts')
  # some validation code for extra_opts?
  # add extra_opts to CLI options in correct place
```

This is abusing this keypath a little bit, since the docs specify that the reference scripts should be consuming the vars set here (in this case, GHDL isn't running from a script). However, we don't have any other free passthrough keys in the tool schema, and I haven't decided on a proper entry in the schema. The reason to not just do this through `['tool', <tool>, 'option', ...]` is that allowing a user to manually tweak the exact CLI options list seems error-prone, since there's a certain behavior that the tool driver tries to produce, and this could be broken if the CLI flags are changed arbitrarily. Giving the user a new field to put additional options in, and then letting the tool driver validate and manage it, seems like a safer interface.

Now, this doesn't fix your point about specifying the step/index being ugly (which is a point well taken, by the way). I'm treating this is an orthogonal issue, and I'm not 100% convinced that we should be adding something else to the schema for it. At the very least, I think the issue applies to multiple values under the ['tool', ...] keypaths, so we'll want a solution that scales to each (I'm leaning towards some sort of helper function). 

TL;DR: as an incremental approach, I'd propose using `['tool', <tool>, 'var', ...]` to pass in extra CLI options, which ghdl.py could then validate and append to existing options in the correct place. The other issues you raised we can address separately.

suppamax:
Hi @nmoroze I changed the usage of `-fsynopsys` following your indication

nmoroze (approved):
Looks good to me, thanks a ton for implementing this!

PR #1016, opened 2022-05-08
Implement schema changes from todo list in schema_rearchitect

nmoroze:
This PR adds some of the small schema tweaks from our backlog into the schema_rearchitect branch. I figure now is a good time to knock these out of the way since we're refactoring the entire thing anyway. Let me know if you don't want to do any of these changes anymore and I can remove it from the PR:

- Rename `'variable'` to `'var'` (in `['pdk', ...]`, `['tool', ...]`, and `['asic', ...]`)
- Rename `['option', 'dir']` to `['option', 'build']`
  - (my 2 cents on this one is I think a rename is good, but not 100% sure about 'build'. Maybe 'builddir' or 'outdir'?)
- Remove `<job>` from `['jobinput', ...]`

### DIFF
--- a/docs/user_guide/examples/flowgraph_doe.py
+++ b/docs/user_guide/examples/flowgraph_doe.py
@@ -9,7 +9,7 @@ for index in range(len(syn_strategies)):
     chip.node(flow, 'syn', 'yosys', index=str(index))
     chip.edge(flow, 'import', 'syn', head_index=str(index))
     chip.edge(flow, 'syn', 'synmin', tail_index=str(index))
-    chip.set('eda', 'yosys', 'variable', 'syn', str(index), 'strategy', syn_strategies[index])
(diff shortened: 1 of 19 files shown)


aolofsson:
>  Rename 'variable' to 'var' (in ['pdk', ...], ['tool', ...], and ['asic', ...])

Looks good. Brevity with equal clarity.

> Rename ['option', 'dir'] to ['option', 'build']

I wouls say 'builddir' is better. It's a bit ugly, but it's not something one would use often, so clarity is more important than brevity or beauty.

> Remove <job> from ['jobinput', ...]

OK

nmoroze:
> I wouls say 'builddir' is better. It's a bit ugly, but it's not something one would use often, so clarity is more important than brevity or beauty.

Sounds good, this should be good for merge now!

aolofsson (approved):

Issue #976, opened 2022-04-07
Distinguish metric goals by flow


nmoroze:
Now that we have the ability to store multiple flows in the schema, I think we need a way to separate out metric goals for each. 

Maybe a good way to do it is in the flowgraph, like we already do for weights:

`'flowgraph', flow, step, index, 'weight', metric`
`'flowgraph', flow, step, index, 'goal', metric`

(might also make sense to reorder the metric and 'weight'/'goal'?)

Both the goals and weights are set up in the flow, while the "real" values are inserted during run(), so this separation seems more consistent to me.

nmoroze:
Fixed in #1023.

PR #1017, opened 2022-05-08
Specify metric goals in flowgraph

nmoroze:
This PR adds a fix for issue #976 into the schema_rearchitect branch. This is another thing that I figure is worth fixing if we're going to refactor everything.

As stated in that issue, I don't think it makes sense to set up metric goals inside the metric schema itself now that we have multiple flows. It seems more consistent to set up the goals alongside the weights. I think this could also lets us get rid of the "group" free key in the metric keypath, since I'm not sure what other categories we'd want to put there besides "goal" and "real".

I am still going back and forth on whether the "flowgraph" metirc-related keypaths should be:

```
'flowgraph', flow, step, index, 'weight', metric
'flowgraph', flow, step, index, 'goal', metric
```
or
```
'flowgraph', flow, step, index, metric, 'weight'
'flowgraph', flow, step, index, metric, 'goal'
```

If we want to merge this, once we've decided on the final keypaths I can go in and update all the uses of these parameters.

### DIFF
--- a/siliconcompiler/schema.py
+++ b/siliconcompiler/schema.py
@@ -949,13 +949,26 @@ def schema_flowgraph(cfg, flow='default', step='default', index='default'):
     scparam(cfg,['flowgraph', flow, step, index, 'weight', metric],
             sctype='float',
             shorthelp="Flowgraph: metric weights",
-            switch="-flowgraph_weight 'flow step metric <float>'",
+            switch="-flowgraph_weight 'flow step index metric <float>'",
             example=[
                 "cli: -flowgraph_weight 'asicflow cts 0 area_cells 1.0'",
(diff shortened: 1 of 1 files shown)


aolofsson:
Not 100% sure about this, you might be right....

- A metric goal is definitely something that the user would specify, on a personal, project, design, or run basis. 
- Flows (like asicflow) are intended to be pretty static.  
- Anything (pdk, libs, flows) can be overriden by clobbered by the user before run(), but I think this will get confusing and breaks the abstraction of.
- Inside the flows we already set flowgraphs and metrics (like in asiflow) so they are mixed up
- As you pointed out weights and goals belong together!

...ok, you sold me. :smile: Let's make the change!

aolofsson (approved):

PR #961, opened 2022-03-27
Stdout eda option

suppamax:
Following up #944 - a proposal for a new option for `eda` configuration, to notify that the tool outputs the result to STDOUT.

PR #1019, opened 2022-05-13
Skip Microwatt test: it causes CI to time out

nmoroze:
I noticed our daily CI tests started timing out after #961 got merged (which adds synthesis). It takes ~35 minutes on my machine to run. I'd say we should skip it for now to get these tests back to green, and perhaps switch over to just running the Verilog we get out of GHDL through a linter or something (since we run synthesis mainly to check that it's OK).


aolofsson (approved):
Good catch. Yes, these tests are expensive from an energy/cost perspective, definitely should be careful with the flow tests.

PR #1025, opened 2022-05-19
Update path to test_show.py in Wheels CI

nmoroze:
This is what made the Wheels CI turn red. Tests are passing, it just failed because it was trying to run a test that doesn't exist (test got moved)!


aolofsson (approved):
Great!

PR #1027, opened 2022-05-20
Fixing schema types, adding 'notes' field to _search() function code

aolofsson:


siliconcompiler/core.py
```diff
@@ -1558,7 +1558,7 @@ def _merge_manifest(self, cfg, job=None, clobber=True, clear=True, check=False,
                 # update other fields that a user might modify
                 for field in self.getdict(*keylist, cfg=cfg).keys():
                     if field in ('value', 'switch', 'type', 'require', 'defvalue',
-                                 'shorthelp', 'example', 'help'):
+                                 'shorthelp', 'notes', 'example', 'help'):
```
nmoroze, line 1561: Do we need to add this here? I just tried and tests pass without, and the behavior with this would be for notes to not get imported when you do a `read_manifest()` (not sure if that's what you intended?)

nmoroze (commented):
Nice catch, just one quick question inline!

By the way, do you want to sneak this into the 0.9.0 release, or wait for the next one?

siliconcompiler/core.py
aolofsson: Great find, fixed!

nmoroze (approved):
Looks good, thanks for fixing!

PR #1033, opened 2022-05-23
Change CLI behavior to store optmode with 'O' prefix

nmoroze:
~I think this default value is not legal, based on the docs there should be no 'O' prefix.~

Update: per our discussion this morning, seems like the intended behavior is for the optmode to be stored with an 'O' prefix. I added a special case to the `create_cmdline()` function to handle this, updated the docs to reflect the intended behavior, and added a test case.


aolofsson (approved):
Very nice!

PR #1034, opened 2022-05-24
Refactor create_cmdline()

### LINKED ISSUE #996: Command line program sc not handling spaces in fields
Place holder. Need a better way of differentiating between "keys" provided at the command line and free string arguments. 
Fails on ['package', 'description']

https://github.com/siliconcompiler/siliconcompiler/blob/fec19aecb4962f439cecdf1fca70a7e89229b9ff/siliconcompiler/core.py#L404-L413

nmoroze:
In preparation for adding filetype inference to create_cmdline() in a future PR, I took a refactor pass to fix some of our backlog CLI items.

**Improvements**

1: Simplified code by constructing argparse 'dest' key directly from keypath. This eliminated the need for `switchmap` and the logic for constructing `dest`.

2: Made `switchlist` based on actual switches. The docstring said that switchlist should match switches without the '-', but the existing behavior didn't match this. I also tweaked the spec to require the '-', since treating the hyphen specially seems inconsistent given we have keys without hyphens (e.g. `+libext+`).

3: Fixed #996: if a keypath has N 'default' keys, the logic now splits out only the first N words from the value provided on the command line, and treats the entire remainder as the value to store. 

```
sc -package_description "My package description"
...
| INFO    | Command line argument entered: ['package', 'description'] Value: My package description
```

Also added a nice error in case a user messes up the formatting of parameters with 'default':
```
$ sc -pdk_grid_xpitch 5
...
| ERROR   | Invalid value 5 for switch -pdk_grid_xpitch. Expected format 'pdkname stackup layer <float>'.
```

4: Removed ['design'], ['arg', 'step'], and ['option', 'mode'] priority, since I don't think there's a need for this anymore?

5: Added 'metavar' to argparse so that user can see the expected format for complex keypaths in the help message. Let me know if you like this.

E.g.
```
$sc-show --help
usage: sc-show [-h] [-design <str>] [-input 'filetype <file>'] [-cfg <file>] [-loglevel <str>] [-version]

...

optional arguments:
  -h, --help            show this help message and exit
  -design <str>         Design top module name
  -input 'filetype <file>'
                        Input files
  -cfg <file>           Configuration manifest
  -loglevel <str>       Logging level
  -version              show program's version number and exit
```

### DIFF
--- a/siliconcompiler/apps/sc_run.py
+++ b/siliconcompiler/apps/sc_run.py
@@ -5,12 +5,12 @@ import siliconcompiler
 def main():
     progname = "sc-run"
     chip = siliconcompiler.Chip(progname)
-    switchlist = ['cfg',
-                  'loglevel',
-                  'checkonly',
-                  'relax',
-                  'quiet',
-                  'version']
+    switchlist = ['-cfg',
+                  '-loglevel',
+                  '-checkonly',
+                  '-relax',
+                  '-quiet',
+                  '-version']
     description = """
     -----------------------------------------------------------
     Restricted SC app that accepts one or more json based cfg files
--- a/siliconcompiler/apps/sc_show.py
+++ b/siliconcompiler/apps/sc_show.py
@@ -31,7 +31,7 @@ def main():
     chip = siliconcompiler.Chip(UNSET_DESIGN)
 
     chip.create_cmdline(progname,
-                        switchlist=['design', 'input', 'option_loglevel', 'option_cfg'],
+                        switchlist=['-design', '-input', '-loglevel', '-cfg'],
                         description=description)
 
     #Error checking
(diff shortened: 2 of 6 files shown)


aolofsson:
This is a great refactor! What do you think about adding a cli test similar to test_setget, that cycles through all the cli examples in the schema.py to make sure we have good test coverage.

nmoroze:
Good idea, done! That caught a handful of typos in the schema, and revealed a slightly bigger problem around handling boolean keypaths that have a 'default' key. The simplest fix was to tweak our argparse setup to allow a user to provide explicit boolean values via the CLI, while still allowing bare boolean switches to default to true. This also frees us from the requirement that boolean options have to default to false for the CLI switch to be useful.

Example:
- `-quiet`: true
- `-quiet false`: quiet is false
- `-quiet true`: quiet is true
- `-tool_continue "yosys true"`: ['tool', 'yosys', 'continue'] is true

PR #1036, opened 2022-05-25
Add filetype inference to command line apps

nmoroze:
This PR adds a parameter to `create_cmdline()` that allows for configurable filetype inference of positional source arguments. This gives `sc` a CLI more similar to pre-0.9, where sources can be provided directly without having to prefix them with a flag. In fact, the interface is even cleaner now, since we have a mechanism to provide things like SDCs, DEFs, etc. directly as well!

The configurable `input_map` allows apps to set up mappings in a context-dependent manner. For example, `sc` maps all filetypes accepted by the asicflow and fpgaflow, whereas `sc-show` maps filetypes that are displayable in KLayout.

### DIFF
--- a/examples/blinky/run.sh
+++ b/examples/blinky/run.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
-sc  -input "verilog blinky.v" \
+sc blinky.v icebreaker.pcf \
     -fpga_partname "ice40up5k-sg48" \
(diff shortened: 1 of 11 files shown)


aolofsson (approved):
Looks good. We should think about ading the file map to the schema. With time there will be many more file extensions (.cpp, .py, ...) soe keeping a map in each app won't scale very well.

Issue #1022, opened 2022-05-17
Running the quickstart guide


arlted:
Per your recommendation, i followed the quick start guide to run the "heartbeat" example. 
Initially,  got the following error after executing "python heartbeat.py" in python virtual environment.  

"ERROR   | job0    | syn          | 0   | File ../third_party/pdks/virtual/freepdk45/libs/nangate45/r1p0/lib/NangateOpenCellLibrary_typical.lib was not found"

It looked like the "third_party" directory was missing in the "venv/lib/python3.6/site-packages" folder.  Then, i copied the "third_party" directory over the ./site-packages" folder to keep things running (this was not part of the guide). 

After running syn, floorplan, placement and cts steps, the execution stopped after giving out the following error
"[ERROR STA-0406] global_route positional arguments not supported. Error: sc_route.tcl, 36 STA-0406"

Am I missing something in the guide?

nmoroze:
For the first problem, that's a good workaround! The intended solution is to clone the repo and then create an environment variable `SCPATH` that points to `<path-to-repo>/siliconcompiler`. Sorry this wasn't stated in the documentation, I'll add it and leave this issue open till then.

For the second problem, I imagine it's an OpenROAD version issue. Which one do you have installed? (`openroad -version`)

arlted:
Thanks!!  The openroad version that I have is "v2.0-1644-g438dbb4".
________________________________
From: Noah Moroze ***@***.***>
Sent: Friday, May 20, 2022 3:49 PM
To: siliconcompiler/siliconcompiler ***@***.***>
Cc: Nigussie, Theodros B CIV USARMY DEVCOM ARL (USA) ***@***.***>; Author ***@***.***>
Subject: [URL Verdict: Neutral][Non-DoD Source] Re: [siliconcompiler/siliconcompiler] Running the quickstart guide (Issue #1022)


All active links contained in this email were disabled. Please verify the identity of the sender, and confirm the authenticity of all links contained within the message prior to copying and pasting the address to a Web browser.


________________________________



For the first problem, that's a good workaround! The intended solution is to clone the repo and then create an environment variable SCPATH that points to <path-to-repo>/siliconcompiler. Sorry this wasn't stated in the documentation, I'll add it and leave this issue open till then.

For the second problem, I imagine it's an OpenROAD version issue. Which one do you have installed? (openroad -version)

nmoroze:
[...]

PR #1037, opened 2022-05-25
Docs fixups

nmoroze:
This PR fixes up some issues with the docs.

- Follow-up tasks from 0.9 upgrade
  - Get rid of "root" in Schema reference manual: all top-level items are now directly under "1.3 Parameters". 
  - Get rid of empty "history" and "library" sections, and replace with a new prose section describing these two keypaths.
  - Document min/typical/max for parameters that use this format under datasheet schema
- Tasks from public issues
  - #1022: specify that user must clone repo/set up SCPATH to run quickstart locally (thanks @arlted!)
  - #1035: fix errors in library user guide (thanks @ValeriiPlumerai!)

### DIFF
--- a/docs/_ext/schemagen.py
+++ b/docs/_ext/schemagen.py
@@ -14,23 +14,7 @@ class SchemaGen(SphinxDirective):
 
         schema = schema_cfg()
 
-        # Split up schema into root and nested keys so that we can add a custom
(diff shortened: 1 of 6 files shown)


aolofsson (approved):
Great!

PR #1039, opened 2022-05-31
Enhance TCL manifest generation

nmoroze:
This PR makes a couple changes to how TCL code is generated by `write_manifest()`. It addresses a few problems:

**Escaping strings**
We weren't escaping any special characters in strings besides semicolons. This became problematic when I had a multi-line value for a field. This caused a tool to error out after sourcing the sc_manifest.tcl file, because SC generated invalid TCL like this:
```tcl
[dict set $sc_cfg package description [list Description line 1
Description line 2]]
```

Along with this, spaces and special characters like '$' would have resulted in issues. The best solution I found was to wrap the entire string in quotes, and escape a few special characters, based on  https://stackoverflow.com/a/70082148 (and verified by looking at the TCL spec).

**Dump tuple as list**
We were previously dumping tuples as tuple-like strings (e.g. `"(1, 2)"`), which generally required regex parsing to consume. I think it makes more sense to dump these directly as a list: `[list 1 2]`, since it's easier to manipulate in the TCL code. 

**Fix envvar substition**
We were previously only matching `$` at the beginning of paths, but I think we should do a search for "$" and replace it with "$env()" anywhere in the paths.

I also added a self-checking test for this function - turns out that a Python interface to TCL is available as a standard library! (although it requires an extra package installed on the system). I wonder if there are other ways we can take advantage of this...

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1479,27 +1479,37 @@ class Chip:
         if val is None:
             return ''
         elif typestr.startswith('('):
+            # Recurse into each item of tuple
             subtypes = typestr.strip('()').split(',')
             valstr = ' '.join(self._escape_val_tcl(v, subtype.strip())
                               for v, subtype in zip(val, subtypes))
             return f'[list {valstr}]'
         elif typestr.startswith('['):
+            # Recurse into each item of list
             subtype = typestr.strip('[]')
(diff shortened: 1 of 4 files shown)


aolofsson:
Makes sense! This was an area where took some short cuts, I am happy you solved it properly. Ready to merge?

nmoroze:
Yup, good to go!

aolofsson (approved):

PR #1040, opened 2022-06-06
Fix up CSV manifest export

nmoroze:
This PR fixes up CSV manifest export. The existing implementation was in pretty good shape, but the output was missing newlines. I decided it might be best to refactor the function to use the built-in CSV library though, since that has configurable logic for switching delimiters, quoting values that contain special characters, etc.. For now, I'm just using the default options, which seem to work well in both LibreOffice and Google Sheets.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -31,6 +31,7 @@ import uuid
 import shlex
 import platform
 import getpass
+import csv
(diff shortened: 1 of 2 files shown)


aolofsson (approved):
Nice, almost always a good to use a mature package if one is available.:-)

PR #1041, opened 2022-06-07
Fix up ['tool', <tool>, 'regex', ...] help message

nmoroze:
- Make list of supported options correct
- Clean up some formatting
- Remove accidental inclusion of licenseserver help

One thought: we might want a schema field to separately document current feature support at an implementation level. I'm not sure how you'll feel about trimming the list of supported options since the schema is supposed to be an aspirational/implementation-agnostic spec, but I think having the list not match the true set of supported features is user-unfriendly.


aolofsson (approved):
The schema is the spec. If we want to doscument supported/unsupported features this should be captured in the implementation. Possibly in the main siliconcompiler Class? I think we possiblydiscussed having a parameter with "unsupported", but I can't remember why we dismissed it?

PR #1042, opened 2022-06-08
Adjust behavior of ['tool', <tool>, 'continue'] to always fail on fatal error

nmoroze:
I think the current behavior of `['tool', <tool> ,'continue']` is unideal. It's necessary in some cases because your tool produces nonfatal errors, but even in these cases I think you don't want the flow to continue running if the tool errors out with a nonzero status, since in that case future steps are probably pretty likely to fail.

This PR proposes the simplest possible change, which is to always kill the flow if the tool exits with a nonzero status. Another proposal that I'd be happy to implement would be to make 'continue' take a few possible values, something like:
- "never": always quit on any error condition
- "nonfatal": only quit on fatal errors (always continue if tool quits successfully)
- "always": flow always continues

### DIFF
--- a/siliconcompiler/client.py
+++ b/siliconcompiler/client.py
@@ -54,11 +54,8 @@ def remote_preprocess(chip):
     for index in indexlist:
         tool = chip.get('flowgraph', flow, local_step, index, 'tool')
         # Setting up tool is optional (step may be a builtin function)
-        if tool:
-            chip.set('arg', 'step', local_step)
-            chip.set('arg', 'index', index)
-            func = chip.find_function(tool, 'setup', 'tools')
-            func(chip)
+        if tool and tool not in chip.builtin:
+            chip._setup_tool(tool, local_step, index)
 
         # Need to override steplist here to make sure check_manifest() doesn't
         # check steps that haven't been setup.
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -4120,18 +4120,22 @@ class Chip:
             sys.exit(1)
         func(self)
 
+        # Add logfile as a report for errors/warnings if they have associated
+        # regexes.
         if self.valid('tool', tool, 'regex', step, index, 'default'):
             re_keys = self.getkeys('tool', tool, 'regex', step, index)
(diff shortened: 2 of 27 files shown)


aolofsson:
Are we interpreting this as continue on error on 'self.error' or an exeuctable error code. There is a fair number of tools (open/closed) that supports this behavior, but I agree that there might be room for more than a binary 1/0 choice.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

From Make:
  -i, --ignore-errors
            Ignore all errors in commands executed to remake files.

nmoroze:
I hadn't considered the self.error case, but the current implementation of this parameter is that it applies to both the "errors" metric (otherwise we always quit if a tool post_process() finds errors in the logfile) as well as executable error code. If we want to have it be able to apply to both, I think a third option is useful, since there are cases where I'd want the flow to continue if the tool reports errors but a zero return code (generally indicating nonfatal errors), but quit early if the tool returns a different return code (generally indicating fatal errors).

nmoroze:
- tool, continue should be fed into tool directly
- always quit if tool returns error code
- separate switch for controlling flow continue behavior

nmoroze:
Working on addressing the changes from our offline discussion now, and I've come across a few decisions that I think could warrant discussion:

- Before merging, I plan to add an option back to the error check at the end of `_runtask()`. One question though: should we make this error check a special case, or should we perhaps broaden it to a check of all flowgraph goals? I could see arguments either way, since errors clearly seem like a special case to me. However, using the goals could make things a little more flexible/configurable, and they aren't currently checked by anything except min/max functions right now.
- Should we keep the return codes in `post_process()`? It looks like all our tools return zero right now, and they can always increment the error metric on error (which can be used to kill the flow depending on configuration), and they can throw an exception for something truly fatal. Not sure what the use case of this third error signalling mechanism is.
- We mentioned `self.error` briefly, so I spent some time thinking about what seems most intuitive there. I opened a separate issue (#1054) with my thoughts on that since it feels like a whole other can of worms.

PR #1043, opened 2022-06-08
Noah/manifest view

nmoroze:
This PR adds a pane to the HTML report that gets generated by chip.summary() that displays the final manifest as an interactive, collapsible tree. 

Example for GCD:
Screenshot from 2022-06-07 22-27-45


aolofsson:
Nice POC!

aolofsson (approved):

PR #1045, opened 2022-06-10
Make report self-contained

nmoroze:
This PR makes SC's generated HTML report a self-contained file that will look the same even if it is copied elsewhere in the filesystem (or sent to someone else).

This required two changes:
- Embedding the bootstrap CSS/JSS
- Embedding the layout preview as a base64 encoded image

I think this is a nice change, the main downside is that this makes the file a bit heavier. However, it looks like the size of the HTML dump of the entire manifest (for the tree viewer) dominates the size - these changes only result in an increase from 3.5MB->3.9MB for the GCD example.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -2588,6 +2588,7 @@ class Chip:
         design = self.get('design')
         jobname = self.get('option', 'jobname')
         buildpath = self.get('option', 'builddir')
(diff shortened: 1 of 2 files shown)


WRansohoff (approved):
Great idea!

aolofsson:
Nice idea!

PR #1046, opened 2022-06-10
Make chip.show() file search more flexible

nmoroze:
This PR adjusts chip.show()'s behavior when a filename isn't explicitly provided. Its previous behavior was to search for and show a {design}.gds or {design}.def file in export's outputs, if either exists. 

This ends up missing a lot of cases, however. There are many possible layout formats (and many possible filename variations): .gds.gz, .DEF, .oas, .OASIS, etc., and hardcoding all of these doesn't scale well. 

Now that we're moving towards registering flow outputs in the schema, we can take advantage of these new keypaths to make the lgoic more generic. This PR's approach is to search through all entries in the ['output', ...] keypath, and displays the first file that is "showable" (i.e. has an associated entry in showtool). 

Since we have pretty much no examples that add anything to ['output', ...], I kept the old search code to make this change fully backwards compatible with existing examples. The new search only occurs if the old method doesn't find a match.

2 future enhancements:
- Think about a standard keypath for final layouts? Searching through all outputs is kind of broad, and might hit on something like a netlist if a showtool is defined for that. I wanted to use ['model', 'layout', ...], but this includes abstract layouts (LEF), and the goal here is to find final layouts.
- Determine a methodology for filling out ['output', ...] (which ties into automated packaging). Right now this could be done in the build script, but might be nice to automate it? I think it might be effective to put it in the flow definition, which should have all information required to do so (knowledge of which steps produces which outputs). 

This change also let me make the layout preview in the HTML report generation more flexible -- it was previously hardcoded to look for GDS.


aolofsson:
I think this one warrants some more discussion. I agree that the automatd search has issues with scaling, especially when we start looking at job histories, multiple flows, looking at intermediate files etc. This might be a case where we either provide the full file path or a very limited number of cases without filepath. The main advantage of this use case is to set up the layer files and tech files for things like klayout. This is the non trivial aspect. Asking folks to find the path to the file inside the job directory is a small effort compared to the layer file. I would put the autofind as a nice to have.

nmoroze:
I agree it's a nice-to-have, and it seems OK to have the user provide a path in general. The main benefit I saw here was for the HTML report, since this also contains its own auto-search logic, and there's no interface to provide a different layout filepath to that code at the moment.

I do think once we have packaging figured out, the autosearch should be trivial (seems likely you'd want any package to register its final layout in a standard schema parameter).

aolofsson (approved):

PR #1047, opened 2022-06-10
Return filepath from chip.archive()

nmoroze:
Just a small API tweak that I think is convenient -- make chip.archive() return the full path to the generated tarball (useful for scripts that call chip.archive() and then want to do something with the result).


aolofsson (approved):
Nice, are there other functions that might benefit from a return value?

PR #1048, opened 2022-06-10
Add final tool CLI options to record schema

nmoroze:
I realized we should probably be recording the final set of CLI arguments used to drive a tool for provenance purposes. The options parameter in the tool schema isn't complete, since we generate arguments on the fly using `runtime_options()`.


aolofsson (approved):
Nice catch!

PR #1049, opened 2022-06-13
Update OpenROAD tool file to better support non-default flows

WRansohoff:
This PR contains a few changes to support customized SiliconCompiler flows which use OpenROAD.

It moves some logic behind `if flow == 'asicflow'` checks, fixes a couple of small metric-parsing issues, and fetches the output DEF filepath from the schema.

I'm open to suggestions on the flow name checks; it may not be unreasonable to expect flows to use a specific naming convention for files used with certain tools.

### DIFF
--- a/siliconcompiler/tools/openroad/openroad.py
+++ b/siliconcompiler/tools/openroad/openroad.py
@@ -78,19 +78,20 @@ def setup(chip, mode='batch'):
 
     chip.set('tool', tool, 'threads', step, index, threads, clobber=clobber)
 
-    # Input/Output requirements
-    if step == 'floorplan':
-        if (not chip.valid('input', 'netlist') or
-            not chip.get('input', 'netlist')):
-            chip.add('tool', tool, 'input', step, index, chip.design +'.vg')
-    else:
-        if (not chip.valid('input', 'def') or
-            not chip.get('input', 'def')):
-            chip.add('tool', tool, 'input', step, index, chip.design +'.def')
+    # Input/Output requirements for default asicflow
+    if chip.get('option', 'flow') == 'asicflow':
+        if step == 'floorplan':
+            if (not chip.valid('input', 'netlist') or
+                not chip.get('input', 'netlist')):
+                chip.add('tool', tool, 'input', step, index, chip.design +'.vg')
+        else:
+            if (not chip.valid('input', 'def') or
+                not chip.get('input', 'def')):
+                chip.add('tool', tool, 'input', step, index, chip.design +'.def')
 
-    chip.add('tool', tool, 'output', step, index, chip.design + '.sdc')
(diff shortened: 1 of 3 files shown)


nmoroze (commented):
Thanks Will! Nice catches and cleanup in here. Left some comments inline.

I feel like I'd prefer not to have the flow name checks, but I don't have a suggestion I'm 100% happy with.

My issue with it is that I view the tool driver as a more fundamental element than the flow, with its interface defined as a set of supported steps that have a specific input/output behavior. The flow is then responsible for stitching together these steps in a way that makes sense, based on this behavior. Making the tool driver aware of the flow seems like information is flowing backwards between layers of abstraction.

A couple solutions I see:
- Forking the tool driver. I don't really like this due to repeated code, but if one needs to tailor a driver to a flow, bundling that driver with the flow seems somewhat intuitive. I think a longer-term version of this solution would be to re-architect things to keep common logic in a core driver (exe name, version checking, metrics processing); and then somehow have a mechanism to allow factoring out flow/step-specific things (other pre/post processing, input/output requirements) into separate files. 
- Keep orthogonal step names: If possible, my preferred "simple" solution would be to avoid repeating step names between our reference flow and the OpenROAD reference flow, although I'm not sure if that breaks something else in our scheme to replicate their Makefile (perhaps something as simple as adding an "_or" suffix could work).

WRansohoff:
Okay, those both sound like good options - I think I'd prefer to use orthogonal step names. I'm not a huge fan of hard-coding names, but you're right that it's probably better to do that with steps instead of flows, and it seems better than maintaining multiple copies of the driver.

Long-term, I like the idea of pre/post-processing steps which can be overridden. There are definitely cases where it would be nice to run some custom Python logic between tasks in a flow.

nmoroze (approved):
Thanks for addressing the feedback, these changes look great to me!

siliconcompiler/tools/openroad/openroad.py
```diff
@@ -78,19 +78,20 @@ def setup(chip, mode='batch'):
...
-            chip.add('tool', tool, 'input', step, index, chip.design +'.vg')
-    else:
-        if (not chip.valid('input', 'def') or
-            not chip.get('input', 'def')):
-            chip.add('tool', tool, 'input', step, index, chip.design +'.def')
-
-    chip.add('tool', tool, 'output', step, index, chip.design + '.sdc')
-    chip.add('tool', tool, 'output', step, index, chip.design + '.vg')
-    chip.add('tool', tool, 'output', step, index, chip.design + '.def')
+    # Input/Output requirements for default asicflow steps
```
WRansohoff, line 81:
```suggestion
    # Input/Output requirements for default asicflow steps
    # TODO: long-term, we want to remove hard-coded step names from tool files.
```

PR #1053, opened 2022-06-16
Fix Verilator compile support, clean up driver

nmoroze:
This PR fixes up the Verilator "compile" step, working towards supporting a Verilator-based simulation flow. Thanks to @sgherbst for trying out this functionality and providing feedback!

One design decision I wanted to highlight was making the input to "compile" an already preprocessed and pickled Verilog file, rather than reading input/include dirs/defines from the schema. This might seem odd given that Verilator can handle the preprocessing itself, but I did this for a few reasons:

1) We have standardized on Surelog for input preprocessing and elaboration, so a simulation pipeline that goes Surelog --> Verilator ensures any Surelog-related quirks that would be present in the implementation RTL will also be present in the simulation RTL. The Verilator driver still supports being used for "import" so we can also swap the two out for A/B debugging.
2) This makes it easier to support simulation of alternate frontends, which all generally produce a pickled Verilog file as output. 
3) It's more consistent with the intended interface for the "lint" step (based on flows/lintflow.py), although I had to fix the implementation of lint to make this the case.

I also took this chance to do a sweep of the driver and clean things up according to our current best-practices. Highlights:
- More precise docs
- Thoughtful use of clobber
- Set logfile as report for errors/warnings
- Using runtime_options() for file-based options only

### DIFF
--- a/siliconcompiler/tools/verilator/verilator.py
+++ b/siliconcompiler/tools/verilator/verilator.py
@@ -88,8 +88,8 @@ def setup(chip):
     chip.set('tool', tool, 'threads', step, index,  os.cpu_count(), clobber=False)
 
     # Basic warning and error grep check on logfile
-    chip.set('tool', tool, 'regex', step, index, 'warnings', "\%Warning", clobber=False)
-    chip.set('tool', tool, 'regex', step, index, 'errors', "\%Error", clobber=False)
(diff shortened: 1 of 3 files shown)


aolofsson (approved):
Looks great!!

PR #1057, opened 2022-06-21
Use tool regexes to populate errors/warnings

nmoroze:
This PR switches the core logic to use the errors/warnings parsed from tool log files by the regexes placed in `['tool', <tool>, 'regex', ...]` as the single source of truth for those metrics.

A few things to highlight:
- I also added some logic to add the logfiles to the ['tool', <tool>, 'report', ...] schema if a regex is defined, since this helps cut down on tool driver boilerplate
   - I'd have done this around `check_logfile()` itself, but those changes would not be reflected in the final manifest, so I added it in setup() instead
- I tweaked the `check_logfile()` logging to map errors/warnings to the correct log type (I can undo this if you'd like, but it seemed nice)
- I made the errors/warnings additive since some tools report warnings from sources other than the logfile (in particular, I made the Netgen driver count a certain type of LVS mismatch that's often a false positive as a warning). So if post_process() sets 3 warnings from the LVS report, and the regex parses 2 warnings from the logfile, the overall warning count is 5. Does that sound alright?

I've updated the core asicflow tools in this PR, I'll do another PR to clean up all the tool drivers.


aolofsson (commented):
Nice change!

PR #1059, opened 2022-06-22
Clean up tool drivers

nmoroze:
- Delete unused imports/variables
- No clobber=False on exe/vswitch, clobber=False on threads/script/refdir/version
- post_process():
  - Delete if empty
  - Remove return 0
  - Remove error/warning processing (use regex)


aolofsson:
Nice clean up!

PR #1060, opened 2022-06-22
Allow tool tasks to have more than one input

nmoroze:
This PR modifies the flowgraph I/O checker to allow tool tasks to have more than one input. As far as I can tell, the logic in `_runtask()` supports this (it works intuitively -- the outputs of each input task are combined into the inputs of the task at hand). However, the checker did not, and would output a nonfatal error.

I've updated the checker to no longer consider this an error, and also properly check the inputs (so it will throw an error if there are  duplicate input files, and it will throw an error if the sum total of all inputs does not provide all the files a task needs). I added tests for these cases.

I also bundled a small tweak to the OR post_process function that checks that the [... 'output' ...] keypath is valid. This cleans up the other nonfatal error we're seeing in the SC+ORFS project.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1937,10 +1937,8 @@ class Chip:
 
                 # Get files we receive from input tasks.
                 in_tasks = self.get('flowgraph', flow, step, index, 'input')
-                if len(in_tasks) > 1:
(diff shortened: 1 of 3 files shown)


aolofsson (approved):
I think this is ok, but in order to support truly modular flow graphs, we will need to make sure it all works.

PR #1064, opened 2022-06-27
Make Python error handling more consistent

### LINKED ISSUE #1054: `self.error` behavior
While discussing PR #1042 offline, we briefly touched on the idea of adding an option to control SC's handling of `self.error`. I'm working on addressing this PR now, and I wanted to offer an alternative, since I think there are a few downsides to error handling using the current error bit mechanism:

- Follow-on errors
  - Core API functions that return data (`get()`, `find_files()`), will set the error bit and return `None` on error, which can cause fatal follow-on errors if the `None` gets used. This makes debugging harder - it's generally easier to quickly figure out the problem when the initial error is fatal.
- No checking late in `_runtask()`
  - Any problems that set the error bit after preprocess() is run get ignored (won't be fatal). This means there might be something weird going on, but a user won't know unless they go ahead and inspect the logs closely
- Tests miss errors
  - I've noticed bugs in core API tests that cause functions to log errors, but since the tests aren't checking the error bit, the problem doesn't get caught
  - We could add an assert to every test that checks the error bit, but this increases boilerplate and still prone to bugs (if this check is missed)
- Not very intuitive in REPL
  - If you're working in an interactive prompt and you make a mistake (say typo in `chip.get()`), this will set the error bit. A user would probably go ahead and fix this (re-run without the typo), but if they call `chip.run()`, it will fail due to error bit being set regardless.
- Used inconsistently
  - Some tool drivers (OpenROAD, KLayout), will set `chip.error`, but many just throw exceptions
  - `_runtask()` checks the error bit after calling `preprocess()`, but it checks the value returned by `postprocess()` (and not the error bit)

I do think there's one advantage to the error bit, which is that it allows you to see and correct multiple errors at once, rather than hitting a single fatal error, fixing it, re-running to hit another one, fixing it, etc. Personally, I generally prefer to fix/re-run one error at a time, but I think that's mainly a personal preference that shouldn't be prescribed. (I'm also curious if there was any other reasoning behind handling errors this way that I've missed). 

My proposal: 

First of all, we should make sure error handling is consistent, no matter what we decide is best (I think we can all agree on this)! The caveat is that we may want to have different classes of errors that are handled in different ways (but it would be good to have a clear idea of what these are). 

My proposal itself is to get rid of `self.error`. Instead, almost every API call that currently sets `self.error` should throw a `SiliconCompilerError` (or some more specific error that inherits from it). This will be fatal in scripts (unless explicitly handled), and can be addressed in real time in an interactive prompt. If a core API function hits an error in a test, this will automatically cause the test to fail (unless explicitly waived). 

The exception will be `check_manifest()`. This is where I think it is useful to see multiple errors in one go. However, instead of setting the error bit, I think it should just log each error and return `False` if it finds any, and it should return `True` if it finds none. The call to `check_manifest()` inside `run()` should then throw a fatal error if `check_manifest()` returns false. 

This feels like the best of both worlds to me, since the checker will give you a full log of setup errors that aren't fatal up until that point, but anything that doesn't make sense (like reading/writing to a keypath that doesn't exist), will cause the program to exit on the spot. The other idea I'd entertain (which I think was being suggested when we discussed the PR last week) is to make an option that lets you select whether an error sets the error bit or throws a fatal exception. I'm open to that as well, although not 100% convinced the extra complexity is worth it.

nmoroze:
This PR refactors our Python error handling. It adjusts the default behavior of most errors to be fatal (throwing a `SiliconCompilerError`), while making it so you can adjust a schema parameter to make some errors try to continue (just logging an error and setting an internal flag).

To do this cleanly (and make it easy to tweak this behavior in the future), I added a public API method called `chip.error()` that takes in a message and raises an error. I tweaked the internal flag to be called `chip._error`.

I also tweaked `check_manifest()` and related checker functions to return a bool indicating success (as described in its docstring -- in this case the implementation didn't match the spec). None of the errors in the checker are considered fatal, but a failed check will cause execution to halt in `run()`. 

I think this default behavior is better since it reduces confusing follow-on errors, catches hidden issues in our tests (hence a few of those being changed in this PR), and makes interactive sessions easier. However, when debugging things like setup scripts it can be useful to catch many keypath typos in one run (for example), so having the 'continue' option be available is helpful for those sorts of cases.

Closes #1054.

### DIFF
--- a/examples/heartbeat_padring/build.py
+++ b/examples/heartbeat_padring/build.py
@@ -71,7 +71,7 @@ def build_top():
     lib = Chip(libname)
     lib.set('model', 'layout', 'lef', stackup, 'floorplan/heartbeat.lef')
     lib.set('model', 'layout', 'gds', stackup, 'heartbeat.gds')
-    lib.set('output', 'netlist', 'verilog', 'heartbeat.vg')
+    lib.set('output', 'netlist', 'heartbeat.vg')
     chip.import_library(lib)
     chip.set('input', 'def', 'floorplan/heartbeat_top.def')
 
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -78,7 +78,7 @@ class Chip:
         # Local variables
         self.scroot = os.path.dirname(os.path.abspath(__file__))
         self.cwd = os.getcwd()
-        self.error = 0
+        self._error = False
         self.cfg = schema_cfg()
         # The 'status' dictionary can be used to store ephemeral config values.
         # Its contents will not be saved, and can be set by parent scripts
@@ -414,8 +414,7 @@ class Chip:
                     # fill in 'default' keys.
                     switches, metavar = self._get_switches(*keypath)
                     switchstr = '/'.join(switches)
-                    self.logger.error(f'Invalid value {item} for switch {switchstr}. Expected format {metavar}.')
-                    raise SiliconCompilerError('Invalid CLI arguments')
+                    self.error(f'Invalid value {item} for switch {switchstr}. Expected format {metavar}.', fatal=True)
 
                 # We replace 'default' in keypath with first N words in provided
                 # value. Remainder is the actual value we want to store in the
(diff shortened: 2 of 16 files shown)


aolofsson (approved):
Good change!!

PR #1075, opened 2022-07-06
Implement ['option', 'entrypoint']

nmoroze:
This PR implements the ['option', 'entrypoint'] parameter, allowing a user to select a different top-level module/main entrypoint for builds. 

An example use case for this parameter is verification. You may have a top-level sim wrapper for a design written in Verilog, and you want that to be the entry point for simulation. However, the design name needs to stay the same so that the simulation job can be recorded in the manifest alongside the RTL2GDS job.

The kernel of this change is a new core API function called `get_entrypoint()`. It returns the name of the top-level module based on the ['option', 'entrypoint'] and ['design'] settings, as specified. (I'm not 100% sold on this name -- since it's used so much, I think a more concise alternative like `chip.top()` or `chip.get_top()` might be better)

 I also expanded `sc_manifest.tcl`'s header to add a proc that duplicates this logic. I think we've discussed providing a library of TCL helpers before, and I figured this would be a reasonable way to start.

Otherwise, the change entailed going through and replacing instances of `chip.get('design')` and `chip.design` with calls to `chip.get_entrypoint()` where needed. One thing that made this change a bit tricky is that we don't want to change over all instances:

- I figure that the build directory and manifest name still want to be based off 'design', for packaging and cradle-to-grave manifest reasons
  - Where possible, I tried to refactor things to use `_getworkdir()` or `find_result()`, which generally makes for cleaner code and easier refactoring going forward
- Everything else, like input/output filenames, should be based on `get_entrypoint()` from here on out


nmoroze:
Rename `get_entrypoint()` to `top()`

aolofsson (approved):
This is great! Veriloator has a --top switch so there is precedence :smile:

PR #1079, opened 2022-07-08
Prepare for 0.9.2 release

nmoroze:
A bit of a hodge-podge PR, with all the small things we discussed today:

- Change `sc` app to use Skywater as default target
  - Also made the default coremargin smaller, since it was particularly large compared to the small heartbeat die
- Made `sc-show` use a dark background for KLayout
- Reverted the copy of heartbeat.v in the user guide directory back to a real file from a symlink, in order to fix our business cards...
- Release notes and version bump to prep for a 0.9.2 release


aolofsson (approved):

aolofsson:
Very nice!

PR #1082, opened 2022-07-10
Fix missing 'import' in Caravel example steplist

nmoroze:
Small issue I noticed when I went to run this myself - I think the steplist needs "import" added.


WRansohoff (approved):
Good catch, thanks!

I guess I didn't notice this because the heartbeat example only has one file, and the `run_chip_flow` method is a recent workaround to avoid modifying the OpenROAD tool driver.

PR #1083, opened 2022-07-11
Set PATH env var in the replay script even if 'path' parameter is not defined.

WRansohoff:
Sometimes it is necessary to use different versions of the same tool for different tasks. Our current `replay.sh` scripts will not work in that case, unless we include the value of the `$PATH` variable from when the task was originally run.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -4821,6 +4821,8 @@ class Chip:
                 envvars[item] = ':'.join(license_file)
         if self.get('tool', tool, 'path'):
             envvars['PATH'] = self.get('tool', tool, 'path') + os.pathsep + '$PATH'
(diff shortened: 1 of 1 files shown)


aolofsson:
Not sure about this solution...seems like we should move back to having path on a per step/index basis. I had not anticipated folks using different executables for different steps. We may need to move the version number back to a per step basis as well? @nmoroze

nmoroze:
This is based on some feedback we got, so we can go back to try and better understand the use case and come up with a good solution. I do see an argument for leaning into the schema more and making path and version be task-indexed.

One thing that I do recall coming up though is that some tools will shell out to other tools internally, and for that case it's useful to capture the full path for consistent replay.

PR #1084, opened 2022-07-12
Improve error handling for some kinds of chip.get / chip.getkeys failures.

WRansohoff:
I've run into a few situations where it has been difficult to debug the source of an error, because a generic exception is thrown at [this line](https://github.com/siliconcompiler/siliconcompiler/blob/fcbf3f7f75b786f0577311fc0714d27db0b1faa5/siliconcompiler/core.py#L796) in a `chip.get` or `chip.getkeys` call:

    keypathstr = ','.join(keypath)

The errors typically come from a `None` value in the keypath which gets passed into the method, along the lines of:

    TypeError: sequence item 1: expected str instance, NoneType found

Usually, this happens because a previous `chip.get` call received a value of `None` when the build script was expecting a valid string. The easiest way to observe this kind of error is:

    import siliconcompiler
    siliconcompiler.Chip('my_design').run()

Tracking down the unexpectedly empty value from a generic Python error is difficult, so I think we should add some checks to make sure that the keypath is valid at the top of the `chip.get` and `chip.getkeys` functions.

With this change, the invalid key path will be printed out, which should help the user determine which value is missing. In the above example, the invalid path is `('flowgraph', None)`, which points towards the root cause that no EDA flow was selected.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -909,7 +909,7 @@ class Chip:
             if not isinstance(key,str):
                 self.logger.error(f"Key [{key}] is not a string [{args}]")
 
-        keypathstr = ','.join(args[:-1])
+        keypathstr = f'{args[:-1]}'
         all_args = list(args)
 
         # Special case to ensure loglevel is updated ASAP
@@ -953,7 +953,7 @@ class Chip:
             if not isinstance(key,str):
                 self.logger.error(f"Key [{key}] is not a string [{args}]")
 
-        keypathstr = ','.join(args[:-1])
+        keypathstr = f'{args[:-1]}'
         all_args = list(args)
 
         self.logger.debug(f'Appending value {args[-1]} to [{keypathstr}]')
@@ -1011,6 +1011,7 @@ class Chip:
                 self.error(
                     f'Invalid keypath: {keypath}\n'
                     'Your Chip configuration may be missing a parameter which is expected by your build script.')
+                return None
 
         #set/add leaf cell (all_args=(param,val))
         if (mode in ('set', 'add')) & (len(all_args) == 2):
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -795,7 +795,7 @@ class Chip:
 
         keypathstr = f'{keypath}'
 
-        self.logger.debug(f"Reading from [{keypathstr}]. Field = '{field}'")
+        self.logger.debug(f"Reading from {keypathstr}. Field = '{field}'")
         return self._search(cfg, keypathstr, *keypath, field=field, mode='get')
 
     ###########################################################################
@@ -904,11 +904,6 @@ class Chip:
         if cfg is None:
             cfg = self.cfg
 
-        # Verify that all keys are strings
-        for key in args[:-1]:
-            if not isinstance(key,str):
-                self.logger.error(f"Key [{key}] is not a string [{args}]")
-
         keypathstr = f'{args[:-1]}'
         all_args = list(args)
 
@@ -916,7 +911,7 @@ class Chip:
         if len(args) == 3 and args[1] == 'loglevel' and field == 'value':
             self.logger.setLevel(args[2])
 
-        self.logger.debug(f"Setting [{keypathstr}] to {args[-1]}")
+        self.logger.debug(f"Setting {keypathstr} to {args[-1]}")
         return self._search(cfg, keypathstr, *all_args, field=field, mode='set', clobber=clobber)
 
     ###########################################################################
@@ -948,15 +943,10 @@ class Chip:
         if cfg is None:
             cfg = self.cfg
 
-        # Verify that all keys are strings
-        for key in args[:-1]:
-            if not isinstance(key,str):
-                self.logger.error(f"Key [{key}] is not a string [{args}]")
-
         keypathstr = f'{args[:-1]}'
         all_args = list(args)
 
-        self.logger.debug(f'Appending value {args[-1]} to [{keypathstr}]')
+        self.logger.debug(f'Appending value {args[-1]} to {keypathstr}')
         return self._search(cfg, keypathstr, *all_args, field=field, mode='add')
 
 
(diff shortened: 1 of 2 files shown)


siliconcompiler/core.py
```diff
@@ -787,6 +787,15 @@ def get(self, *keypath, field='value', job=None, cfg=None):
 
         """
 
+        # Ensure that all keypath values are strings.
+        # Scripts may accidentally pass in [None] if a prior schema entry was unexpectedly empty.
+        for kp in keypath:
+            if not type(kp) == str:
```
nmoroze, line 793: A couple nits:
- I think [`isinstance()`](https://docs.python.org/3/library/functions.html#isinstance) is a bit more idiomatic than `type(...) == ...`
- `key` might be a better index variable name than `kp`

siliconcompiler/core.py
```diff
@@ -787,6 +787,15 @@ def get(self, *keypath, field='value', job=None, cfg=None):
...
+                self.error(\
+f'''An invalid keypath was passed to "chip.get":
+{keypath}
+Your Chip configuration may be missing a parameter which is expected by your build script.''')
```
nmoroze, line 797: How about formatting like?
```python
self.error(
   'An invalid keypath was passed to chip.get: \n'
   f'{keypath} '
   ...
)
```
I think the full de-indent is a bit jarring stylistically.

I'd also be wary of multi-line error messages, since they kind of break the table formatting when logged. Maybe we could handle this by allowing self.error() to accept a list of messages and then log each one individually?

nmoroze: I also wonder if it's a better idea to make the error message hew more closely to the error and be less speculative? Something like a tweaked version of the exisisting error message in `chip.add()` seems useful enough to me:

```
"Key [{key}] in keypath  {keypath} is not a string" 
```

I imagine seeing the invalid keypath and the call location in the traceback is probably enough to infer, although it does make sense to strive to be as helpful as possible with error messages.

siliconcompiler/core.py
```diff
@@ -787,6 +787,15 @@ def get(self, *keypath, field='value', job=None, cfg=None):
 
         """
 
+        # Ensure that all keypath values are strings.
+        # Scripts may accidentally pass in [None] if a prior schema entry was unexpectedly empty.
+        for kp in keypath:
+            if not type(kp) == str:
+                self.error(\
```
nmoroze, line 794: Keep in mind that if `['option', 'continue']` is set this won't be fatal, so we'll probably bubble to another exception inside `_search()`. Might want to return early after with `None`?

nmoroze (commented):
This is a very helpful change, thanks! These errors have bugged me for a while, so it'll be nice to finally have this cleaned up.

I do think this PR could give us much broader benefit with just a bit more effort, and clean up some tech debt in the meantime.

First of all, I think we can get rid of all the `keypathstr` logic in get()/getkeys()/set()/add(), and just log the keypath directly. Note that this block:

```python
keypathstr = ','.join(keypath)
self.logger.debug(f"Reading from [{keypathstr}]. Field = '{field}'")
```

logs the same thing as

```python
self.logger.debug(f"Reading from {keypath}. Field = '{field}'")
```

Except it doesn't choke on `None`, and it makes the code simpler. We might as well fix all instances of this pattern.

The next issue is that calls to `_search()` require a keypathstr argument. However, this could be made an optional keyword argument that gets computed within `_search()` when it is first called, and then it can be passed along to each recursive call. 

Along with building the keypathstr in `_search()`, we can also lower the the type-checking logic itself to this function, and not have repeated code across `get()/getkeys()`. 

Finally, we could then get rid of the similar blocks that already exist in set()/add(), which do this check but aren't actually fatal (meaning they don't prevent the problem in the first place).

siliconcompiler/core.py
WRansohoff: Good idea, thanks.

siliconcompiler/core.py
WRansohoff: Good point, I re-formatted the error message a bit.

siliconcompiler/core.py
WRansohoff: Good point, I hadn't tested with `('option', 'continue')` set to `True`. Added an early `return None` for that case.

WRansohoff:
Good idea about using `f'{keypath}'` to avoid surfacing errors, and moving the key-checking into the shared `_search` method. I've made that change, and added a few set/get unit tests for these corner cases.

siliconcompiler/core.py
```diff
@@ -793,7 +793,7 @@ def get(self, *keypath, field='value', job=None, cfg=None):
             else:
                 cfg = self.cfg
 
-        keypathstr = ','.join(keypath)
+        keypathstr = f'{keypath}'
 
         self.logger.debug(f"Reading from [{keypathstr}]. Field = '{field}'")
```
nmoroze, line 798: I think we should get rid of the "[ ]" around `{keypathstr}` now, since casting the keypath to a string will include those (I think the keypath would be printed with two brackets now). Same applies to the debug logs in the other related functions.

siliconcompiler/core.py
```diff
@@ -953,7 +953,7 @@ def add(self, *args, cfg=None, field='value'):
             if not isinstance(key,str):
                 self.logger.error(f"Key [{key}] is not a string [{args}]")
```
nmoroze, line 954: I think we should delete the old checks now that we have it inside `_search()`. Same applies to set().

nmoroze (approved):
Great, thanks for addressing the comments and for adding the tests! Noticed a few last small things, but otherwise this looks good to go.

siliconcompiler/core.py
WRansohoff: Good idea, done.

siliconcompiler/core.py
WRansohoff: You're right - the f-string formatting uses `()` instead of `[]`, but that should work fine for logging statements.

aolofsson:
This is a great usability fix! I have resorted to putting print statements in the core code more than once to debug these None user errors, which clearly is not something that we want the end user to have to do.

PR #1086, opened 2022-07-16
Have KLayout run def2stream script directly

nmoroze:
This PR renames our KLayout def2stream.py script to klayout_export.py, and deletes the old klayout_export.py (which previously read in and `exec()`-ed the def2stream.py)

This change is based on feedback from @arlpetergadfordt that the old pattern resulted in more confusing error messages.

For demonstration, I inserted a syntax error in def2stream.py on line 221 and ran it before and after this change:
Before:
```
ERROR: invalid syntax (<string>, line 221)
  /home/noah/code/siliconcompiler/siliconcompiler/tools/klayout/klayout_export.py:11
```

After:
```
ERROR: /home/noah/code/siliconcompiler/siliconcompiler/tools/klayout/klayout_export.py:221: invalid syntax
```

I think the old approach was done to enable us to call two scripts sequentially. Since we're no longer doing that, I think this change makes sense.


aolofsson (approved):
Make sense!

PR #1088, opened 2022-07-16
Fix wheels build

nmoroze:
Looks like something must have changed on the MacOS Github environment that makes this directory already exist (causing the lack of `-p` flag to cause the `mkdir` command to fail).

This PR fixes the breakage. Manual run to verify: https://github.com/siliconcompiler/siliconcompiler/actions/runs/2680480664 (I think the error here is transient)


aolofsson (approved):
That's pretty weird, but I guess we can't question the stock install?

PR #1089, opened 2022-07-18
Gracefully handle invalid characters in tool output

### LINKED ISSUE #1073: SC crashes when logfiles contain non-ASCII characters
I noticed that when running Surelog over this [file](https://github.com/openhwgroup/cvfpu/blob/develop/src/fpnew_fma_multi.sv#L322) with non-ASCII characters, SC would crash due to Surelog writing them out to the logfile. 

Surelog has been updated to no longer emit these characters: https://github.com/chipsalliance/Surelog/pull/2877, but it still seems worth making SC more robust to this.

Crashes came in two places:
1) the actual executable run logic, if quiet=False
2) `check_logfile()`

nmoroze:
This PR addresses issue #1073, which reported SC crashing when displaying output or parsing the logs of tools that send invalid characters to stdout or stderr.

The details of the issue were slightly misstated: I originally thought non-ASCII characters were the problem, but it turns out SC can handle non-ASCII output from tools fine. The issue came up because an old version of Surelog had a bug that would cause it to output bytes that were not even valid UTF-8 (which is the default encoding for most Python installations).

Although this is an error in the tool, it should be handled gracefully and not cause SC to crash. Python's `open()` function lets you pass in an error handler to specify what it does when invalid characters are encountered. There are a handful of built-in handlers, but they all handle the issue silently. Since an invalid character indicates an issue with the tool, I implemented custom handlers that log a warning.

One of the handlers, used for displaying tool output, passes through to the builtin "replace" handler that replaces the invalid character with "�". The other, used for processing the logfile, simply ignores the invalid character. This seemed like a reasonable set of behaviors to me.

This might seem a little funky since it uses some less well-known Python features, but I think this is the best approach to handling this edge case. I tested it with the old version of Surelog that had the relevant bug, and it fixed the crash and output warnings in the correct places.

Closes #1073.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -35,6 +35,7 @@ import csv
 import distro
 import netifaces
 import webbrowser
+import codecs
 import packaging.version
 import packaging.specifiers
 from jinja2 import Environment, FileSystemLoader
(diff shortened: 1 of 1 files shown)


aolofsson:
Are you sure this warrants a custom handler? If we have only ever seen it with surelog, and there seems to be a fix already, is it worth the implementation and package dependency cost?

PR #1090, opened 2022-07-18
Some cleanup of check_logfile()

nmoroze:
This PR cleans up some things that gave me trouble when I was testing out PR #1089. I also did a pass on the docstring.

Code changes:
- If logfile is not supplied, construct full path based on jobname, step, and index. This makes this function more usable in a user's scripts, right now it is implicitly designed to run inside the task directory.
- Make default behavior for 'index' consistent with 'step' (use ['arg', 'index'] instead of first item in flowgraph)
- raise exception if no step or index provided (otherwise this bubbles into a fatal error, since these are needed to find regexes)
- Close the report files when done. Python garbage collector probably handles this, but it's good practice
- Fix inconsistent indentation

Docs changes:
- "logged" instead of "displayed to STDOUT" in docstring, since errors are printed to stderr (and logging module can have output redirected)
- Specify behavior of None values in docstring
- Fix example

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -2905,22 +2905,22 @@ class Chip:
 
         Reads the content of the task's log file and compares the content found
         with the task's 'regex' parameter. The matches are stored in the file
-        '<design>.<suffix>' in the current directory. The matches are printed to
-        STDOUT if display is set to True.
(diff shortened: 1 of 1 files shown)


nmoroze:
Thanks, should be all set now!

PR #1091, opened 2022-07-19
Implement basic support for pure-Python tools

nmoroze:
This PR implements support for pure-Python tools, a feature we talked about a while back when discussing how to more cleanly port other flows into SC. With both the ORFS and Caravel flows, we've noticed cases where we need to do some intermediate processing on files being passed around the flow. The existing options we explored each have flaws:

- Modify `pre_process()` or `post_process()` in a tool driver
  - This isn't great since we want to keep these functions as flow-agnostic as possible, so that we can reuse the tool drivers broadly. Forking the tool drivers per-flow isn't a great solution either, since this results in lots of code duplication.
- Break the run into multiple `run()` calls that each use a partial steplist, with Python logic in between
  - This is what the Caravel example currently does, but it breaks some of SC's nice features for debugging (steplist is overwritten, flowgraph viz doesn't capture intermediate processing, the intermediate processing results aren't captured in the build directory)

Natively supporting pure Python steps seems like a cleaner solution to this problem. As a basic approach, I changed SC to treat any tool driver that defines a `run(chip)` function as a pure-Python tool. The SC runtime will then call this function instead of an executable. 

The one problem with this implementation is that it introduces a third code path that doesn't include support for some nice features, like output redirection, memory tracking, and task timeouts. If we run into a situation where we want these features, I think we could explore running these tools in a separate process that can then be monitored by the runtime in order to provide this functionality, but this was the simplest first approach.

As a proof-of-concept, I updated the Caravel example to use these features. To validate the changes, I ran the flow before and after and made sure the input DEF to KLayout and that the final DEF passed into KLayout are the same. If you'd like more certainty that the changes don't break anything, I can try re-running the MPW prechecks using the new flow.

### DIFF
--- a/examples/heartbeat_caravel_wrapper/flows/mpwflow.py
+++ b/examples/heartbeat_caravel_wrapper/flows/mpwflow.py
@@ -1,5 +1,15 @@
 import siliconcompiler
 
+def make_docs():
+    '''An RTL2GDS flow for the eFabless MPW shuttle. This flow is based off the
+    basic serial asicflow, with the addition of fixdef and addvias steps for
+    doing MPW-specific fixes to the post-routing DEF and gate-level netlist,
+    respectively.'''
+    chip = siliconcompiler.Chip('<design>')
+    setup(chip)
+    chip.set('option', 'flow', 'mpwflow')
+    return chip
+
 def setup(chip):
     flow = 'mpwflow'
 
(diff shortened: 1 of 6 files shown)


examples/heartbeat_caravel_wrapper/flows/mpwflow.py
```diff
@@ -0,0 +1,42 @@
...
+    chip.node(flow, step, tool)
+
+    prevstep = step
+    for step, tool in flowpipe[1:]:
+        chip.node(flow, step, tool)
+        chip.edge(flow, prevstep, step)
+        prevstep = step
+
+    chip.node(flow, 'addvias', 'addvias')
+    chip.edge(flow, 'dfm', 'addvias')
```
WRansohoff, line 38: I like how this demonstrates the potential for parallel flowgraph steps; the netlist post-processing can happen in parallel with the GDS-streaming step, which this captures.

WRansohoff (approved):
This looks great! I really like the approach, and it demonstrably simplifies the Caravel flow's custom Python logic.

aolofsson (approved):

PR #1093, opened 2022-07-20
Filter out manifests when checking flowgraph I/O

### LINKED ISSUE #1077: Overconservative check when receiving input from multiple tasks steplist
If two tasks `A` and `B` feed into task `C`, and then task `C`is re-run with a steplist, the checker will fail on an error:

```
Invalid flow: or_final_report0 receives <design>.pkg.json from multiple input tasks
```

The checker should be loosened to exempt manifests, since those aren't copied forward from task outputs.

nmoroze:
This PR fixes a `check_manifest()` bug when checking a run that uses a steplist to re-run a task that joins 2+ other tasks. The logic that checks what inputs the task will receive needs to filter out the manifest (since this isn't copied forward), otherwise the checker complains that the task will receive a manifest from both.

Also a couple other small changes I noticed:
- Helpful error message when you forget ['option', 'flow']
- We don't need to quit early with a steplist length 1, it's still useful to check that a single node receives the required inputs

Closes #1077

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -4178,6 +4178,9 @@ class Chip:
         '''
 
         flow = self.get('option', 'flow')
+        if flow is None:
+            self.error("['option', 'flow'] must be set before calling run()",
+                       fatal=True)
 
         # Re-init logger to include run info after setting up flowgraph.
         self._init_logger(in_run=True)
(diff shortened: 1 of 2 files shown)


aolofsson (approved):
Nice catch and fix!

PR #1096, opened 2022-07-22
Simplify collection of input files

### LINKED ISSUE #1063: Source code debugging
The current SC approach copies files into the import directory before starting work, making it tough to debug source code for linting/dv since the reference is not the original code, but some file inside the build structure. The approach we took so far was done to enable remote processing and clean encapsulation of the design for rtl2gds, but it makes correlating the errors back to the original source file cumbersome.

For any action (linting, formal, dv, codecov) this is going to be a pretty big issue that we need to address.

nmoroze:
This PR is a simple proposal to fix issue #1063. 

It makes two changes:
- Only collect files for remote runs
- Don't copy collected files from `import/0/inputs/` to `import/0/outputs/`

This was a pretty easy change, since `find_files()` automatically falls back to searching the original file path for any copied file that's not in the file collection directory.

Here are the before and after results.

**Before**
```console
$ ls build/gcd/job0/import/0/inputs/
gcd_0a998a746037a840df8d7f2133c57f4846e3405f.v  gcd_9fb1319e7d576ade53f9b17ec8e05417afa27bca.sdc

$ ls build/gcd/job0/import/0/outputs/
gcd_0a998a746037a840df8d7f2133c57f4846e3405f.v    gcd.pkg.json
gcd_9fb1319e7d576ade53f9b17ec8e05417afa27bca.sdc  gcd.v

$ cat build/gcd/job0/import/0/replay.sh
#!/bin/bash
surelog -parse /home/noah/code/siliconcompiler/build/gcd/job0/import/0/outputs/gcd_0a998a746037a840df8d7f2133c57f4846e3405f.v -top gcd +libext+.sv+.v
```

**After**
```console
$ ls build/gcd/job0/import/0/inputs/

$ ls build/gcd/job0/import/0/outputs/
gcd.pkg.json  gcd.v

$ cat build/gcd/job0/import/0/replay.sh
#!/bin/bash
surelog -parse /home/noah/code/siliconcompiler/examples/gcd/gcd.v -top gcd +libext+.sv+.v
``` 

This makes the results from a remote run a little different, but that seems acceptable to me. It also helps that the `import/0/outputs` directory is cleaner:
```console
$ ls build/gcd/job0/import/0/inputs/                          
gcd_0a998a746037a840df8d7f2133c57f4846e3405f.v  gcd_9fb1319e7d576ade53f9b17ec8e05417afa27bca.sdc

$ ls build/gcd/job0/import/0/outputs/ 
gcd.pkg.json  gcd.v

$ cat build/gcd/job0/import/0/replay.sh  
#!/bin/bash
surelog -parse /home/noah/code/siliconcompiler/build/gcd/job0/import/0/inputs/gcd_0a998a746037a840df8d7f2133c57f4846e3405f.v -top gcd +libext+.sv+.v
```

In a future change, it may be good have to `archive()` call `_collect()` now, if we care about capturing inputs in the archive.

Closes  #1063.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1392,7 +1392,7 @@ class Chip:
                 name = self._get_imported_filename(path)
                 abspath = os.path.join(self._getworkdir(jobname=job, step='import'), 'inputs', name)
                 if os.path.isfile(abspath):
-                    # if copy is True and file is found in import outputs,
+                    # if copy is True and file is found in import inputs,
                     # continue. Otherwise, fall through to _find_sc_file (the
                     # file may not have been gathered in imports yet)
                     result.append(abspath)
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -2529,7 +2529,6 @@ class Chip:
         '''
 
         indir = 'inputs'
-        flow = self.get('option', 'flow')
 
         if not os.path.exists(indir):
             os.makedirs(indir)
@@ -2544,10 +2543,6 @@ class Chip:
                 shutil.copy(abspath, os.path.join(indir, filename))
             else:
                 self._haltstep(step, index)
-
-        outdir = 'outputs'
-        if not os.path.exists(outdir):
-            os.makedirs(outdir)
 
     ###########################################################################
     def archive(self, step=None, index=None, all_files=False):
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -2500,6 +2500,8 @@ class Chip:
         copyall = self.get('option', 'copyall')
         allkeys = self.getkeys()
         for key in allkeys:
+            if key[0] == 'history':
+                continue
             leaftype = self.get(*key, field='type')
             if re.search('file', leaftype):
                 copy = self.get(*key, field='copy')
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1390,7 +1390,7 @@ class Chip:
         for path in paths:
             if (copyall or copy) and ('file' in paramtype):
                 name = self._get_imported_filename(path)
-                abspath = os.path.join(self._getworkdir(jobname=job, step='import'), 'outputs', name)
+                abspath = os.path.join(self._getworkdir(jobname=job, step='import'), 'inputs', name)
                 if os.path.isfile(abspath):
                     # if copy is True and file is found in import outputs,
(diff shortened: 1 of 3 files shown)


aolofsson (approved):

tests/core/test_check_manifest.py
```diff
@@ -35,9 +35,12 @@ def test_check_allowed_filepaths_pass(scroot, monkeypatch):
...
 
-    # run an import just to collect files
-    chip.set('option', 'steplist', 'import')
-    chip.run()
+    # collect input files
+    workdir = chip._getworkdir(step='import', index='0')
+    os.makedirs(workdir)
+    os.chdir(workdir)
+    chip._collect('import', '0')
+    os.chdir(chip.cwd)
```
WRansohoff, line 43: Should we store the previous dir instead of using `chip.cwd` to be safe, since `os.chdir` is a global interpreter setting? Same for the following test.

WRansohoff (approved):
Looks good to me.

I was a bit worried about how remote scheduled jobs could be impacted by the _collect method changes, but I think it will be fine since _collect is only called on the import stage. The heartbeat example ran successfully while using this branch in our test cluster, too.

siliconcompiler/core.py
```diff
@@ -1390,7 +1390,7 @@ def find_files(self, *keypath, cfg=None, missing_ok=False, job=None):
         for path in paths:
             if (copyall or copy) and ('file' in paramtype):
                 name = self._get_imported_filename(path)
-                abspath = os.path.join(self._getworkdir(jobname=job, step='import'), 'outputs', name)
+                abspath = os.path.join(self._getworkdir(jobname=job, step='import'), 'inputs', name)
                 if os.path.isfile(abspath):
                     # if copy is True and file is found in import outputs,
```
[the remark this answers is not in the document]
nmoroze: Good catch, thanks!

tests/core/test_check_manifest.py
nmoroze: Yeah that seems safer and more explicit, fixed. I'm curious now if `pytest` has some mechanism to isolate a changed working directory between tests - I wanted to see if they have an official way to do this, but looks like their monkeypatch library just uses os.chdir as well: https://docs.pytest.org/en/7.1.x/_modules/_pytest/monkeypatch.html#MonkeyPatch.chdir. 

Either way, I think this is a good change!

nmoroze:
Great, thanks for the review and thanks for trying this out on the test cluster!

PR #1097, opened 2022-07-22
Don't overwrite ['input', 'verilog'] on remote

nmoroze:
I noticed these lines while trying to debug a test failure in my last PR. They weren't relevant, but they gave me pause - I don't think they're needed anymore. The inputs will be automatically resolved to the right directory by `find_files()` anyways.

Seem ok to delete?


aolofsson (approved):

WRansohoff (approved):
Looks good, thanks for keeping the reference server implementation up-to-date!

PR #1104, opened 2022-07-27
Fix pin sizes and PDN vias in Caravel wrapper example

WRansohoff:
This PR fixes two issues with our Caravel wrapper example:

* While merging the heartbeat/Caravel example into main, I deleted a line in `sc_floorplan.tcl` which set custom pin sizes. The default sizes ended up triggering some metal grid offset DRCs in the MPW precheck scripts, so I added a way to set I/O signal sizes from the schema and added a fix to the example build script.

* When I was inspecting the final GDS results, I noticed that the top-level build was only adding vias between the PDN grid and standard cell rails outside of the core macro area. This did not cause DRC or LVS failures because the power signals were technically connected to the correct nets, but I would imagine that the voltage drop caused by the missing power vias would cause problems in a real chip. Fortunately, this can be fixed with a small adjustment to the PDN script.

### DIFF
--- a/examples/heartbeat_caravel_wrapper/build.py
+++ b/examples/heartbeat_caravel_wrapper/build.py
@@ -52,6 +52,8 @@ def build_core():
     core_chip = configure_chip('heartbeat')
     design = core_chip.get('design')
     core_chip.set('input', 'verilog', 'heartbeat.v')
+    core_chip.set('tool', 'openroad', 'var', 'floorplan', '0', 'pin_thickness_h', ['2'])
+    core_chip.set('tool', 'openroad', 'var', 'floorplan', '0', 'pin_thickness_v', ['2'])
     core_chip.set('tool', 'openroad', 'var', 'place', '0', 'place_density', ['0.15'])
     core_chip.set('tool', 'openroad', 'var', 'route', '0', 'grt_allow_congestion', ['true'])
     core_chip.clock('clk', period=20)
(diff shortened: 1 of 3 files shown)


aolofsson:
We should probably discuss this some more. Driving variables from Python directly into the TCL file with the var should be a last resort. As much as possible, we should try to capture these details in standardized variables that can be used across all tools.

WRansohoff:
I tried to mimic our existing solution for configurations such as placement density. I could add [a similar value in `sc_apr.tcl`](https://github.com/siliconcompiler/siliconcompiler/blob/main/siliconcompiler/tools/openroad/sc_apr.tcl#L100) if you think that would be cleaner?

WRansohoff:
Added one more small fix: in testing, we discovered that one of the updates to `klayout_export.py` caused the mpw_precheck scripts to take ~24h to complete, instead of the ~3h which we were used to.

We made the change to retain parity with the OpenROAD-flow-scripts `def2stream.py` script, [which we derived the export script from](https://github.com/siliconcompiler/siliconcompiler/blob/34c1889cc122605edc95960eff7f9060c09eabf1/siliconcompiler/tools/klayout/klayout_export.py#L3). They have made further changes in the meantime, so I applied [a more recent update](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/79dc475ee7d24b7b7e95e6d2f377565a1cba4844/flow/util/def2stream.py#L164) with a similar intent, which does not negatively impact our MPW precheck runtimes.

aolofsson (approved):

Issue #1103, opened 2022-07-27
Use .odb rather than .def between OpenROAD steps


maliberty:
There is information that can't be stored in DEF so OpenROAD is moving away from it as a persistence format.  For example congestion, routing guides, pin access, Verilog hierarchy, and power domains have no representation in DEF.  RTL-MP and UPF support are near term features that will require odb.

You should use read_db / write_db rather than read_def / write_def.  ORFS has already made the switch and can be used as a reference.

maliberty:
Note that write_def remains available for interfacing with other tools (eg KLayout) at the end of the OR flow.

WRansohoff:
Thank you for the suggestion and context - this explains the removal of the `-guide` parameter in newer versions of OpenROADs' detailed routing steps.

aolofsson:
@maliberty In principle, what features are deprecated in the flow that would preclude the use of a def based flow? If the -guide input is gone from the detailed router, does that mean there is no way to take in results from an external placer/global router? Since openroad is currently the only APR that supports odb, this seems like a step backwards?

I think ASCII readable hand off should always be an available fall back option (even if only applicable at older nodes).

QuantamHD:
@aolofsson The DEF writer is still fully featured, but it lacks the ability to store tool specific information.

It's still fully possible to read_def and route that with openroad. What we're advocating for is that when you're using openroad the odb should take precedent over DEF.

DEF has a number of missing features that make it non ideal for flow infrastructure, namely it has no hierarchy support.

QuantamHD:
I do believe that in the long term we should try to move all open source tools to odb. DEF will always have a place, but odb will allow tighter integration with the rest of the open source tools.

WRansohoff:
Thanks for the extra information. Is it safe to read and write both OpenDB and DEF files to gather all available information from both formats? Or do we risk getting duplicate data if we call `read_db` followed by `read_def`?

QuantamHD:
In general those commands are pretty much mutually exclusive. Write commands are idempotent, and def is always a subset of odb so there would never be a benefit to do so for a *single block*.

read_def has a few modes on it that allow you to read pins or floorplan information in addition to the db..

 OpenROAD assumes a single unique instance of a chip/block in its code. Which read_def and read_db  either instantiate or overwrite. This is something @maliberty and I are working on rectifying.

In short read_db and read_def tend to overwrite everything.

aolofsson:
[...]

PR #1106, opened 2022-07-27
Update OpenROAD version

WRansohoff:
Wenting recently noticed that our build scripts are incompatible with new versions of OpenROAD, because of a change to the `detail_route` API.

On the same note, @maliberty mentioned in #1103 that OpenROAD is starting to store interstitial data in OpenDB files rather than DEFs.

So it seems like a good time to update the version of OpenROAD that we support. I added `write_db` and `read_db` calls to our tool drivers, but I did not remove or modify our existing `read_def` calls. I am assuming that re-reading data which is captured by both formats won't cause issues.

I will need to update the version of OpenROAD which is installed on our test runner to make the CI tests pass, but doing that will also cause test failures on new PRs until this change is merged, so I'm holding off until we have a chance to discuss it.

### DIFF
--- a/siliconcompiler/tools/openroad/sc_apr.tcl
+++ b/siliconcompiler/tools/openroad/sc_apr.tcl
@@ -156,12 +156,13 @@ if {[dict exists $sc_cfg "input" def]} {
             read_def $def
         }
     }
-} elseif {[file exists "inputs/$sc_design.odb"]} {
-    # Prioritize ODB files over DEF, if available.
-    read_db "inputs/$sc_design.odb"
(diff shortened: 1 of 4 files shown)


WRansohoff:
Updated this based on feedback that we should not read a DEF input after reading an OpenDB input.

We still need to update the version of OpenROAD installed on the test runner for the CI tests to pass, but that will cause the CI tests to fail on other PRs until this change is merged.

aolofsson:
Good to go?

WRansohoff:
Yep!

aolofsson (approved):

PR #1107, opened 2022-07-28
HTML report: add basic editing functionality for signoff checklists.

WRansohoff:
This change builds on the tree-view manifest visualization which @nmoroze recently added to our auto-generated HTML reports. It adds a way for users to edit signoff checklist items, including checkboxes for the boolean "ok" parameters, and free-form text entry for the "report" parameters.

The "report" checklist parameters are currently intended as a list of files which are associated with a given signoff item, but it seems like we should have a way for users to enter arbitrary metadata for each item. I can think of two easy ways to address that:

* Change the `report` parameter type from `[file]` to `[str]`, or
* Add a new checklist parameter for free-text input.

### DIFF
--- a/siliconcompiler/templates/report/sc_report.j2
+++ b/siliconcompiler/templates/report/sc_report.j2
@@ -121,6 +121,7 @@
                 ok_li.innerHTML = '<b>Item Accepted:</b> <input id="' + ok_id + '" type="checkbox" ' + checked + '/>';
                 new_child_list.appendChild(ok_li);
               }
+              /* TODO: Optional free-text entry, but the 'report' parameter is intended for file paths.
               else if (el == 'report') {
                 // Provide a text input field for arbitrary checklist item metadata.
                 rpt_li = document.createElement("li");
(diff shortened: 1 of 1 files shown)


nmoroze:
I think we should add a new parameter. `check_checklist()` adds paths to tool reports in the `report` parameter, and I think it would be ideal to keep all the file-handling niceties we get from the stricter type.

Also, was there any particular reason to switch over to using Javascript to generate the tree viewer instead of generating it as an HTML template (my best guess is to regenerate the tree view after it's modified, but not sure)?

WRansohoff:
Good call. And you're right, I switched the tree-generating logic from Jinja to JS so that the tree view could be re-generated when the user uploads a previously-edited manifest.

WRansohoff:
After some offline discussion, it sounds like the free-text entry isn't too important. So I commented out the 'report' input for this initial PR, to avoid cluttering the schema before we have a chance to discuss a more holistic approach to signoffs.

aolofsson (approved):

PR #1108, opened 2022-07-28
Fix ['option', 'jobincr']

nmoroze:
This PR fixes up some issues with ['option', 'jobincr'] and adds a test for it.

Changes:
- No more blanket `try... except` wrapper. This was hiding the fact that this block was writing to a keypath that no longer exists :). Inserted more specific error checking logic instead.
- Run before `_init_logger()`, otherwise pre-runtask logs use old jobname in messages
- Strip off any digits from existing jobname to handle case where the existing name includes digits already (`job0`) or not (`job`)

The behavior now matches the existing schema help string, including appending "1" if the existing build directory does not end in any digits.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -4207,6 +4207,22 @@ class Chip:
             self.error("['option', 'flow'] must be set before calling run()",
                        fatal=True)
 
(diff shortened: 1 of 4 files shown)


aolofsson (approved):
Nice fix! I think that was the last extraneous try/except in core.py.

PR #1109, opened 2022-07-28
Merge in manifests from all leaf tasks

### LINKED ISSUE #1101: We should merge in manifests from all "terminal steps"
We should implement this TODO: https://github.com/siliconcompiler/siliconcompiler/blob/main/siliconcompiler/core.py#L4489-L4490

Metrics from steps that don't merge into the very last step of the steplist don't get picked up in the final manifest/printed properly in summary(). 

Enforcing that all steps merge into a final "export" step doesn't solve this, since we may run a steplist that doesn't go all the way until export. I think all metrics should be displayed in summary() even in this scenario.

nmoroze:
This PR implements an old TODO to merge in manifests from all "leaf" tasks at the end of run() (i.e. any task that doesn't have any further dependencies).

This is important to:
a) make the summary nice
b) capture all final metrics in one manifest

Even though we have had an informal requirement that all flows should merge into a final 'export' task, this has only been loosely followed, and it also doesn't help if someone runs a steplist that doesn't include the final export task. 

We already have the relevant helper function implemented from a previous fix to summary(), so this was a pretty easy change.

Demo of the difference this makes -

Example flow:
flow

Summary before:
```
                   import0        syn0       syn1     floorplan0
 errors               0            0         ERR           0    
 warnings             10           74        ERR           1    
 drvs                 0            0         ERR           0    
 unconstrained        0            0         ERR           0    
 coverage            0.0          0.0        ERR          0.0   
 security            0.0          0.0        ERR          0.0   
...
```

Summary after:
```
                   import0        syn0         syn1     floorplan0
 errors               0            0            0            0    
 warnings             10           74           74           1    
 drvs                 0            0            0            0    
 unconstrained        0            0            0            0    
 coverage            0.0          0.0          0.0          0.0   
 security            0.0          0.0          0.0          0.0   
...
```

Closes  #1101

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -4472,7 +4472,7 @@ class Chip:
                 if not index_succeeded:
                     self.error('Run() failed, see previous errors.', fatal=True)
 
-            # On success, write out status dict to flowgraph status'. We do this
+            # On success, write out status dict to flowgraph status. We do this
             # since certain scenarios won't be caught by reading in manifests (a
             # failing step doesn't dump a manifest). For example, if the
             # steplist's final step has two indices and one fails.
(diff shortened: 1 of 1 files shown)


aolofsson (approved):
Nice!

PR #1120, opened 2022-08-04
Making libcorner a list

aolofsson:
- Address the scenario of having multiple different corners applied to a single scecnario, which is needed in cases like multi-voltage IO. For example, we might have multiple IO models, one for 0.9V internal and 1.8V external, or 0.9V internal and 1.72V external. With a large number of IO standardards and on chip voltages, we need a way to group these library corners together within a single scenario.

### DIFF
--- a/siliconcompiler/tools/openroad/sc_apr.tcl
+++ b/siliconcompiler/tools/openroad/sc_apr.tcl
@@ -121,12 +121,12 @@ read_lef  $sc_techlef
 
 # Read Liberty
 foreach item $sc_scenarios {
-    set corner [dict get $sc_cfg constraint $item libcorner]
-    foreach lib "$sc_targetlibs $sc_macrolibs" {
-	#Liberty
-	if {[dict exists $sc_cfg library $lib model timing $sc_delaymodel]} {
-	    set lib_file [dict get $sc_cfg library $lib model timing $sc_delaymodel $corner]
-	    read_liberty $lib_file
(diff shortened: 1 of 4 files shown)


siliconcompiler/tools/openroad/sc_apr.tcl
```diff
@@ -121,12 +121,12 @@ read_lef  $sc_techlef
...
-    foreach lib "$sc_targetlibs $sc_macrolibs" {
-	#Liberty
-	if {[dict exists $sc_cfg library $lib model timing $sc_delaymodel]} {
-	    set lib_file [dict get $sc_cfg library $lib model timing $sc_delaymodel $corner]
-	    read_liberty $lib_file
+    foreach corner [dict get $sc_cfg constraint $item libcorner] {
+	foreach lib "$sc_targetlibs $sc_macrolibs" {
+	    if {[dict exists $sc_cfg library $lib model timing $sc_delaymodel $corner]} {
+		set lib_file [dict get $sc_cfg library $lib model timing $sc_delaymodel $corner]
+		read_liberty $lib_file
```
nmoroze, line 128: Is no indent under the if statement intentional to keep line width short? Generally not a fan of this, but I'm not as picky about TCL style as I am Python :)

nmoroze (approved):
Looks good to me, nice change! Thanks for cleaning up the random debug messages in the TCL too.

PR #1123, opened 2022-08-07
add --latches to the list of valid extraopts for ghdl

suppamax:


nmoroze (approved):
Looks good to me!

I'm wondering now if my suggestion to hard-code the list of valid options was a bit short-sighted. I could see this getting onerous since I think GHDL has a lot of switches one might want to use. Perhaps in a future PR it's worth getting rid of this extra layer of error checking entirely, and letting the user interpret the tool error if they get something wrong.

aolofsson (approved):

PR #1128, opened 2022-08-25
Fix `find_files()` job argument + more

### LINKED ISSUE #1122: `find_files()` job parameter doesn't work

nmoroze:
This PR contains a slew of changes that I think should be in the next release. It got a bit more varied than I originally intended, let me know if you want me to break it up.

Summary:
- Fix `find_files()` behavior with job argument (fixes #1122)
  - Was passing along `job` args to each `set()`/`get()` call internally, but was also passing in a `cfg` arg that was hardcoded to `self.cfg`, and that always takes precedence over the job.
  -  I have a longer-term fix in the works that will refactor basic schema access into a separate class (which we discussed a while back), and that will let us remove the `cfg` arguments and make bugs like this less likely to occur in the future
- Adds reports/final manifest to `chip.archive()`
- Export custom Sphinx `:keypath:` directive 
  - Could be useful for documenting dynamic modules in a third-party collection
- Tweak how libraries are imported when calling `read_manifest()` with job arg
  - Puts them under `history, <jobname>, library, <libname>` instead of top-level libraries, which is more consistent with `record_history()`
  - Might want to move away from this going forward to reduce manifest size, but then would need a way to make sure that different jobs actually used the same library config for proper provenance
- Small misc. tweaks

### DIFF
--- a/docs/_ext/schemagen.py
+++ b/docs/_ext/schemagen.py
@@ -92,20 +92,9 @@ class CategorySummary(SphinxDirective):
 
         return new_doc
 
-def keypath_role(name, rawtext, text, lineno, inliner, options={}, content=[]):
-    # Split and clean up keypath
-    keys = [key.strip() for key in text.split(',')]
-    try:
-        return [keypath(*keys)], []
-    except ValueError as e:
-        msg = inliner.reporter.error(f'{rawtext}: {e}', line=lineno)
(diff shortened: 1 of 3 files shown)


aolofsson:
Looks good!

aolofsson (approved):

PR #1129, opened 2022-08-25
Bump Surelog version

### LINKED ISSUE #1117: Surelog fails to parse parameter with partial vector select
There seems to be an issue with the surelog parser for partial parameter selections. It's a problem because we rely on surelog to help us bring in files to the next step. 

https://github.com/chipsalliance/Surelog/issues/3133

I have veriried that the design goes through yosys with verilator as the front end parser.

nmoroze:
Fixes #1117

This one was kind of a struggle to get working. For some reason the bundled Surelog only works w/ Linux if we add the `-nocache` option. I'll need to revisit that when I get a chance. I also think it might be worth revisiting our approach to bundling Surelog, since upgrading is a pretty high burden.

### DIFF
--- a/.github/workflows/bin/install_surelog_macos.sh
+++ b/.github/workflows/bin/install_surelog_macos.sh
@@ -6,5 +6,6 @@ pip3 install orderedmultidict
 # Install Surelog
 git submodule update --init --recursive third_party/tools/surelog
 cd third_party/tools/surelog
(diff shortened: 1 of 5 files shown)


aolofsson:
Agreed! If the barrier is too high it means we can't afford to upgrade very often.

aolofsson (approved):

PR #1133, opened 2022-08-31
Add more file entries to tool schema

nmoroze:
This PR adds back dedicated pre/post scripts, as well as a free file entry.

I was thinking another way to do this could be to add a free key after 'script', and standardize on values "entry", "pre", "post", and then allow free keys otherwise. This is a bit painful for backwards compatibility though.


aolofsson:
Nice change! pre/post scripts are fairly standardized and adding a free variable for file is a good enough catch all!

aolofsson (approved):

Issue #1087, opened 2022-07-16
Proposal: flexible task I/O driven by flow


nmoroze:
The way most of our tool drivers are currently written, each tool sets certain expected input/output filenames in the `['tool', <tool>, 'input'/'output', ...]` parameters. The values are hard-coded on a per-step basis, effectively defining the "step" as a function with a certain signature.

However, we've found this isn't quite flexible enough for our purposes. For example, we wanted to reuse the OpenROAD and Yosys tool drivers for our ORFS Makefile replacement project, but that project uses a different set of steps and different I/O structure than our scripts.

In addition, our current approach has resulted in some persistent messiness. A lot of tools have checks like:

```tcl
    if {[dict exists $sc_cfg "input" netlist]} {
        foreach netlist [dict get $sc_cfg "input" netlist] {
            read_verilog $netlist
        }
    } else {
        read_verilog "inputs/$sc_design.vg"
    }
```

and we've had to do things like add prefixes to input names to prevent tools from trying to step on each other when we used in the same flow (e.g. `floorplan.def` vs `def`). This feels like something that won't scale well, especially if we want to start standardizing input names.

Based on some of our discussion around this, I think there's a consensus that we should specify expected tool input/output files in the flow itself. We should add additional parameters in the flowgraph schema to support this. Exact structure TBD, but I think each task should include the following information:

- What input filetypes does the task take in
- What output filetypes does the task output
    - For now, think of "filetypes" as free keys equivalent to the free keys that may be supplied via `['input', ...]`
- For each input filetype:
    - Is this filetype supplied from the `['input', ...]` schema, or supplied from a previous flow step?
    - If supplied from a previous step, what filename do we expect? [bonus: should we support globs for accepting arbitrary # of inputs?]
    - If supplied from the schema, how many of this filetype are we expecting? [bonus: for error checking]
- For each output filetype:
    - What are the names of the output files?
    - [note: do we want to support "dealer's choice" to feed into glob tools?]

On the tool side, the `['tool', <tool>, 'input'/'output', <step>, <index>]` keypath should be tweaked to add a 'filetype' key. Then, tool drivers can be refactored to extract filenames from these keypaths, similarly to how Will refactored the OR driver for post-processing:

```python
    out_def = ''
    out_files = []
    if chip.valid('tool', tool, 'output', step, index):
        out_files = chip.get('tool', tool, 'output', step, index)

    if out_files:
        for fn in out_files:
            if fn.endswith('.def'):
[...]

PR #1143, opened 2022-09-13
Restrict which files get copied between tasks

nmoroze:
This PR adds a couple schema parameters and basic functionality to enable:
- Restricting which files get copied between tasks
- Distinguishing types of input files to a task

For an example use case: imagine we wanted to run gate-to-gate LEC between `cts` and `place`. We could create a `lec` task that depends on both `place` and `cts`, but then we run into an issue: 

```
$ ls place/0/outputs/
design.def   design.place.vg

$ ls cts/0/outputs/
design.def design.cts.vg
```

Since these files will get copied directly by the runtime into `lec/0/inputs`, the checker will throw an error due to the duplicated `design.def` inputs. In addition, the LEC step has no way of distinguishing the two netlists to be compared.

This PR solves the problem by allowing a user to filter out a specific set of files to be copied between one step and another, which is exposed as an optional kwarg to the `edge()` API method. For this example, the flow definition might be tweaked to look like:

```python
chip.node(flow, 'lec', 'yosys')
chip.edge(flow, 'place', 'lec', files=[
  ('design.place.vg', 'netlist-ref')
])
chip.edge(flow, 'cts', 'lec', files=[
  ('design.cts.vg', 'netlist')
])
```

Then, during runtime two things will be different:
- Only `design.place.vg` and `design.cts.vg` will be copied into `lec/0/inputs`, solving the first problem.
- The new parameter `['tool', <tool>, 'infile', <step>, <index>, <filetype>]` will be set for this task to point to these inputs (with distinct filetypes of `netlist-ref` and `netlist`), solving the second problem.

We've previously discussed moving towards defining info about flow inputs/outputs directly in the flow definition ( #1087), and this PR enables the immediate use case in a backwards-compatible way while moving towards this broader change. To make this change forward-looking, the schema parameters proposed are a bit more complex than needed for just this functionality.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1966,6 +1966,7 @@ class Chip:
                 in_tasks = self.get('flowgraph', flow, step, index, 'input')
                 all_inputs = set()
                 for in_step, in_index in in_tasks:
+                    copied_inputs = self._inputs_expected_from_task(step, index, in_step, in_index)
                     if in_step not in steplist:
                         # If we're not running the input step, the required
                         # inputs need to already be copied into the build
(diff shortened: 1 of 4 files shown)


nmoroze:
Postponing this change for later!

PR #1145, opened 2022-09-15
Read in manifests from previously completed tasks before running

nmoroze:
This PR fixes a bug where tool setup gets lost when a partially run flow is continued using the steplist. This means things like the tool report mappings are no longer fully captured when a flow is executed over multiple runs. 

That immediate problem is caused by the fact that the tool setup functions are executed at runtime, and only for the steps currently in the steplist for that run. Therefore, setup for tools from the steps that have already been run don't end up in the chip configuration.

One way to fix this could be to re-run setup() for all steps every time, but I think it would make more sense to capture the setup that was used to run those particular steps, in case the user changes something between re-runs. 

One cause of this problem was introducing the partial read_manifest between steps, so I was also considering adding some of the keypaths we want to carry forward as things that get copied by a partial read. However, reading in the entire manifest seems a little more comprehensive, even if a bit ugly.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1832,13 +1832,10 @@ class Chip:
 
         for step in steplist:
             for index in indexlist[step]:
-                in_job = None
-                if (step in self.getkeys('option', 'jobinput') and
-                    index in self.getkeys('option', 'jobinput', step)):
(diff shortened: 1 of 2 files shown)


aolofsson (approved):
A good fix now!

Issue #1146, opened 2022-09-19
Make clobber interact appropriately with default values


nmoroze:
I think the `clobber` flag to `chip.set()` currently works in an unintuitive way. The problem is it will always overwrite an "empty" value, even if it has been explicitly set. E.g.:

```
>>> import siliconcompiler
>>> chip = siliconcompiler.Chip('test')
>>> chip.set('option', 'remote', False)
>>> chip.get('option', 'remote')
False
>>> chip.set('option', 'remote', True, clobber=False)
>>> chip.get('option', 'remote')
True
```

I think it should not be considered correct behavior for the second `chip.set()` to override the value that the user has explicitly set. This behavior caused the server tests to fail in the first iteration of #1145, and I think a proper solution would be to fix this issue.

I think the fix for this would be to no longer set the 'value' field of parameters when the schema is first initialized, and rather just have chip.get() fall back to the defvalue transparently when no value is set. Then, sets with clobber=False will only take effect if the value field does not exist.

I'd like to refactor `_search()` at some point soon, after I finish splitting out schema access into a separate class. Once that happens, I can make this change and fix this issue.

PR #1147, opened 2022-09-19
Refactor schema access into separate class

nmoroze:
This PR implements a big refactor that's been in our backlog for a while - adding a `Schema()` class that encapsulates basic manipulation of the config dictionary.

## Rationale
The primary value I see in this change is to give us an alternative for the `cfg` keyword arguments that many of the schema manipulation functions had. These were being used to manipulate config dictionaries other than the current Chip's (for example, when merging in a dictionary from a manifest). However, these keyword arguments go against our object-oriented model, and handling them incorrectly can result in subtle bugs like  #1122. 

This refactor also allows for low-level config manipulation without accessing the dictionary directly. We do this a lot in the docs generation and the server-side code, but this is a bit clunky, and abstracting with an API will enable us to make changes to the underlying dictionary representation (which will let us fix other issues down the road, such as #1146).

Another benefit is this helps organize our codebase a bit and keep down the size of our bloated core.py. I also see this as a chance to push for improved coding style, and I'm hoping to revisit an incremental approach to linting in CI starting with this new file.

Despite citing this as a key thing in my original proposal, I don't see design name bootstrapping as a key reason for this change (although this will help us clean some hacks in the doc generator). Rather, I think my proposal in #1169 will cover that well.

## API changes

One goal of this change was to make it a pure refactor with minimal-to-no API changes. The one exception is the removal of the 'cfg` keyword argument from all functions that used it. However, I did take a few other opportunities to clean up the API, here are the changes:

- `valid()`
  - Get rid of `quiet` keyword argument
  - `Schema()` doesn't have access to logger, making this tricky to implement, but the non-default value was only used in one place
- `write_manifest()`
  - Get rid of `job` keyword argument - this was unused
- `record_history()`
   - Moved fully into `Schema()`, since it seems to only be used internally at the end of `run()`
   - Open to reconsidering this
  
## Follow-up PRs

- Clean up docs gen to take advantage of `Schema()`
- Clean up server code to take advantage of `Schema()`
- Add linting to CI for schema_obj.py
- Refactor fundamental schema functions to clean up, fix latent bugs (tuple handling), improve performance (valid), and improve error messages
- Fix default value issue(s) (#1146,  #1150)


nmoroze:
Yep!

siliconcompiler/flows/asicflow.py
```diff
@@ -101,7 +101,7 @@ def setup(chip, flowname='asicflow'):
 
     # Clear old flowgraph if it exists
     if flowname in chip.getkeys('flowgraph'):
-        del chip.cfg['flowgraph'][flowname]
```
aolofsson, line 104: MIght cover up some other issue?

siliconcompiler/schema/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .schema_cfg import SCHEMA_VERSION
```
aolofsson, line 1: May want to reconsider tree struture org at some point

aolofsson (approved):

PR #1155, opened 2022-10-03
Enhancements to Verilator driver

nmoroze:
- Implement ['option', 'warningoff']
- Use ['option', 'trace'] instead of passthrough for enabling tracing
- Supply passthroughs for CFLAGS and LDFLAGS
- Error out if Make post-process step fails


aolofsson (approved):
Nice changes!

Issue #1134, opened 2022-08-31
SC crashes when run in non-existent directory


nmoroze:
Kind of a funny edge case, but to reproduce in two terminal sessions:

```bash
$ mkdir ~/temp # TERMINAL A
$ cd ~/temp # TERMINAL A
$ rm -rf ~/temp # TERMINAL B
$ sc -version # TERMINAL A
```

You get a traceback like:
```
Traceback (most recent call last):
  File "/home/noah/pyenvs/scdev/bin/sc", line 33, in <module>
    sys.exit(load_entry_point('siliconcompiler', 'console_scripts', 'sc')())
  File "/home/noah/code/siliconcompiler/siliconcompiler/apps/sc.py", line 36, in main
    chip = siliconcompiler.Chip(UNSET_DESIGN)
  File "/home/noah/code/siliconcompiler/siliconcompiler/core.py", line 81, in __init__
    self.cwd = os.getcwd()
FileNotFoundError: [Errno 2] No such file or directory
```

should probably handle this a bit more gracefully :)

nmoroze:
Closed by #1158.

PR #1158, opened 2022-10-04
Add an explanatory error message if SC is run from a non-existant directory

WRansohoff:
Waiting for something to build, I thought I should get back into the habit of looking at small SC bugs when I have a spare minute.

This change should address #1134, by printing an explanatory error if SC is run from a directory that does not exist. It still involves  a stack trace because I used the `self.error` method, but it's a bit more user-friendly:

```
  File "<stdin>", line 1, in <module>
  File "siliconcompiler/siliconcompiler/core.py", line 85, in __init__
    self.error("""SiliconCompiler must be run from a directory that exists.
  File "siliconcompiler/siliconcompiler/core.py", line 5203, in error
    raise SiliconCompilerError(msg)
siliconcompiler.core.SiliconCompilerError: SiliconCompiler must be run from a directory that exists.
If you are sure that your working directory is valid, try running `cd $(pwd)`.
```

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -78,8 +78,12 @@ class Chip:
 
         # Local variables
         self.scroot = os.path.dirname(os.path.abspath(__file__))
-        self.cwd = os.getcwd()
         self._error = False
+        try:
+            self.cwd = os.getcwd()
+        except FileNotFoundError:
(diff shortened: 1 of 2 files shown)


nmoroze (approved):
Looks good to me, thanks!

Issue #1135, opened 2022-09-03
Specifying twice the same file results in yosys failure


suzizecat:
Using SV sources and 
```python
chip.set('input',"verilog", ["src/myfile.sv","src/myfile.sv"])             # define list of source files
chip.set('option','frontend', 'systemverilog')
```

It seems that the SV files are actually added twice in the output from conversion before synthesis and therefore Yosys crashes with "redefining <modulename>".

It would be better to check for duplicate source files and warn the user about those while filtering them out before synthesis.

nmoroze:
Thanks for opening the issue, looks like it was fixed by #1159. Feel free to reopen if this didn't do what you expect.

PR #1159, opened 2022-10-04
Deduplicate Surelog file inputs

WRansohoff:
One more small change, to address #1135.

If the same Verilog file is added to a Chip object twice, Surelog will duplicate its content in the pickled output, which causes synthesis failures with Yosys.

Like the reporter suggested, we can de-duplicate the inputs when we are assembling the import step's Surelog command, and print a warning if any duplicate files are found.

### DIFF
--- a/siliconcompiler/tools/surelog/surelog.py
+++ b/siliconcompiler/tools/surelog/surelog.py
@@ -104,23 +104,23 @@ def runtime_options(chip):
     # Deduplicated source files
     # Library directories.
     ydir_files = chip.find_files('option', 'ydir')
-    if ydir_files != list(set(ydir_files)):
+    if len(ydir_files) != len(set(ydir_files)):
         chip.logger.warning(f"Removing duplicate 'ydir' inputs from: {ydir_files}")
-    for value in list(set(ydir_files)):
+    for value in set(ydir_files):
         cmdlist.append('-y ' + value)
 
     # Library files.
     vlib_files = chip.find_files('option', 'vlib')
-    if vlib_files != list(set(vlib_files)):
+    if len(vlib_files) != len(set(vlib_files)):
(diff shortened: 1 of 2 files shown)


siliconcompiler/tools/surelog/surelog.py
```diff
@@ -101,18 +101,44 @@ def runtime_options(chip):
 
     cmdlist = []
 
-    # source files
-    for value in chip.find_files('option', 'ydir'):
+    # Deduplicated source files
+    # Library directories.
+    ydir_files = chip.find_files('option', 'ydir')
+    if ydir_files != list(set(ydir_files)):
```
nmoroze, line 107: I think `set()` isn't guaranteed to preserve order, so I'm not sure if this always works. How about something like 

```python
if len(ydir_files) != len(set(ydir_files)):
  ... 
```

siliconcompiler/tools/surelog/surelog.py
```diff
@@ -101,18 +101,44 @@ def runtime_options(chip):
...
+        chip.logger.warning(f"Removing duplicate 'ydir' inputs from: {ydir_files}")
+    for value in list(set(ydir_files)):
```
nmoroze, line 109: `set()` is an iterator, so could just do:

```python
for value in set(ydir_files):
  ...
```

nmoroze (changes requested):
Thanks for implementing this! A few small comments, but otherwise looks good.

siliconcompiler/tools/surelog/surelog.py
WRansohoff: Good point, thanks.

nmoroze (approved):
Great, thanks!

aolofsson:
Nice feature to protect the user against casua mistakes!

PR #1161, opened 2022-10-05
Remove setup workaround for old Surelog git repo URL

WRansohoff:
Since Noah has updated the version of Surelog that we are using, I don't think we need this workaround in the install script anymore.

We added it because Surelog's `antlr4` submodule used a `git://` URL for the repository, and GitHub stopped supporting anonymous pulls from some SSH endpoints. Newer versions of Surelog use an `https://` URL, so we don't need to explicitly specify that anymore.


nmoroze (approved):
Looks good to me, thanks!

aolofsson (approved):

PR #1164, opened 2022-10-10
SC client messaging update

WRansohoff:
In preparation for some SCServer updates, this change lets the client print a bit more information about an ongoing remote job run.

It will print the current step and a tail of the task's log if the server provides that information. If not, it will continue printing the generic "Job is still running" message for compatibility with old (and current) versions of the server.

### DIFF
--- a/siliconcompiler/client.py
+++ b/siliconcompiler/client.py
@@ -113,9 +113,9 @@ def remote_run(chip):
                                    int(time.monotonic() - step_start), cur_step))
                   if cur_log:
                       chip.logger.info(f"Last 10 lines of logfile:\n{cur_log}\n")
-          else:
-              chip.logger.info(f"Job is still running (%d seconds, step: unknown)"%(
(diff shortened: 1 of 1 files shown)


nmoroze (approved):
Helpful change! Would it be possible for the API to return the current step name rather than relying on the client parsing it from the log? That could be a change for a future PR, since I know that would require additional work on the sc-server side.

Also, how does this handle task parallelism? Can this be made to tell the user if multiple tasks are running at once?

WRansohoff:
Yes, my first instinct was to add a parameter to the response, but using the message body lets us be backwards-compatible and update the server separately. But since you mention it, I suppose we could still do that by having the client behave differently depending on which values are present in the response.

Right now, I'm having the server return information on the first running step that it finds, but if we do update the API to contain an extra field, there's no reason why we couldn't loop through all of them. It's a little bit moot at the moment though, because in the current beta we're only allowing the scheduler to queue one job at a time per user.

PR #1165, opened 2022-10-18
Adding .vh extension

aolofsson:
- Commonly used methodology for non-module verilog 2001 header files (by me and others).
 - https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md


nmoroze:
Since .vh files are generally non-module headers, is there any reason to include it in the libext? That's for picking up modules with `['option', 'ydir']`, but I figure someone would point to their vh files using `['option', 'idir']`.

aolofsson:
Yeah, not sure what I was thinking (tired), closing this.

PR #1170, opened 2022-11-07
[Draft] Facilitate loading SC modules from Python packages

nmoroze:
This PR is an experiment in supporting the distribution of SC setup modules/drivers as a Python package. 

It adds two main features to facilitate this:

1: Module lookup via setuptools' [Entry Points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html). This is Python's mechanism for a package to advertise that it provides certain functionality. It is the same mechanism used by pytest's plugin system.

This PR proposes having SC load modules from a set of entry points that map to our current set of setup modules:
- `siliconcompiler.checklists`
- `siliconcompiler.flows`
- `siliconcompiler.libs`
- `siliconcompiler.pdks`
- `siliconcompiler.tools`

If you create a package that provides dynamic modules for SiliconCompiler, you would include entries in that package's setup.cfg/setup.py that map those modules to the associated entry points.

This PR makes these entry points the default mechanism for loading packages, and then it falls back to the SCPATH method (with a deprecation warning) to ensure backwards-compatibility for now.

2: `package` field for filetype schema entries.

Since SCPATH was used for resolving both setup modules and filepaths, it was convenient for setup modules to point to data that sits in the same directory structure underneath SCPATH. The most common example was our informal convention that all tool drivers point their refdir to a `scripts/` directory next to them However, packages can no longer rely on this if they're not being set up with SCPATH. 

This was a convenient feature because absolute paths don't work in general - a `setup()` function may run in a different environment than the actual machine executing the task (in the case of a remote run or using a job scheduler/cluster). To facilitate something like the old mechanism, this PR proposes adding a `package` field to all file/dir-typed schema parameters that defines whether the path should be resolved relative to a certain package's directory. To keep setup concise, `package` can be exposed as an optional keyword arg to the `set()` function, and used like so:

```python
chip.set('tool', 'openroad', 'refdir', step, index, 'tools/openroad/scripts', package='siliconcompiler')
```

This path can then be resolved on any machine that has the Python package "siliconcompiler" installed.


nmoroze:
Going to close this after some discussion today - it was really addressing a different proposal (a unified `load(...)` that takes in strings) rather than our new `use()` method that takes in modules directly.

We have other ideas to handle some of the other things this PR could have helped with, including:
- target loading -- might be better to just facilitate this via a mechanism to allow additional CLI arguments when using `chip.create_cmdline()`
- docs -- can handle this via arguments to the docs generator extension

The last aspect is file path resolution. There's been some discussion about this, but no concrete solutions as far as I know. However, I think it would be cleaner if we didn't have to add an additional `package` kwarg to `set()` as proposed here.

Issue #1061, opened 2022-06-22
Output file hashing won't persist


nmoroze:
I notice we hash output files in `_runtask()`, but I think those hashes will probably get thrown away:

https://github.com/siliconcompiler/siliconcompiler/blob/main/siliconcompiler/core.py#L4013-L4022

Now that we do a partial merge manifest for each task, any modifications to non-blessed keypaths (metrics, records, "flowstatus") don't persist.

PR #1172, opened 2022-11-10
Fix input/output file hash tracking

WRansohoff:
This is a small change to address issue #1061 so that SiliconCompiler's manifests can be used to determine the file hashes for a design's final inputs and outputs, but it has performance implications that we may want to discuss.

The change allows the schema's file hashes to propagate through a flow by updating the `merge_manifest` method to include file-type parameters in partial updates. I also tried allowing `library` schemas (whose files sometimes get hashed) to be read from the prior manifest rather than re-imported, but I ran into issues with CI and runtime.

This impacts the performance of the manifest-merging, because `chip.get` is a fairly hefty call to add for each keypath. On the minimal heartbeat example, merge runtimes (measured using `time.monotonic` around the calls to `_merge_manifset`) increased from 2-15ms to 28-44ms. That is a significant increase, but still a large reduction from a non-`partial` merge which takes 110-130ms per step.

I experimented with parallelizing this logic using `multiprocessing` to let us remove the `_key_may_be_updated` check, but that ended up significantly increasing the runtime. I think that Python's GIL and the need for a thread-safe dictionary data structure may prevent efficient parallel writes.


siliconcompiler/core.py
```diff
@@ -3798,7 +3800,7 @@ def _runtask(self, step, index, status):
         self.set('arg', 'step', None, clobber=True)
         self.set('arg', 'index', None, clobber=True)
         os.makedirs('inputs', exist_ok=True)
-        #self.write_manifest(f'inputs/{design}.pkg.json')
+        self.write_manifest(f'inputs/{design}.pkg.json')
```
aolofsson, line 3803: Can we remove this? Is it used?

WRansohoff: I don't think that anything in SC relies on it, but adding it back may be a good idea.

Pros:
* Saves a copy of the manifest to each task's `inputs/` directory before each task is run; could help with debugging.
* There is a comment indicating that it should be done.

Cons:
* May add a very small amount of runtime to each task, as the manifest is processed and written to disk.

gadfort: I would vote for keeping it in. I've wished for it to be there several times

WRansohoff:
I've resolved the conflicts with `main` - I think the only open question is whether we should go back to setting (`arg`, `[step|index]`) to `None` early in `_runtask` (it still gets set to the step/index value later, before the tool is run).

Sorry to let these changes get stale; we never really resolved the question of whether it was worth pursuing #1173. Since we ended up with a `Chip.use` implementation which relies on `_merge_manifest`, though, I think we should abandon it for now.

That said, if we do manage to restrict the `PDK`, `Library`, and `Flow` schemas to their respective top-level key paths in the future, we might still be able to make the core `Schema` additive in a way that does not require complex merging logic.

nmoroze (approved):
Looks good to me! I'm not sure about the ['arg', 'step'/'index'] stuff - I don't think it'll mater either way, since in most cases I"d expect them to already be None at the point we get into _runtask(). Perhaps they'd be set if the user provides them, and that's why they get cleared out?

It'll be good to get this in and close #1061! It might be helpful to add a test (perhaps in a future PR) of the file hashing as well, so that we can make sure things don't regress if we do revisit this logic at any point.

PR #1173, opened 2022-11-11
Draft idea: Remove `merge_manifest` function

WRansohoff:
This PR contains an alternative, more radical solution to #1061 (compared with #1172): maybe we could remove the whole process of merging manifests.

Considering that we originally implemented partial merges for the sake of performance, maybe it's worth considering a less holistic approach to the process of updating the manifest between tasks. Instead of striving for a one-size-fits-all set of rules to merge arbitrary schema parameters, maybe we could handle parallel input tasks by fully re-loading one task's output manifest, and then importing known output fields from the others (metrics, reports, flowgraph status, file hashes).

Being able to resume or review a run using only its manifest, with no critical state saved during the run, seems like a generally good goal.

There are a few workarounds in this PR to get all of the CI tests to pass, and I can think of good reasons to avoid making such a deep change, so we should review this more closely if it seems like it could be a good idea.


nmoroze:
This is a cool idea! Some initial thoughts to frame later discussion:
- One subtle behavior difference between this solution and  #1172 is that file hashes that are computed on file requirements that only apply to merged parallel tasks besides the first will get dropped.

- Similarly, I think while this fixes the problem I tried to solve in #1145 for a basic linear flowgraph, the issue could still arise for certain parallel jobs.

- You mention "Being able to resume or review a run using only its manifest, with no critical state saved during the run, seems like a generally good goal." I agree this is a good design goal, but it's not clear to me how this change facilitates that any more than what we had previously.

WRansohoff:
Considering that the `merge_manifest` function is important for our new `Chip.use` functionality, I think we should close this in favor of the simpler solution in #1172

PR #1175, opened 2022-11-27
python: don't catch BaseException, such as KeyboardInterruptException

oharboe:
"except:" will catch e.g. KeyboardInterruptException, which can result in unintended effects like ctrl-c not working.

Instead use "except Exception:".

The code could be further refined to catch narrower exceptions than all exceptions, such as dealing with file not found differently than problems parsing a .yaml file, but that's for another pull request.

https://docs.python.org/3/library/exceptions.html

### DIFF
--- a/siliconcompiler/apps/sc_ping.py
+++ b/siliconcompiler/apps/sc_ping.py
@@ -42,7 +42,7 @@ def main():
            (not 'address' in remote_cfg):
             print('Error reading remote configuration file.')
             sys.exit(1)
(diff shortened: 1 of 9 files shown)


nmoroze (approved):
Thank you very much for the PR, you're completely right!

We merged a major change recently, so I took care of the rebase (one of the functions you touched moved). I also took the liberty of making the exceptions for that function even more specific, since it was a fairly clear case.

PR #1183, opened 2023-01-02
Schema layout constraints update

aolofsson:
- Adding component and pin section to constraints dictionary.
- Adding lambda to prepare for process agnostic design planning.


siliconcompiler/schema/schema_cfg.py
```diff
@@ -3507,96 +3595,185 @@ def schema_asic(cfg):
...
+    scparam(cfg, ['constraint', 'component', instance, 'flip'],
+            sctype='bool',
+            shorthelp="Constraint: Component flip option",
+            switch="-constraint_component_flip 'inst <bool>'",
+            example=[
+                "cli: -constraint_component_flip 'i0 true'",
+                "api: chip.set('constraint', 'component', 'i0', 'flip', 'true')"],
+            schelp="""
+            Boolean parameter specifying that the instanced library component should be flipped
+            around the vertical axis befong being placed on the substrate. The need to
```
gadfort, line 3753: befong -> before

siliconcompiler/schema/schema_cfg.py
```diff
@@ -3507,96 +3595,185 @@ def schema_asic(cfg):
...
+            sctype='(float,float,float)',
+            shorthelp="Constraint: Pin placement",
+            switch="-constraint_pin_placement 'inst <(float,float, float)>'",
+            example=[
+                "cli: -constraint_pin_placement 'i0 (2.0,3.0,0.0)'",
+                "api: chip.set('constraint', 'pin', 'i0', 'placement', (2.0,3.0,0.0)"],
+            schelp="""
+            Placement location of a named pin, specified as a (x,y,z) tuple of
+            floats. The location refers to the placement of the center of the
+            pin. Rhe 'placement' parameter is a goal/intent, not an exact specification.
```
gadfort, line 3769: Rhe -> The

siliconcompiler/schema/schema_cfg.py
```diff
@@ -3507,96 +3595,185 @@ def schema_asic(cfg):
...
+            The compiler and layout system may adjust sizes to meet competing
+            goals such as manufacturing design  rules and grid placement
```
gadfort, line 3771: "  " -> " "

siliconcompiler/schema/schema_cfg.py
```diff
@@ -3507,96 +3595,185 @@ def schema_asic(cfg):
...
+            schelp="""
+            Boolean parameter specifying that the instanced library component should be flipped
+            around the vertical axis befong being placed on the substrate. The need to
+            flip a component depends on the component footprint. Most dies have pads
+            facing up and so must be flipped when assembled face down (eg. flip-chip,
+            WCSP).""")
+
+    # PINS
+    scparam(cfg, ['constraint', 'pin', instance, 'placement'],
+            sctype='(float,float,float)',
```
gadfort, line 3760: How would this work on an ASIC design? there the natural thing would be to set the `z` parameters to the metal name. Or would you not be able to constrain that here and instead use the `sc_hpinmetal/sc_vpinmetal`?

siliconcompiler/schema/schema_obj.py
```diff
@@ -485,10 +488,14 @@ def write_tcl(self, fout, prefix=""):
...
             valstr = utils.escape_val_tcl(value, typestr)
 
+            # Turning scalars into lists
             if not (typestr.startswith('[') or typestr.startswith('(')):
-                # treat scalars as lists as well
                 valstr = f'[list {valstr}]'
 
+            # TODO: Temp fix to get rid of empty args
+            if valstr=='':
+                valstr = f'[list ]'
```
gadfort, line 497: What would the longer term solution be?

siliconcompiler/tools/yosys/yosys.py
```diff
@@ -402,15 +402,15 @@ def get_synthesis_corner(chip):
 
     # determine corner based on setup corner from constraints
     corner = None
-    for constraint in chip.getkeys('constraint'):
-        if "setup" in chip.get('constraint', constraint, 'check') and not corner:
-            corner = chip.get('constraint', constraint, 'libcorner')
+    for constraint in chip.getkeys('constraint', 'timing'):
```
gadfort, line 405: should we check if `'constraint','timing'` is valid first?

siliconcompiler/schema/schema_cfg.py
```diff
@@ -3507,96 +3595,185 @@ def schema_asic(cfg):
...
+                "api: chip.add('constraint', 'timing', 'worst','check','setup')"],
             schelp="""
             List of checks for to perform for the scenario. The checks must
             align with the capabilities of the EDA tools and flow being used.
             Checks generally include objectives like meeting setup and hold goals
             and minimize power. Standard check names include setup, hold, power,
             noise, reliability.""")
 
+    # COMPONENTS
+    scparam(cfg, ['constraint', 'component', instance, 'outline'],
```
gadfort, line 3692: What is the intent of this parameter? It's not clear to me what an outline constraint means?

siliconcompiler/core.py
```diff
@@ -675,6 +675,16 @@ def load_checklist(self, name):
...
             self.error(f'Checklist module {name} not found in $SCPATH or siliconcompiler/checklists/.')
 
+    ##########################################################################
+    def use(self, module):
+        '''
+        Loads a SiliconCompiler module into the current chip object by calling
+        a module.setup() method.
+        '''
+
+        self.import_library(module.setup())
```
gadfort, line 685: Maybe add a type check (and check if `.setup()` is present) to provide a user friendly error if they call this incorrectly?

siliconcompiler/schema/schema_cfg.py
```diff
@@ -188,14 +199,90 @@ def schema_cfg():
...
+            shorthelp="Schematic: pin definition",
+            switch="-schematic_pin 'name <(str,str)>'",
+            example=[
+                "cli: -schematic_pin 'out[7:0] (output,signal)'",
+                "api: chip.set('schematic', 'pin', 'out[7:0]', ('output','signal'))"],
+            schelp="""
+            Design primary pin definitions, specified (direction,type) tuples
+            on a per pin basis. Allowed directions are: input, output, and
+            inout. Allowed types are: analog, clock, ground, power, signal.
+            An empty 'type' defaults to signal.""")
```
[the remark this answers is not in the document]
nmoroze: Not yet, but I think we should add this - opened an issue to track:  #1199.

siliconcompiler/schema/schema_cfg.py
```diff
@@ -188,14 +199,90 @@ def schema_cfg():
...
+    scparam(cfg, ['schematic', 'parameter', name],
+            sctype='[(str,str)]',
+            shorthelp="Schematic: parameter definition ",
+            switch="-schematic_parameter 'obj <(str,str)>'",
+            example=[
+                "cli: -schematic_parameter 'i0 (speed,fast)'",
+                "api: chip.set('schematic','parameter', 'i0', ('speed','fast')"],
+            schelp="""
+            List of parameter definitions attached to a named pin, net, or
+            component pecified as (key,value) tuples.""")
```
nmoroze, line 282: typo: pecified -> specified

siliconcompiler/tools/surelog/surelog.py
```diff
@@ -109,51 +109,109 @@ def runtime_options(chip):
...
-    # Deduplicated source files
-    # Library directories.
+    #####################
+    # Library directories
+    #####################
+
     ydir_files = chip.find_files('option', 'ydir')
+
+    for item in chip.getkeys('library'):
+        ydir_files.extend(chip.find_files('library', item, 'option', 'ydir'))
```
nmoroze, line 119: Seems OK, although I wonder if how exactly soft libraries are incorporated requires more thought. Seems unrelated to main focus of PR.

nmoroze (commented):
Looks good to me overall, I'm pretty happy with the component constraint schema (I agree with Peter's comments there). Left some additional comments inline.

I'm a bit less convinced about adding the schematic section at the moment. This definitely seems to be a case of adding design data to the schema. Did we come up with any justification for adding this as opposed to keeping it in a separate project for now?

siliconcompiler/core.py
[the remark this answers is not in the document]
gadfort: @WRansohoff as for Yosys, I'm pretty sure the blackbox thing is gone in favor of a ".lib". Which makes macros a lot easier to implement.

siliconcompiler/schema/schema_cfg.py
```diff
@@ -141,7 +139,20 @@ def schema_cfg():
             schelp="""Name of the top level module or library. Required for all
             chip objects.""")
 
+    # Lambda value
+    scparam(cfg,['lambda'],
```
WRansohoff, line 143: Does this belong in the top-level schema? It seems like it might only apply to a specific type of PDK, if I'm reading the description correctly.

siliconcompiler/schema/schema_cfg.py
```diff
@@ -188,14 +199,90 @@ def schema_cfg():
...
+            sctype='str',
+            shorthelp="Schematic: component (instance)",
+            switch="-schematic_component 'name <str>'",
+            example=[
+                "cli: -schematic_component 'i0 za001'",
+                "api:  chip.set('schematic', 'component', 'i0', 'za001')"],
+            schelp="""Unique manufacturer part number (MPN) of the  named
+            component.""")
+
+    scparam(cfg, ['schematic', 'net', name],
```
WRansohoff, line 234: +1 to Noah's comment that we might want to discuss some of the `'schematic'` values a bit more. I like the idea of letting designers tell SC where core components like macros and pins should be placed, but net names might be closer to the "duplicating details which are represented in source files" end of the spectrum.

siliconcompiler/schema/schema_cfg.py
WRansohoff: It looks like a way to represent a macro's size in the schema? We may be able to get those values from the `('asic', 'diearea')` parameter in the macro's schema, or the LEF file if the designs were auto-sized and/or built outside of SC.

Although, the `diearea` parameter doesn't support rectilinear shapes, and this may be intended to represent components in other contexts besides asic physical design flows.

siliconcompiler/schema/schema_cfg.py
```diff
@@ -3507,96 +3595,185 @@ def schema_asic(cfg):
...
+            schelp="""
+            Placement location of a named component, specified as a (x,y,z) tuple of
+            floats. The location refers to the placement of the center/centroid of the
+            component. The 'placement' parameter is a goal/intent, not an exact specification.
+            The compiler and layout system may adjust sizes to meet competing
+            goals such as manufacturing design  rules and grid placement
+            guidelines. The 'z' coordinate shall be set to 0 for planar systems
+            with only (x,y) coordinates. Discretized systems like PCB stacks,
+            package stacks, and breadboards only allow a reduced
+            set of floating point values (0,1,2,3). The user specifying the
```
WRansohoff, line 3724: This isn't a big deal, but we may need to account for floating-point imprecision if we use the Z-values as discrete steps. We should be okay if we use round numbers and don't perform arithmetic on them, but for example:

```
>>> a = 1.6
>>> '%.20f'%a
'1.60000000000000008882'
```

siliconcompiler/core.py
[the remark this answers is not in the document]
nmoroze: Yes, we should discuss. Moving to one chip object/module seems tedious, it would require creating 10s of files of boilerplate for more advanced libraries.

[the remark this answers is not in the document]
aolofsson: Yes, but it's done once per library...considering that downloading and procuring libraries for advanced nodes is a huge investment (and involves terabytes of data) I am not sure the boiler plate effort of putting together a python module per lib is big enough? It still seems like a corner case that should not drive the architecture?

siliconcompiler/schema/schema_cfg.py
```diff
@@ -3507,96 +3595,185 @@ def schema_asic(cfg):
...
             noise, reliability.""")
 
+    # COMPONENTS
+    scparam(cfg, ['constraint', 'component', instance, 'outline'],
+            sctype='[(float,float)]',
+            shorthelp="Constraint: Component outline",
+            switch="-constraint_component_outline 'i0 <(float,float)'>",
+            example=[
+                "cli: -constraint_component_outline 'i0 (0,0)'",
+                "api: chip.add('constraint', 'component', 'i0', 'outline', (0,0))"],
```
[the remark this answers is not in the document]
aolofsson: Although it would require fixing the auto generated setget test.

siliconcompiler/schema/schema_cfg.py
aolofsson: Added to schema

siliconcompiler/tools/yosys/yosys.py
aolofsson: Actually, should not be needed since timing is always defined.

siliconcompiler/core.py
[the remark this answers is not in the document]
aolofsson: ...on the other hand, reconsidering. Besides the library, there are instances where we have a large number of verilog files where by convention people put one module per file. There is no way we would want to have a python wrapper for each module. Returning a list of objects would be pretty nice for a library of open source verilog components, each chip object possible representing a trivial module (nand, flip flop)....

siliconcompiler/schema/schema_cfg.py
```diff
@@ -3507,96 +3595,185 @@ def schema_asic(cfg):
...
+            shorthelp="Constraint: Component placement",
+            switch="-constraint_component_placement 'inst <(float,float, float)>'",
+            example=[
+                "cli: -constraint_component_placement 'i0 (2.0,3.0,0.0)'",
+                "api: chip.set('constraint', 'component', 'i0', 'placement', (2.0,3.0,0.0)"],
+            schelp="""
+            Placement location of a named component, specified as a (x,y,z) tuple of
+            floats. The location refers to the placement of the center/centroid of the
+            component. The 'placement' parameter is a goal/intent, not an exact specification.
+            The compiler and layout system may adjust sizes to meet competing
```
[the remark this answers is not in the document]
aolofsson: thanks, yes!

siliconcompiler/schema/schema_cfg.py
```diff
@@ -3507,96 +3519,184 @@ def schema_asic(cfg):
...
+            around the vertical axis before being placed on the substrate. The need to
+            flip a component depends on the component footprint. Most dies have pads
+            facing up and so must be flipped when assembled face down (eg. flip-chip,
+            WCSP).""")
+
+    # PINS
+    scparam(cfg, ['constraint', 'pin', name, 'placement'],
+            sctype='(float,float,float)',
+            shorthelp="Constraint: Pin placement",
+            switch="-constraint_pin_placement 'inst <(float,float, float)>'",
```
gadfort, line 3672: inst->name

gadfort (changes requested):
Just one more pin name->inst remaining as far as I can see

gadfort (approved):
LGTM

PR #1184, opened 2023-01-03
Fix test_steplist_keep_reports

nmoroze:
This PR fixes a regression in our daily tests caused by #1147. `Schema.copy()` performs a deep-copy, so this should be equivalent.


WRansohoff (approved):
Good catch!

Out of curiosity, is `Chip.cfg` still used anywhere, or should we try to replace it with `Chip.schema.cfg` anywhere we see it?

nmoroze:
Good question - `Chip.cfg` shouldn't be used anywhere, since it no longer exists. Anything that still references it is definitely a bug.

A quick fix that should work in just about all cases is to use `Chip.schema.cfg` instead, but if possible the ideal solution is to use schema access methods directly instead (I think the goal should be to treat it as a private member, and it might be worth renaming it at some point to indicate this).

PR #1185, opened 2023-01-03
yosys: update ASIC flow to handle ABC constraints, techmapping, wire naming, and dont_use

gadfort:
Changes to Yosys tool:
- Metrics are handled via JSON report, which makes importing easier and less error prone with line parsing
- Liberty files are copied into input/ and unzipped (if zipped), cells in ignore list are marked dont_use to prevent ABC from using + a few other parsing issues are handled here to prevent faults in flow.
- New var's:
- - preserve_modules - list of modules in input files to prevent flatten from "flattening"
- - flatten - True/False (default True), invoke `synth` with the flatten option
- - autoname - True / False (default True), call autoname to rename wires based on registers 
- - map_adders - False / path to map_adders (default False), map adders in Yosys
- - synthesis_corner - corner to use for synthesis
- - dff_liberty - liberty file to use for flip-flop mapping, if not specified the first in the logiclib is used
- - abc_constraint_driver - buffer that drives the abc techmapping, defaults to first buffer specified
- - abc_constraint_load - load for the abc techmapping in fF, if not specified it will not be used
- - abc_clock_period - clock period to use for synthesis in ps, if more than one clock is specified, the smallest period is used.
- - abc_clock_derating - (0 - 1) used to derate the clock period to further constrain the clock
- - synthesis_libraries - internally used parameter to keep track of logiclibs
- - synthesis_libraries_macros - internally used parameter to keep track of macrolibs


siliconcompiler/tools/yosys/syn_asic.tcl
```diff
@@ -52,27 +44,59 @@ foreach item $sc_scenarios {
...
-set mainlib      [dict get $sc_cfg library $sc_mainlib model timing $sc_delaymodel $libcorner]
+if {[dict get $sc_cfg tool $sc_tool var $sc_step $sc_index autoname] == "True"} {
+    # use autoname to preserve some design naming
+    # by doing it before dfflibmap the names will be slightly shorter since they will
+    # only contain the $DFF_P names vs. the full library name of the associated flip-flop
+    yosys autoname
+}
 
-yosys dfflibmap -liberty $mainlib
+yosys dfflibmap -liberty "inputs/dff_liberty.lib"
```
aolofsson, line 96: This hard coded name seems problematic.

siliconcompiler/tools/yosys/syn_asic.tcl
```diff
@@ -2,40 +2,32 @@
...
-	foreach lib $sc_macrolibs {
-	    if {[dict exists $sc_cfg library $lib model timing $sc_delaymodel $libcorner]} {
-		puts "SC: Reading liberty file (corner=$libcorner, lib=$lib, mode=$sc_delaymodel)"
-		set lib_file [dict get $sc_cfg library $lib model timing $sc_delaymodel $libcorner]
-		yosys read_liberty -lib $lib_file
-		append stat_libs "-liberty $lib_file "
-	    }
-	}
-    }
+set lib_args ""
```
aolofsson, line 19: Are there opportunities in the schema/python to put more on the emphasis on the python wrapper for the tool to simplfy the TCL code? Having logic that straddles tcl/python bondary has always beeb a proble, epsecially when the TCL interface is limited like in Yosys.

siliconcompiler/tools/yosys/syn_asic.tcl
```diff
@@ -95,32 +119,49 @@ if {[dict exists $sc_cfg tool $sc_tool var $sc_step $sc_index strategy]} {
...
     yosys insbuf -buf {*}$sc_buf_split
 }
 
-yosys splitnets
+yosys clean -purge
 
-yosys clean
+# turn off echo to prevent the stat command from showing up in the json file
+yosys echo off
+yosys tee -o ./reports/stat.json stat -json -top $sc_design {*}$stat_libs
```
aolofsson, line 163: Nice!

siliconcompiler/tools/yosys/syn_asic.tcl
gadfort: What would you suggest? I could pass it in via the schema and set the name in python, would that be better?

siliconcompiler/tools/yosys/syn_asic.tcl
gadfort: Are you suggesting that the python wrapper generate a list for the stat and abc commands and pass that in instead? So we don't build the argument list here? That would certainly simplify some of the openroad scripts

gadfort: I changed it, such that the path is set by the python script, however, it is still a constant (just now it's controlled in one place and from the python tool driver)

siliconcompiler/tools/yosys/syn_asic.tcl
gadfort: changed it, such that the path is set by the python script, however, it is still a constant (just now it's controlled in one place and from the python tool driver)

siliconcompiler/tools/yosys/syn_asic.tcl
gadfort: I moved this code around a little to simplify it a bit in each place

aolofsson (approved):

PR #1186, opened 2023-01-03
Update the scheduler module to work with Slurm power-save scripts

WRansohoff:
This PR contains a handful of changes that came out of our recent infrastructure explorations.

* Cluster partitions can be overridden on a per-step basis.
* Support implementations which do not require slurmdbd accounting.
* Support extra job statuses used by implementations which use power-save scripts for elastic scaling.

### DIFF
--- a/siliconcompiler/scheduler.py
+++ b/siliconcompiler/scheduler.py
@@ -65,8 +65,6 @@ def _deferstep(chip, step, index, status):
         # TODO: LSF support is untested and currently unsupported.
         schedule_cmd = ['lsrun']
 
-    time.sleep(1.0)
-
     # Create a command to defer execution to a compute node.
     script_path = f'{cfg_dir}/{step}{index}.sh'
     with open(script_path, 'w') as sf:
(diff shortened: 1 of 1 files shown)


siliconcompiler/scheduler.py
```diff
@@ -16,63 +17,87 @@ def _deferstep(chip, step, index, status):
...
 
-    # Job data is not encrypted, so it can be run in shared storage.
-    # Write out the current schema for the compute node to pick up.
-    job_dir = chip._getworkdir()
-    cfg_dir = f'{job_dir}/configs'
-    cfg_file = f'{cfg_dir}/{step}{index}.json'
-    if not os.path.isdir(cfg_dir):
-        os.mkdir(cfg_dir)
-    chip.write_manifest(cfg_file)
+    time.sleep(1.0)
```
aolofsson, line 68: Whe the need for sleep here?

WRansohoff: There shouldn't be a need for it anymore; removed.

I think that I originally inserted this delay to give NFS caches some time to catch up after seeing some unwarranted 'file not found' errors. The long-term fix for that was to add prolog/epilog scripts which refresh the mounts before and after each job, though.

aolofsson (approved):
Nice!

PR #1196, opened 2023-01-08
core: change the order of searching SCPATH

gadfort:
Changes:
- SC search order to: 'cwd / scpath / ::env(SCPATH) / scroot' instead of 'scroot / cwd / scpath / ::env(SCPATH)'

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -972,10 +972,11 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
             return os.path.abspath(filename)
 
         # Otherwise, search relative to scpaths
-        scpaths = [self.scroot, self.cwd]
+        scpaths = [self.cwd]
         scpaths.extend(self.get('option', 'scpath'))
         if 'SCPATH' in os.environ:
             scpaths.extend(os.environ['SCPATH'].split(os.pathsep))
+        scpaths.append(self.scroot)
 
         searchdirs = ', '.join(scpaths)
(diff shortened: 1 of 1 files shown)


aolofsson:
Are you sure about this? The intent behind the original path order was to not let people override the built in paths.  

https://tldp.org/HOWTO/Path-12.html

gadfort:
@aolofsson I see the point, lets discuss tomorrow. I think this is partially moot based on some of the other changes planned, but the remainder I think should be considered.

nmoroze:
I think this is a nice change, I actually implemented it to fix an issue I had while working on some experiments (but Peter beat me to the PR!)

Searching the CWD first is important to make referencing your current directory work properly, e.g.:

```python
chip.set('option', 'idir', '.')
```

(this is what #1178 is meant to track). 

And I think it actually seems reasonable to have the user be able to override the built-in paths if they want to. 

Either way, like Peter said I think this will be moot once we fully switch to importing setup modules via Python, since we should then be able to deprecate SCPATH.

nmoroze:
@gadfort I just opened PR #1198 or the changes I previously committed (sorry that I forgot to do that, I feel bad for creating duplicate work!)

Are you ok with closing this in favor of the new PR? It updates the docs and has one other tweak that is necessary to fix #1178. I'd also be happy with merging this one and rebasing mine on top of it, or add the additional changes into this one.

gadfort:
@noah yes, I'll close this in favor of the other one.

PR #1200, opened 2023-01-11
update openroad tool

gadfort:
Changes:
- Updates OpenROAD flow steps to more closely mirror ORFS.
- Reorganize PDKs/Libs used by OpenROAD in open PDKs
- Fill in metrics from OpenROAD

Todo:
- Fill in openroad parameter help information


siliconcompiler/libs/asap7sc7p5t.py
```diff
@@ -82,12 +82,25 @@ def _setup_lib(libname, suffix):
...
 
-    # Techmap
+    # Yosys techmap
     if libname.endswith('rvt'):
         # TODO: write map files for other groups
         lib.add('asic', 'file', 'yosys', 'techmap',
                     libdir + '/techmap/yosys/cells_latch.v')
 
+    # Defaults for OpenROAD tool variables
+    lib.set('asic', 'var', 'openroad', 'place_density', ['0.77'])
```
aolofsson, line 92: Yes, we definitely need to resolve the grouping. Who "owns" a variable like place_density? Is that a PDK variable, stdlib variable, or user variable. It definitely gets messy if we have multiple setters and we get an ordering issue.

aolofsson (approved):
Looks great! This is a massive improvement to the OR scripts. Some of the changes are beyond my understanding of OR, so I will have to trust you.

PR #1202, opened 2023-01-11
convert strings to raw strings when they contain escape characters

gadfort:
Fixes:
- Invalid escape character warnings in CI


nmoroze (approved):
Looks good to me, thanks Peter!

PR #1203, opened 2023-01-11
Fix daily tests

nmoroze:
- Add missing filesets
- Fix KLayout input logic to handle filesets properly
- Get rid of unecessary OR settings
- Clean up some imports while we're at it


gadfort (approved):
LGTM

PR #1204, opened 2023-01-12
fix daily CI after openroad update

gadfort:
Fixes:
- incorrect library used for requirements

Adds:
- knob to control pdngen

### DIFF
--- a/siliconcompiler/tools/openroad/openroad.py
+++ b/siliconcompiler/tools/openroad/openroad.py
@@ -165,8 +165,10 @@ def setup(chip, mode='batch'):
             chip.set('tool', tool, 'var', step, index, variable, value,
                      clobber=False)
 
-        keypath = ','.join(['library', mainlib, 'asic', 'var', tool, variable])
-        chip.add('tool', tool, 'require', step, index, keypath)
+            keypath = ','.join(['library', mainlib, 'asic', 'var', tool, variable])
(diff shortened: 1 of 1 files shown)


siliconcompiler/tools/openroad/openroad.py
```diff
@@ -158,14 +158,14 @@ def setup(chip, mode='batch'):
...
+        if chip.valid('library', mainlib, 'asic', 'var', tool, variable):
+            value = chip.get('library', mainlib, 'asic', 'var', tool, variable)
             # Clobber needs to be False here, since a user might want to
             # overwrite these.
             chip.set('tool', tool, 'var', step, index, variable, value,
                      clobber=False)
 
-        keypath = ','.join(['library', lib, 'asic', 'var', tool, variable])
+        keypath = ','.join(['library', mainlib, 'asic', 'var', tool, variable])
         chip.add('tool', tool, 'require', step, index, keypath)
```
nmoroze, line 169: I think this block might want to be indented under the `chip.valid(...)` if statement, since the requirement seems optional. I think this would still fail out if the mainlib didn't actually define a variable.

(it's still valuable to conditionally add the keypath to "require" since this will make sure SC checks that the paths resolve correctly before runtime).

nmoroze (approved):
Just one small comment I left inline, otherwise LGTM!

PR #1206, opened 2023-01-12
Vivado improvements

nmoroze:
This PR fixes up and enhances SC's Vivado support. Changes:

- Modernize/fix up the driver
- Add metrics parsing
- Updates for fileset change
- Fix blinky example to make sure it at least builds (verified locally, since we don't have CI for this)
- Small core change to skip version checking if no target versions are specified (seems like a reasonable default to me?)


aolofsson (commented):
Looks great. Nice changes! I find it funny that every vendor now has some kind of report_qor command that outputs json, but they all have different syntax. At least its an improvement over grepping log files....

aolofsson (approved):

PR #1207, opened 2023-01-13
remove tool submodules

gadfort:


CMakeLists.txt
```diff
@@ -11,6 +11,14 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 find_package(PythonExtensions REQUIRED)
 find_package(Cython REQUIRED)
 
+include(FetchContent)
+FetchContent_Declare(openroad
+  GIT_REPOSITORY    https://github.com/The-OpenROAD-Project/OpenROAD.git
+  GIT_TAG           81706aef55de40a6f2eccadf25c20f418a554ab6
```
aolofsson, line 17: Looks good. Maybe you left this for later, but a hard coded hash in the cmake file makes me really uncomfortable. Can we collect these magic constants in one place? 

We have an attempt with the version number here:
https://github.com/siliconcompiler/siliconcompiler/blob/main/siliconcompiler/_metadata.py

setup/install-openroad.sh
```diff
@@ -1,36 +1,19 @@
 #!/bin/sh
 src_path=$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)
```
aolofsson, line 3: So all the stuff in our install file is no longer needed?

setup/install-surelog.sh
```diff
@@ -6,10 +6,14 @@ src_path=$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)
...
 sudo apt-get install -y default-jre
 
-cd ${src_path}/..
-git submodule update --init --recursive third_party/tools/surelog
-cd third_party/tools/surelog
+mkdir -p deps
+cd deps
+
+git clone https://github.com/chipsalliance/Surelog.git
+cd Surelog
```
aolofsson, line 13: Same comment regarding clearly auto documenting hash dependencies (for developers and end users).

aolofsson (approved):

setup/install-openroad.sh
gadfort: OpenROAD has their own dependency installer and builder, which supports several OSs so it's easier to simply lean on theirs instead

CMakeLists.txt
gadfort: Yes we can (and should), I have a "mockup" of one way to handle it. We can either merge this in or wait (since there is no urgency on this) to find a "limited" magic number solution

gadfort: In retrospect, once @WRansohoff verifies that the changes to CMakeLists doesn't break the install, I think we will merge this. The change to remove the "magic" numbers will come in a separate PR since there may be some nice additional things we can do with that which would be orthogonal to this PR.

WRansohoff: That build is currently running, and I just kicked off [a 'wheels' action](https://github.com/siliconcompiler/siliconcompiler/actions/runs/3913966431) which will attempt to build the package on all common platforms.

WRansohoff: Looks like the Wheels action may need updating to remove the submodule reference; I made a `remove-tools-submodules-wheels` branch to test that.

The local `pip install -e .` process works on a clean Ubuntu machine after updating `pyopenssl` and installing `bison`, but it only seems to install the Python package, not the command-line entry points such as `sc` and `sc-show`.

WRansohoff: Per our offline discussion, it looks like the command-line entry points do get installed, but in a different location: `~/.local/bin/`

gadfort:
@aolofsson I finally got the wheels build working on MacOS and Windows with the newest version of OpenROAD. I added the _tools.json and _tools.py to start tracking the tools and versions in a single place. I chose the json file to make it easier to update via a script instead of adding it to the python file and rewrite it. In a followon PR, we can add all remaining tools. I'm open to suggestions if you think a different form would be better.

WRansohoff (approved):
This looks great! Our setup/install steps haven't been keeping pace with OpenROAD's improved setup support, and using CMake to run the builds from the github source is a nice improvement.

Thanks for addressing some tech debt in our wheels CI action, too!

PR #1208, opened 2023-01-15
core/apps: add iomap to _metadata to unify checking of file extension…

gadfort:
…s to filetype and fileset, reenable in sc-show and sc, and add infer to input/output core functions

Fixes cmdline apps to place -inputs into filesets and types
Adds a single place for all default filetypes and sets in _metadata.py (We can move it somewhere else if that works better).


gadfort:
@aolofsson I moved the code into utils and reset _metadata as per our discussion

aolofsson (approved):
Awesome!

PR #1209, opened 2023-01-16
Splitting tools schema into tools+tasks

aolofsson:
Since the beginning of this project we have struggled some with creating setup scripts for tools that have multiple use scenarios. Examples include Yoys (asic, fpga, lec), Openroad (place, floorplan, route, ...), Verilator (lint, import, compile), Klayout (export, merge, show). The more use cases were added,  the uglier the code would get.  Many times, we resorted to using the informal flowgraph driven 'step' name to choose large branches in the code to execute before handing off a manifest to  the executable's native API (usually TCL).

Only a small subset of parameters are really 'tool specific' and shared in all possible tool tasks (eg. exe, version, path, vswitch, format).  Everything else (entry scripts, options, vars,...) is often unique to a task and should be broken out into separate dictionaries.

This PR adds a named task as a key into the tool configuration to improve modularity and reduce "namespace" collision in the setup.

**Old Keypath:**
['tool', 'yosys','var',...]

**New Format:**

['tool', 'yosys', 'task', 'synasic', 'var',...]
['tool', 'yosys', 'task', 'synfpga, 'var',...]
['tool', 'yosys', 'task', 'lec, 'var',...]

This change is proposed in coordination with the modularity PR # 1205 and pending step/index change.


examples/heartbeat_caravel_wrapper/tools/addvias/addvias.py
```diff
@@ -7,12 +7,14 @@ def make_docs():
 
 def setup(chip):
     tool = 'addvias'
-    design = chip.get('design')
+    design = chip.get('design')x
```
gadfort, line 10: remove `x` (although this test clearly isn't in the CI)

examples/heartbeat_caravel_wrapper/tools/addvias/addvias.py
```diff
@@ -7,12 +7,14 @@ def make_docs():
...
     step = chip.get('arg', 'step')
     index = chip.get('arg', 'index')
+    #TODO fix
+    task = step
```
gadfort, line 14: To be fixed with a different PR which adds the importing, I assume

siliconcompiler/core.py
```diff
@@ -1533,12 +1541,14 @@ def check_manifest(self):
 
         #4. Check per tool parameter requirements (when tool exists)
         for step in steplist:
+
             for index in self.getkeys('flowgraph', flow, step):
                 tool = self.get('flowgraph', flow, step, index, 'tool')
+                tool = self.get('flowgraph', flow, step, index, 'task')
```
gadfort, line 1547: task? Since this is also validation, we should add checks and raise error if tool and task are not set

siliconcompiler/core.py
```diff
@@ -1533,12 +1541,14 @@ def check_manifest(self):
...
                 if (tool not in self.builtin) and (tool in self.getkeys('tool')):
```
gadfort, line 1548: Should the tool be "builtin" for the builtin functions? And then the task is "min"?

siliconcompiler/core.py
```diff
@@ -2987,15 +3004,14 @@ def clock(self, pin, period, jitter=0):
         self.set('datasheet', design, 'pin', pin, 'tjitter', 'global', jitter_range)
 
     ###########################################################################
-    def node(self, flow, step, tool, index=0):
+    def node(self, flow, step, tool, task=None, index=0):
```
gadfort, line 3007: Does it make sense to have task==None, it seems like this is a required parameter

siliconcompiler/core.py
```diff
@@ -3015,6 +3032,14 @@ def node(self, flow, step, tool, index=0):
 
         # bind tool to node
         self.set('flowgraph', flow, step, str(index), 'tool', tool)
+
+        tasks = self.getkeys('tool', tool, 'task')
+
+        if task is None:
+            task = step
```
gadfort, line 3039: Not a huge fan of blindly setting this, if task is a requirement, then this is not needed. I think you could go the other way since step is just a name and can be anything, so 
```
if step is None:
  step = task
```

siliconcompiler/core.py
```diff
@@ -4407,23 +4436,25 @@ def _getexe(self, tool):
...
+    def _makecmd(self, tool, task, step, index, extra_options=None):
         '''
         Constructs a subprocess run command based on eda tool setup.
         Creates a replay script in current directory.
         '''
 
         fullexe = self._getexe(tool)
 
+        self.write_manifest(f"{step}.dump.tcl")
+
```
gadfort, line 4448: Remove

siliconcompiler/flows/_common.py
```diff
@@ -6,19 +6,19 @@ def setup_frontend(chip):
     frontend_flow = []
 
     if frontend in ('verilog', 'systemverilog'):
-        frontend_flow.append(('import', 'surelog'))
+        frontend_flow.append(('import', 'surelog', 'import'))
```
gadfort, line 9: This is fine, but it might be easier to read as a dict, I'm not a fan of tuples especially since indexing and changes to the format will be hard to track:
```
.append({"step": "import", 
         "tool": "surelog",
         "task": "import")

File: siliconcompiler/flows/asicflow.py [hunk 9]
```diff
@@ -82,21 +82,24 @@ def setup(chip, flowname='asicflow'):
...
+        'physyn' : ['openroad','physyn'],
+        'physynmin' : ['minimum','physynmin'],
+        'place' : ['openroad','place'],
+        'placemin' : ['minimum','placemin'],
+        'cts' : ['openroad','cts'],
+        'ctsmin' : ['minimum','ctsmin'],
+        'route' : ['openroad','route'],
+        'routemin' : ['minimum','routemin'],
+        'dfm' : ['openroad','dfm'],
+        'export' : ['klayout', 'export']
```
gadfort, line 102: Same as above

siliconcompiler/flows/asicflow.py
```diff
@@ -122,18 +125,19 @@ def setup(chip, flowname='asicflow'):
...
         fanout = 1
         if param in chip.getkeys('arg', 'flow'):
             fanout = int(chip.get('arg', 'flow', param)[0])
         # create nodes
         for index in range(fanout):
             # nodes
-            chip.node(flowname, step, tool, index=index)
+            chip.node(flowname, step, tool, task, index=index)
             # edges
             if re.search(r'join|maximum|minimum|verify', tool):
```
gadfort, line 142: task?

siliconcompiler/targets/freepdk45_demo.py
```diff
@@ -32,9 +32,9 @@ def setup(chip):
 
     #2. Load PDK, flow, libs combo
     chip.load_pdk('freepdk45')
-    chip.load_flow('lintflow')
+    #chip.load_flow('lintflow')
```
gadfort, line 35: why?

siliconcompiler/targets/freepdk45_demo.py
```diff
@@ -32,9 +32,9 @@ def setup(chip):
...
     chip.load_flow('asicflow')
-    chip.load_flow('asictopflow')
+    #chip.load_flow('asictopflow')
```
gadfort, line 37: why?

siliconcompiler/tools/surelog/surelog.py
```diff
@@ -84,12 +86,12 @@ def setup(chip):
...
-    chip.set('tool', tool, 'regex', step, index, 'errors', r'^\[(ERR|FTL|SNT):', clobber=False)
+    chip.set('tool', tool, 'task', task, 'regex', step, index, 'warnings', r'^\[WRN:', clobber=False)
+    chip.set('tool', tool, 'task', task, 'regex', step, index, 'errors', r'^\[(ERR|FTL|SNT):', clobber=False)
 
-    warnings_off = chip.get('tool', tool, 'warningoff')
-    for warning in warnings_off:
-        chip.add('tool', tool, 'regex', step, index, 'warnings', f'-v {warning}')
+    #warnings_off = chip.get('tool', tool, 'warningoff')
+    #for warning in warnings_off:
+    #    chip.add('tool', tool, 'regex', step, index, 'warnings', f'-v {warning}')
```
gadfort, line 94: why?

gadfort (changes requested):

examples/heartbeat_caravel_wrapper/tools/addvias/addvias.py
aolofsson: Clearly, will fix:-)

siliconcompiler/flows/_common.py
aolofsson: I agree, will leave it for now. We should discuss replacements for the frontend.

siliconcompiler/flows/asicflow.py
aolofsson: The node() method was changed to add an optional task argument, letting us handle the new case where multiple steps exist for one task. The default behavior is for step==task.

siliconcompiler/targets/freepdk45_demo.py
aolofsson: Thank you! Restored, it was a temporary hack that got checked in by mistake.  Looks like another missing test.

siliconcompiler/flows/asicflow.py
gadfort: I was actually referring to line 142, but that would be in conjunction with my comments about the builtin functions changing their tool name and taking a task.

examples/heartbeat_caravel_wrapper/tools/addvias/addvias.py
nmoroze: It's a daily test so would have broke those :)

siliconcompiler/tools/verilator/verilator.py
```diff
@@ -76,74 +76,55 @@ def setup(chip):
...
-            chip.add('tool', tool, 'option', step, index, '-LDFLAGS')
-            chip.add('tool', tool, 'option', step, index, ldflags_str)
+        # Make warnings non-fatal in relaxed mode
+        if chip.get('option', 'relax'):
+            chip.add('tool', tool, 'task', task, 'option', step, index, ['-Wno-fatal', '-Wno-UNOPTFLAT'])
 
-        # File-based arguments added in runtime_options()
+        # Converting user setting to verilator specific filter
+        #for warning in chip.get('tool', tool, 'task', task, step, index, 'warningoff'):
+        #    chip.add('tool', tool, 'task', task, 'option', step, index, f'-Wno-{warning}')
```
nmoroze, line 108: Should this be restored?

siliconcompiler/core.py
aolofsson: Sounds good, do you think you can fix the tests/flows that call (). They will break if we make task required.

nmoroze (approved):
Looks good once the fixes requested by Peter are done. I commented on a few other small things inline. 

One concern I have is that adding this will blunt the benefit of our step/index abstraction a bit. As one specific example, I like the idea of being able to set a global thread count easily:

```python
chip.set('tool', 'openroad', 'threads', 2)
```

but now we'll still have something like:

```python
# MUST COME AFTER chip.load_flow() !!

for step in chip.getkeys('flowgraph', flow):
    for index in chip.getkeys('flowgraph', 'flow', step):
        task = chip.get('flowgraph', flow, step, index, 'task')
        chip.set('tool', 'openroad', 'task', task, 'threads', 2)
```

I'm not sure how best to fix this. We could consider moving the task into the schema as well, but this feels like something that's a bit more tool-specific. I also wonder if it would make sense to restrict the additional "task" tag to things that really are often task specific, like script/refdir, input/output, and CLI options, and not add the extra level of indenting to things like threads, keep, continue, etc. Perhaps we can discuss a bit more at our next meeting?

siliconcompiler/flows/asicflow.py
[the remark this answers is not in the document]
gadfort: @aolofsson suggested doing the conversion of builtins in a separate PR

gadfort:
@nmoroze you bring up a good point about the blunting the benefits of the step/index, lets discuss tomorrow.

gadfort (approved):
@aolofsson `.node` now requires the task and is passing the initial CI

PR #1210, opened 2023-01-16
Simplify show/screenshots as steps

gadfort:
Changes:
- Makes show and screenshot a step name in OpenROAD and klayout (this will also allow us to build show steps for other tools as where it might be useful to have)
- Removes the need for special running code for tools (the tool drivers handle this themselves), this required the use of `show_exit` as a variable to indicate if it needs to close, which the tests can use to for closing always.
- OpenROAD copies in SDC file from source to to ensure timing is persevered when viewing
- `def show` now builds a flow of its own to use so it ends up behaving like other tool interactions.

Thoughts:
- Some of the duplicated code can be handled better once imports are supported.
- Once tasks are available, this will need to be fixed, but I'm happy to do that.


siliconcompiler/core.py
```diff
@@ -4269,66 +4252,78 @@ def show(self, filename=None, extra_options=None):
...
+        saved_config = self.schema.copy()
+
+        filetype = utils.get_file_ext(filepath)
+
+        if filetype not in self.getkeys('option', 'showtool'):
             self.logger.error(f"Filetype '{filetype}' not set up in 'showtool' parameter.")
+            return False
+
+        # Setup new show/screenshot flow
+        flow = 'showflow'
```
aolofsson, line 4273: Can this be a flow inside the "flows" directory?

gadfort: Yes, I originally had it that way, but switched it back. I'll make it a loadable flow

siliconcompiler/tools/klayout/klayout.py
```diff
@@ -78,7 +78,9 @@ def setup(chip, mode="batch"):
     else:
         klayout_exe = 'klayout'
 
-    if mode == 'show':
+    is_show = mode == 'show' or step == 'show'
```
aolofsson, line 81: This is an example of a place where having separate task definitions for klayout would avid confusion. Otherwise it looks fine.

siliconcompiler/tools/openroad/openroad.py
```diff
@@ -99,6 +103,20 @@ def setup(chip, mode='batch'):
         chip.add('tool', tool, 'output', step, index, design + '.def')
         chip.add('tool', tool, 'output', step, index, design + '.odb')
         chip.add('tool', tool, 'output', step, index, design + '.lef')
+    elif is_show_screenshot:
```
aolofsson, line 106: Not looking forward to merging this with my current PR draft:-)

aolofsson (approved):
Looks good

siliconcompiler/tools/klayout/klayout.py
gadfort: yes, that will simplify this a lot, there are some confusing if trees now.

siliconcompiler/tools/openroad/openroad.py
gadfort: If the merge is too onerous, I can take a stab at it, just let me know

PR #1211, opened 2023-01-16
Rudimentary upgrades to Yosys+VPR FPGA flow

petergrossmann21:
Here I have added hooks to call nearly all of the different synth_<partname> commands in yosys instead of just ice40, and have added a complete(ish) reference synthesis flow for use with VPR.  Also added some small additions to the VPR invocation setup.  Note that some upgrades to CI testing may be warranted to test this properly.


gadfort:
@petergrossmann21 making this a draft to make this easier to track

PR #1212, opened 2023-01-16
More Vivado/FPGA enhancements

nmoroze:
This PR contains another grab bag of tweaks related to Vivado/FPGA flows.

- Make it optional to supply an XDC file to the Vivado flow
  - Suggested by @petergrossmann21, this will make it easy to stand up quick tests to estimate the area of a design.
  - Verified working by manually removing constraint from Blinky example and running
- Add FPGA constraint files to the I/O map so we can supply them directly to `sc`
- Add warnings/errors regexes to Vivado driver

### DIFF
--- a/examples/blinky/run.sh
+++ b/examples/blinky/run.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
-sc -input "rtl verilog blinky.v" \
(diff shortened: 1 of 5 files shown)


aolofsson (approved):
Looks great!

PR #1213, opened 2023-01-17
Update GH Actions workflow dependencies

nmoroze:
A lot of our Actions dependencies were pinned to old versions that use deprecated features. Updating them should silence the warnings we see in our workflow summary pages.

### DIFF
--- a/.github/workflows/bin/install_klayout_win.bat
+++ b/.github/workflows/bin/install_klayout_win.bat
@@ -1,3 +1,3 @@
-curl https://www.klayout.org/downloads/Windows/klayout-0.27.5-win64.zip --output klayout.zip
+curl https://www.klayout.org/downloads/Windows/klayout-0.28.3-win64.zip --output klayout.zip
 7z x klayout.zip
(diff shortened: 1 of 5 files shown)


nmoroze:
Should probably hold off on merge till this goes green: https://github.com/siliconcompiler/siliconcompiler/actions/runs/3943957889.

I'm not too worried about trying daily tests before merging, that's a subset of the quick test changes and it will bog down the test runner for a while.

gadfort:
@nmoroze I spoke with @aolofsson and I think it will be easiest to fix the wheels once everything else is done, it's failing on some tests that require openroad, so to avoid repeating the same fixes over and over again, I suggested leaving that one red (daily CI and push CI should still stay green). And once the bulk of the changes are one we can fix the wheels.
Let me know if you think we should do otherwise.

nmoroze:
Whoops, I just went in and fixed it before noticing your reply. I think if this works we might as well merge, but otherwise I'll leave them alone until after the dust settles.

gadfort (approved):
Lgtm

PR #1215, opened 2023-01-18
Remove direct dictionary access of schema values in core

nmoroze:
This PR removes the last few direct accesses of schema values from core.py. 

In order to facilitate this, I added some new functionality to the `Schema` class:
- A class method for instantiating a `Schema` object from a manifest file, which can be shared by `read_manifest()` and `_find_deps()`
- A method for getting a `Schema` object that provides a mutable reference to a history dictionary, for use in `_merge_manifest()`

I think there are a few other ways to tackle this, including moving the `read_manifest()` and `_merge_manifest()` methods into `Schema` itself. However, one minor annoyance is that the `Schema` object doesn't have access to the logger, so we'd have to remove a few warnings. Keeping the methods in the same place also gives a diff that's easier to review, and I made some changes to `_merge_manifest()` to fix a bug.

This PR also makes a few other semi-related changes:
- Removes `merge_manifest()` from the core public API
  - This might have wanted to be its own PR, but I included it here since I had to touch it anyways to account for the `_merge_manifest()` API change. As far as I can tell we're not using this method in any of our own user code, and I don't anticipate other users really needing this - all user code I've seen just uses `read_manifest()`.
- Fixes a bug I introduced in #1147 that breaks the ability to read manifests into a particular job.


nmoroze:
I snuck a few more changes in to refactor `graph()` - it includes a change to the `getkeys()` function to replace its behavior with no args with a new `allkeys()` function, which can take in a keypath prefix to restrict the return values.

siliconcompiler/core.py
```diff
@@ -1235,29 +1243,6 @@ def _abspath(self):
         return schema
 
     ###########################################################################
-    def merge_manifest(self, cfg, job=None, clobber=True, clear=True, check=False):
```
gadfort, line 1238: no merge_manifest in this class?

gadfort (changes requested):
Overall looks good, just wondering about the case sensitivity of those checks.

siliconcompiler/core.py
```diff
@@ -310,7 +310,7 @@ def create_cmdline(self, progname, description=None, switchlist=None, input_map=
                                          description=description)
 
         # Get all keys from global dictionary or override at command line
-        allkeys = self.getkeys()
+        allkeys = self.allkeys()
```
aolofsson, line 313: Is the intent to make allkeys() a public method now and if so, does that completely replace the getkeys() behavior (or do we now have two ways of getting all the keys back?)

siliconcompiler/core.py
```diff
@@ -3091,31 +3055,27 @@ def graph(self, flow, subflow, name=None):
...
-            self.schema.cfg['flowgraph'][flow] ={}
-
-        # uniquify each step
-        for step in self.getkeys('flowgraph',subflow):
+        for step in self.getkeys('flowgraph', subflow):
+            # uniquify each step
             if name is None:
                 newstep = step
             else:
                 newstep = name + "." + step
```
aolofsson, line 3063: This is good!

siliconcompiler/schema/schema_obj.py
```diff
@@ -31,6 +32,40 @@ def __init__(self, cfg=None):
         else:
             self.cfg = copy.deepcopy(cfg)
 
+    ###########################################################################
+    @classmethod
+    def from_manifest(cls, filepath):
```
aolofsson, line 37: Shouldn't this replace read_manifest with the new schema object approach?

siliconcompiler/schema/schema_obj.py
```diff
@@ -591,6 +638,26 @@ def _keypath_empty(self, key):
 
         return value_empty
 
+    ###########################################################################
+    def get_history(self, job):
```
aolofsson, line 642: Is this meant to be a public function. Can we rename to 'history()' or 'getschema()'  We currently have {get, getkeys,  getdict}

aolofsson (changes requested):

nmoroze:
Thanks for the feedback! I renamed `get_history()` and made the manifest filepath checks case-insensitive.

I should have been a bit clearer summarizing the public API changes, I'll do that here (with rationale):

## `Chip`

**Remove `merge_manifest()`**
  - This method is a wrapper over a private method `_merge_manifest()`. The private method will stay.
  - We don't have any examples or user code that I can find that uses `merge_manifest()`. The private method is used to implement `read_manifest()`, which is what all the user code calls.
  - Happy to keep it if we can see a use for it, but I'd probably change the API slightly to take in a `Schema` object rather than a dictionary.

## `Schema`

**Add `from_manifest()`**
  - This is a class method (think alternate constructor) that reads in a manifest and returns a schema object. It's a generalization of code that was repeated in 2 places: `read_manifest()` and `_find_deps()`. `read_manifest()` performs additional merging logic, so we want to keep that around. I do think it would be nice to add a `from_manifest()` method to the `Chip` class as well in a future PR.

**Add `history()`**
  - This is a getter that allows us to manipulate the history of a `Schema` object as a `Schema` object itself. It was necessary for cleaning up the low-level schema access in `read_manifest()` and `_merge_manifest()`. I'm not super happy with since it feels like a weird special-case, but I think in the future we may be able to eliminate it if we reimplement the history stuff to be nested instances of `Schema` objects.

## `Chip`/`Schema`

**Change `getkeys()` to eliminate special-case behavior**

- `getkeys()` with an argument returns a list of individual keys, but without an argument it returns a list of lists of all keypaths
- I think there are cases where it's useful to iterate over the top-level keypaths, so I think it would be best to make its behavior more consistent

```python
>>> chip.getkeys('option') # same behavior before and after change...
['remote', 'credentials', 'jobscheduler', 'mode', ...] 
>>> chip.getkeys() # behavior BEFORE change
[['schemaversion'], ['design'], ['input', 'default', 'default'], ['output', 'default', 'default'], ...] 
>>> chip.getkeys() # behavior AFTER change
['history', 'library', 'schemaversion', 'design', ...]
```

**Add `allkeys()`**

  - This already existed as a private method, and replaces the old `getkeys()` special-case behavior
  - This also provides additional functionality by accepting a keypath argument that specifies a prefix for what's returned. Currently the returned values do not include the prefix (which made some other code I'm working on simpler), but we can tweak that behavior
  
```python
>>> chip.allkeys()
[['schemaversion'], ['design'], ['input', 'default', 'default'], ['output', 'default', 'default'], ...]
>>> chip.allkeys('flowgraph', 'default', 'default', 'default')
[['input'], ['weight', 'default'], ['goal', 'default'], ['tool'], ['task'], ['args'], ['valid'], ['timeout'], ['status'], ['select']]
```

siliconcompiler/core.py
nmoroze: Nope, two ways that do different things (see new comment summarizing methods)

siliconcompiler/schema/schema_obj.py
nmoroze: Not exactly the same, `read_manifest()` has additional logic for merging. Just added a comment in the main thread providing more detail.

aolofsson:
Thanks for the justifications and updates!
- Is 'from_manifest()' meant to be a public method? I am ok with it as long as we agree to do an API review as well in the next few weeks.
- I find the new allkeys() / getkeys() behavior bit confusing for a public facing api. They do similar things but with different slightly different behavior. What is the actual motivation/use case for the user facting change?

aolofsson:
Thanks for breaking down allkeys/getkeys offline.With changes to constructor, I am happy.😁

nmoroze:
Sounds good, I made the constructor changes!

aolofsson (approved):

aolofsson:
@gadfort I think there are quite a few places where we only check for lower case file suffixes. Can we defer that to a separate holistic PR and merge this one?

gadfort:
> @gadfort I think there are quite a few places where we only check for lower case file suffixes. Can we defer that to a separate holistic PR and merge this one?

I think we can defer and merge this one. I'll make a note of it.

gadfort (approved):

PR #1216, opened 2023-01-19
Adding footprint parameter to datasheet schema

aolofsson:
- A footprint is a a must have parameter for all physical designs and it is a datasheet level parameter. (ie would be in the PDF).  https://www.jedec.org/category/technology-focus-area/jc-10/registered-outlines-jep95

 - The parameter is currently embedded in the asic group which is too limited.
 - This should be able to serve the need of the 'asic' 'footprint' as well...


gadfort (approved):
I don't know about changing the asic footprint. If anything I would want to rename the asic one to site and make is clear that that is what it's for. I think this should be used in reference to the wirebond or bump patterns (naming that is). We can discuss.

gadfort:
> I think you are right, as long as we agree that the asic "site" is very different from anything one might see in package, board. To deconflict and to be consistent with lef/def, we should probable rename asic footprint to asic site.

That was discussed at today meeting

PR #1217, opened 2023-01-20
ensure environmental variables set in run are cleared and returned to normal user environment

gadfort:


nmoroze (approved):
LGTM. 

I recall that you proposed passing in the SC environment variables directly to subprocess a while back. I think my only concern with that would be if a tool's `pre_process()` or `post_process()` relied on them, but otherwise that feels a bit more elegant to me. In any case though, this is a safety net that makes sure we don't mess up.

gadfort:
> LGTM.
> 
> I recall that you proposed passing in the SC environment variables directly to subprocess a while back. I think my only concern with that would be if a tool's `pre_process()` or `post_process()` relied on them, but otherwise that feels a bit more elegant to me. In any case though, this is a safety net that makes sure we don't mess up.

Turns out the subprocess spawns a new environment anyways so there is no need to clean up the work in _runtask since they that is already isolated.

PR #1219, opened 2023-01-21
Fix daily CI tests

WRansohoff:
They take a bit longer than 20 minutes now, and it looks like the Verilator `compile` runtime options need to include input Verilog files.


gadfort (approved):
lgtm

Issue #1188, opened 2023-01-04
Allow user to set empty values


nmoroze:
Calls of these sort should be considered legal and not trigger an exception:

```python
chip.set(..., None) # scalars
chip.set(..., []) # lists
```

Issue #1085, opened 2022-07-15
Make error handling more useful and Pythonic


nmoroze:
Based on some user feedback, I've been thinking about how to improve our SC error handling.

I think we should try to adhere to the following principles:

- Use built-in [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError) and [`TypeError`](https://docs.python.org/3/library/exceptions.html#TypeError) where appropriate
    - We solely use `SiliconCompilerError` for exceptions, but it's considered good practice to use these built-in types when appropriate, since they have standardized meetings 
- Raise errors close to public API waterline
  - The main piece of feedback that inspired this issue is that SC tracebacks can be confusing 
  - Both @WRansohoff and I have looked into this, and it seems to be considered bad practice to attempt to modify tracebacks, even for readability
  - The best-practice approach seems to be to be thoughtful about where exceptions are raised, and in some cases catch exceptions that occur deep in the implementation and raise a fresh one in the top-level public API call
  - Note: our `self.error()` approach works against us here, since it adds a layer to the traceback. I'm not sure if it's worth doing anything at this point **
- Gracefully catch any errors that may arise
  - PR #1084 is a useful push in this direction
- Document when an API method may raise an exception 

Some proposed changes based on these principles:

- [ ] Make chip.error() take in an optional exception type to throw as a kwarg
   - This will allow for raising TypeError and ValueError
- [ ] Make set()/get()/add() errors cleaner
  - These errors are common since it's easy to make a mistake in input, and the extra layer of _search() clutters the traceback a bit
  - _search() can be made to just raise a custom exception type directly (a custom type is better to ensure we don't accidentally swallow other errors)
  - set()/get()/add() can catch the internal exceptions and raise a ValueError or TypeError
- [ ] Audit other chip.error() calls to make sure they throw the correct exception type
- [ ] Document which public API calls may raise a SiliconCompilerError and when

I think we can keep this issue open until we have a chance to discuss and implement these proposals. After that, we can close the issue but we should keep these principles in mind going forward, and open issues if we notice existing code that violates them. 

** A somewhat radical idea we could do to fix the chip.error() in the traceback could be to throw exceptions directly in the code and then put a decorator on public API methods, but this might not be that worthwhile: 

```python
def exc_decorator(func):
    def wrap(self, *args, **kwargs):
        try:
            return func(self, *args, **kwargs)
        except SCOptionalTypeError as e:
            msg = str(e)
[...]

PR #1222, opened 2023-01-21
schema_obj: refactor _search()

### LINKED ISSUE #1146: Make clobber interact appropriately with default values

nmoroze:
This PR refactors our monolithic `_search()` function to make the code clearer, fix some latent bugs, improve error messages, and, most relevantly, facilitate adding special handling for global values with step/index overrides.

This PR is meant to be (mostly) a clean refactor with no change to user-facing functionality. However, there are some  differences in behavior described below. 

First, a couple key design decisions made in this PR:
- The underlying `cfg` dictionary now stores native Python types, rather than just strings. As far as I can tell, there's no reason to store values as strings -- all the Python types used are serializable to/from JSON (with the exception of tuples, which get serialized as lists). Using native Python types makes the code much clearer.
- `_search()` is no longer a monolithic function. I kept the name around, but the new `_search()` is a simple traversal function. The actual `set()`/`get()`/`add()`, etc. functionality is now written in those functions. They make use of a variety of new internal helper functions for shared functionality. I think this approach is much clearer.

### `get()` no longer creates keypaths

I noticed that with the old `_search()`, `get()`-ing a keypath with unset free keys actually adds keys to the cfg dictionary:
```python
>>> chip.getkeys('tool', 'surelog', 'task')
[] # (Surelog not set up yet)
>>> chip.get('tool', 'surelog', 'task', 'import', 'stdout', 'import', '0', 'suffix')
'log'
>>> chip.getkeys('tool', 'surelog', 'task')
['import']
```

It seemed a little odd to me that `get()` could have this side effect. I've changed the behavior in this PR to allow it to keep descending through the defaults (so you can recover a default value), but the dictionary doesn't actually get created. New behavior:

```python
>>> chip.getkeys('tool', 'surelog', 'task')
[]
>>> chip.get('tool', 'surelog', 'task', 'import', 'stdout', 'import', '0', 'suffix')
'log'
>>> chip.getkeys()
[]
```

### Can no longer set lists to `None`

I think `set()` previously allowed you to provide `None` for a list value. This PR changes that behavior to cause it to fail the type check, since we've generally been using `[]` as the equivalent null value for list types. However, I'd be open to changing this behavior to accept `None` for list types, and have it get normalized as `[]` in the schema.

### Add 'set' flag for clobber

This PR adds another flag, 'set', to parameters to indicate whether their value has been set by a user yet. Previously, clobber operated based on whether the value was of an "empty" type (generally equivalent to False-y), but this caused unintuitive behavior if a value was explicitly set to an "empty" value (#1146). Adding the "set" flag allows us to determine whether a value has actually been exlicitly set by user code before

Closes #1146, #1188


gadfort (changes requested):
Changes look good, I like the simplification of the type checking. If we keep `set` you will need to update the schema version.

siliconcompiler/schema/schema_obj.py
```diff
@@ -73,12 +73,20 @@ def get(self, *keypath, field='value', job=None):
...
         """
-        if job is None:
-            cfg = self.cfg
-        else:
-            cfg = self.cfg['history'][job]
+        cfg = self._search(*keypath, job=job)
 
-        return self._search(cfg, str(keypath), *keypath, field=field, mode='get')
+        if not Schema._is_leaf(cfg):
+            raise ValueError(f'Invalid keypath {keypath}: get() must be called on a leaf')
```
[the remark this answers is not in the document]
nmoroze: Maybe "complete keypath"? (or "full keypath")

siliconcompiler/core.py
```diff
@@ -3967,10 +3981,19 @@ def run(self):
                     # we're not running with -resume, we also re-run anything
                     # in the steplist.
                     self.set('flowgraph', flow, step, index, 'status', None)
+
+                    # Reset metrics and records
+                    # We need to clear the 'set' flag to allow new metric values
+                    # to be merged in (since we merge manifests from each node
+                    # with clobber=False).
+                    # TODO: should there be a more formal way to do this? Or
+                    # should writing None do it implicitly?
```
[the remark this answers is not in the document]
nmoroze: This is solving a different issue (which is that the `chip.set(...)` here will prevent the record from being set to the correct value later). However, if we change things so that setting a value to `None` effectively un-sets it like you mentioned in other comments, then this would be simplified.

siliconcompiler/schema/schema_obj.py
```diff
@@ -194,204 +229,162 @@ def record_history(self):
...
+                value = [value]
+            base_type = sc_type[1:-1]
+            return [Schema._normalize_value(v, base_type) for v in value]
+
+        if sc_type.startswith('('):
+            # TODO: make parsing more robust to support tuples-of-tuples
+            if isinstance(value, str):
+                value = value[1:-1].split(',')
+            elif not (isinstance(value, tuple) or isinstance(value, list)):
+                raise TypeError
```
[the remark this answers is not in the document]
nmoroze: These `TypeErrors` are caught by the user of this function and re-raised with a message (I noticed I was copy-pasting the same message for each). I'm happy to add messages back to this method to make it more explicit what's going on, though.

siliconcompiler/schema/schema_obj.py
```diff
@@ -194,204 +229,162 @@ def record_history(self):
...
+
+        raise ValueError(f'Invalid field: {field}')
+
+    @staticmethod
+    def _is_empty(cfg):
+        return not cfg['set']
+
+    @staticmethod
+    def _is_leaf(cfg):
+        return 'shorthelp' in cfg and isinstance(cfg['shorthelp'], str)
```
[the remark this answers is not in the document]
nmoroze: Honestly it's somewhat arbitrary (anything that is guaranteed to be there and have a consistent type works). I'll add a comment

nmoroze:
Thanks for the review! I'll fix up the various messages per your suggestions. 

RE: `cfg['set']`, I was originally envisioning explicitly distinguishing when the user has chosen to set the value to `None`, and making that require clobber to overwrite: 

```python
>>> chip.set('metric', 'syn', '0', 'errors', None)
>>> chip.set('metric', 'syn', '0', 'errors', 0, clobber=False)
# Fails!
```

However, in retrospect this is probably counterintuitive. I'd be open to making `None` a sentinel for an un-set value (and allow a user to use it to reset a value's status), and then we don't need to introduce `cfg['set']`.

I'm also thinking it might be reasonable to fall back to the default value in the case when a user sets `None`, rather than returning `None` itself:

```python
>>> chip.get('option', 'quiet')
False
>>> chip.set('option', 'quiet', True)
>>> chip.set('option', 'quiet', None)
False # (not None!)
```

siliconcompiler/schema/schema_cfg.py
```diff
@@ -18,8 +18,8 @@ def scparam(cfg,
             require=None,
             defvalue=None,
             scope='job',
```
aolofsson, line 20: The switch to actual types is a great change!! The initial decision to make everything a string was made pre Python API when I had approximately zero years of Python experience so I wanted to keep things uniform:-)

There is at least one schema change (aspect ratio) with defvalue as string that should also be fixed.

siliconcompiler/schema/schema_obj.py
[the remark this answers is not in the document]
aolofsson: I assume by user, you mean developer?  I thought a leaf cell/function was generally understood as the last stop on the tree. (ie leaf).

siliconcompiler/schema/schema_obj.py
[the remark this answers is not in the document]
aolofsson: Can we support our standard logger messages for this? What does the TypeError report back?  Some of the error messages are not exactly user friendly. We should be able to help the user quickly figure out what they did wrong with the right fields (value, sc_type, keypath?

siliconcompiler/schema/schema_obj.py
```diff
@@ -194,204 +229,162 @@ def record_history(self):
...
+
+    @staticmethod
+    def _is_list(field, type):
+        is_list = type.startswith('[')
+
+        if field in ('filehash', 'date', 'author', 'example'):
```
aolofsson, line 355: Having these fields hard coded deep in the core/search function was always a bit annoying when we made a change in the schema to add a leaf field, but given how seldom we do that, I think this is fine.

siliconcompiler/schema/schema_obj.py
```diff
@@ -194,204 +229,162 @@ def record_history(self):
...
-                #all non-value fields are strings (or lists of strings)
+            if Schema._is_leaf(cfg):
+                raise ValueError(f'Invalid keypath {keypath}: unexpected key: {key}')
+
+            if key in cfg:
+                cfg = cfg[key]
+            elif 'default' in cfg:
+                if insert_defaults:
+                    cfg[key] = copy.deepcopy(cfg['default'])
+                    cfg = cfg[key]
```
aolofsson, line 381: A beautiful example of replacing technical debt (mine) with a much cleaner and better implementation. Thank you!!

aolofsson (approved):

aolofsson:
Not sure I fully grasp the set vs sentinel vs clobber tradeoffs here, especially as it relates to bool. Setting a parameter that we have specified as type bool to None per your example seems confusing to me. Same thing goes for lists. Is the idea to intercept a special "None" value entered through set and take some action. How do we distinguish from someone who actually wants to set a value to None? 

```python
print(type(None))
```
<class 'NoneType'>

siliconcompiler/schema/schema_obj.py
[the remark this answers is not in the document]
gadfort: I mean user, these errors propagate up to the user when they make a mistake, the error messages should make sense to them and developers.
@nmoroze "complete keypath" is good and consistent with the other messages throughout.

siliconcompiler/schema/schema_obj.py
```diff
@@ -194,204 +229,162 @@ def record_history(self):
...
+                raise TypeError
+
+            base_types = sc_type[1:-1].split(',')
+            if len(value) != len(base_types):
+                raise TypeError
+            return tuple(Schema._normalize_value(v, base_type) for v, base_type in zip(value, base_types))
+
+        if sc_type == 'bool':
+            if value == 'true': return True
+            if value == 'false': return False
```
WRansohoff, line 295: `if value in ('true', 'True')`? Ditto for false?

WRansohoff (approved):
Nice simplification! As a general comment, should we consider using `SiliconCompilerError` in any of the places where `TypeError` / `ValueError` is used?

Pro: Easy to catch in build scripts.

Con: Could accidentally swallow errors, and the Schema object is sort of logically separate from runtime SC errors.

siliconcompiler/schema/schema_obj.py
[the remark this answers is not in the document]
nmoroze: We can't support our standard logger messages directly, since we're not in the Chip object. I've just opened an issue tracking this, so we can fix it eventually: #1231. 

However, note that the TypeErrors raised by these functions are caught and have their messages re-routed to the normal error logging mechanism inside the equivalent Chip object methods.

To make the behavior more explicit, I've added messages to all of these bare TypeErrors, using a variable to avoid repeating the message in multiple places.

siliconcompiler/schema/schema_obj.py
nmoroze: I'm thinking of keeping as-is to maintain parity with our current API, which only supports lower-case "true"/"false" for strings. However, I'd be open to changing this if others think it's a good idea.

nmoroze:
@WRansohoff good points RE: exceptions. I think tweaking how they work in a future PR could be reasonable. Some of my thinking about how things should work is captured here: https://github.com/siliconcompiler/siliconcompiler/issues/1085. 

I think ultimately we should have `set()`/`get()`/`add()` throw `TypeError`/`ValueError` where relevant, since those are standardized - but the way we're currently catching them and re-throwing `SiliconCompilerError`s in the `core.py` implementations of `set()/get()/add()` is definitely unideal, since that doesn't even make the standardized errors user-facing, and it can swallow the source of other errors.

nmoroze:
@gadfort I think I've addressed all the feedback! At a high level:

- Keep `cfg['set']`, add `clear()` method for unsetting values (was between `unset()` and `clear()`, I liked `clear()` because of the analogy to `list.clear()` in Python)
- Improved error messages. Pushed down the keypaths into the recursive checking functions to display them, added variables/helper functions for deduplicating error messages. I realized this is more clear than catching and re-raising the exception at a higher level.

siliconcompiler/core.py
```diff
@@ -928,10 +928,35 @@ def set(self, *args, field='value', clobber=True):
             self.logger.setLevel(value)
 
         try:
-            self.schema.set(*keypath, value, field=field, clobber=clobber)
+            if not self.schema.set(*keypath, value, field=field, clobber=clobber):
+                # TODO: this message should be pushed down into Schema.set()
+                # once we have a static logger.
+                if clobber:
+                    self.logger.debug(f'Failed to set value for {keypath}: '
```
gadfort, line 935: should this be a warning or error instead of debug?

siliconcompiler/core.py
```diff
@@ -928,10 +928,35 @@ def set(self, *args, field='value', clobber=True):
...
+                        'parameter is locked')
+                else:
+                    self.logger.debug(f'Failed to set value for {keypath}: '
```
gadfort, line 938: same

siliconcompiler/core.py
```diff
@@ -928,10 +928,35 @@ def set(self, *args, field='value', clobber=True):
...
+        default value. A call to ``set()`` with ``clobber=False`` will once
+        again be able to modify the value.
+
+        Args:
+            keypath (list): Parameter keypath to clear.
+        '''
+        self.logger.debug(f'Clearing {keypath}')
+
+        if not self.schema.clear(*keypath):
+            self.logger.debug(f'Failed to clear value for {keypath}: parameter is locked')
```
gadfort, line 958: same

siliconcompiler/core.py
```diff
@@ -956,10 +981,16 @@ def add(self, *args, field='value'):
...
+        keypath = args[:-1]
+        value = args[-1]
+        self.logger.debug(f'Appending value {value} to {keypath}')
 
         try:
-            self.schema.add(*args, field=field)
+            if not self.schema.add(*args, field=field):
+                # TODO: this message should be pushed down into Schema.add()
+                # once we have a static logger.
+                self.logger.debug(f'Failed to add value for {keypath}: '
```
gadfort, line 992: same

gadfort (approved):
just down to the debug vs. warning. If it's debug else where maybe we should leave it and make a note to look at them later.

PR #1223, opened 2023-01-22
Pruning ASIC section of schema

aolofsson:
- Removing maxcap, maxslew, maxfanout (duplication, belongs in  sdc)
- Making aspectratio defvalue to 1.0 (only used for auto generation, special cases always come with floorplan).
- Changing ignore to dontuse for cell list (better name)
- Removing load cell list (belongs in sdc)
- Moving asic,stackup to option, stackup (consistent with current PDK setting)
- Removing redundant asic,pdk value (was already in option)
- Moving lambda to pdk (natural grouping)


siliconcompiler/schema/schema_cfg.py
```diff
@@ -311,6 +311,19 @@ def schema_pdk(cfg, stackup='default'):
             optimization. Node examples include 180, 130, 90, 65, 45, 32, 22 14,
             10, 7, 5, 3.""")
 
+    scparam(cfg,['pdk', pdkname, 'lambda'],
```
gadfort, line 314: It feels like this is intended to be a design parameters, so should it be in `option,lambda` instead?

siliconcompiler/schema/schema_cfg.py
```diff
@@ -2112,6 +2116,27 @@ def schema_option(cfg):
             may load multiple flows and libraries.
             """)
 
+    scparam(cfg, ['option','pdk'],
+            sctype='str',
+            scope='job',
+            shorthelp="PDK target",
+            switch="-option_pdk <str>",
+            example=["cli: -option_pdk freepdk45",
```
gadfort, line 2124: other options don't have option prefix, should be `-pdk`

siliconcompiler/schema/schema_cfg.py
```diff
@@ -2112,6 +2116,27 @@ def schema_option(cfg):
...
+                     "api: chip.set('option','pdk','freepdk45')"],
+            schelp="""
+            Target PDK used during compilation.""")
+
+    scparam(cfg, ['option', 'stackup'],
+            sctype='str',
+            scope='job',
+            shorthelp="Stackup target",
+            switch="-option_stackup <str>",
+            example=["cli: -option_stackup 2MA4MB2MC",
```
gadfort, line 2134: same, should be `-stackup`

gadfort (changes requested):
overall looks good, just need to correct the cli switches and discuss the location of lambda

siliconcompiler/schema/schema_cfg.py
```diff
@@ -2034,6 +2035,9 @@ def schema_option(cfg):
     ''' Technology agnostic run time options
     '''
 
+
+
+
```
WRansohoff, line 2040: Nit: delete?

siliconcompiler/targets/asap7_demo.py
```diff
@@ -27,25 +27,20 @@ def setup(chip):
     #3. Select default flow/PDK
     chip.set('option', 'flow', 'asicflow')
     chip.set('option', 'pdk', 'asap7')
+    chip.set('option', 'stackup', '10M')
```
WRansohoff, line 30: Is `10M` right for asap7, or is it possible we accidentally copied it from freepdk45?

WRansohoff (approved):
LGTM, just a couple of minor comments.

nmoroze (approved):

siliconcompiler/schema/schema_cfg.py
aolofsson: NIce catch, fixed.

siliconcompiler/schema/schema_cfg.py
aolofsson: Fixed

gadfort (approved):

PR #1226, opened 2023-01-23
Split yosys synthesis task into syn_fpga and syn_asic

WRansohoff:
Like we discussed today. I left the mode-dependent setup logic in `lec` for the sake of our tests - we can decide how to partition that logic when we start to use it more.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1218,7 +1218,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
 
         Args:
             filetype (str): File extension (v, def, etc)
-            step (str): Task step name ('syn', 'place', etc)
+            step (str): Task step name ('syn_asic', 'place', etc)
             jobname (str): Jobid directory name
             index (str): Task index
 
@@ -1226,7 +1226,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
             Returns absolute path to file.
 
         Examples:
-            >>> manifest_filepath = chip.find_result('vg', 'syn')
+            >>> manifest_filepath = chip.find_result('vg', 'syn_asic')
            Returns the absolute path to the manifest.
         """
         if jobname is None:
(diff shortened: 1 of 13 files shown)


siliconcompiler/tools/vivado/compile.tcl
```diff
@@ -18,12 +18,14 @@ if {[dict exists $sc_cfg input fpga xdc]} {
...
 set sc_index      [dict get $sc_cfg arg index]
+set sc_flow       [dict get $sc_cfg option flow]
+set sc_task       [dict get $sc_cfg flowgraph $sc_flow $sc_step $sc_index task]
 
 ##############################
 # Flow control
 ##############################
 
-if {$sc_step == "syn"} {
+if {$sc_task == "syn_fpga"} {
```
gadfort, line 28: how do we get here? The task is compile, so we can't be checking for syn_fpga, is this check redundant?

gadfort (changes requested):
Could you double check the logic on compile.tcl in vivado? Otherwise looks good.

siliconcompiler/tools/vivado/compile.tcl
WRansohoff: Good question - I don't know why there is a check for the synthesis step in the Vivado `compile` script.

It looks like there is synthesis logic contained in the block, so I don't want to delete it if it might be used somewhere. @nmoroze ?

nmoroze: The Vivado flow is currently implemented as a single file that implements every step/task using a branch:

```tcl
# common setup...

if {$sc_step == "syn"} {
  # syn logic
 } elseif ($sc_step == "place"} {
  # place logic
} # etc...

# common finish...
```

so this branch needs to stay, and we do want to make this change. The other checks in this file should change to be based on `$sc_task` as well, but I can address this in a follow-up PR.

siliconcompiler/core.py
```diff
@@ -1218,15 +1218,15 @@ def find_result(self, filetype, step, jobname=None, index='0'):
...
+            step (str): Task step name ('syn_asic', 'place', etc)
             jobname (str): Jobid directory name
             index (str): Task index
 
         Returns:
             Returns absolute path to file.
 
         Examples:
-            >>> manifest_filepath = chip.find_result('vg', 'syn')
+            >>> manifest_filepath = chip.find_result('vg', 'syn_asic')
```
nmoroze, line 1229: Same here as above.

nmoroze (approved):
Looks good, the compile.tcl change seems correct!

I commented on a few places in documentation where it looks like step names got updated with "syn_asic" (my understanding is the step name stays "syn", while the task names are what change?)

gadfort (approved):

PR #1229, opened 2023-01-24
Use of .input and cmdline fix

gadfort:
Fixes:
- Ensures the create_cmdline uses the same infrastructure for .input to avoid that code duplication.
- Fix bvs->bsv again

Changes:
- examples and tests use .input
- vg renamed as a netlist / verilog instead of rtl / verilog


siliconcompiler/core.py
```diff
@@ -436,18 +436,7 @@ def create_cmdline(self, progname, description=None, switchlist=None, input_map=
...
-                    if self.valid('input', fileset, filetype):
-                        self.add('input', fileset, filetype, source)
-                    else:
-                        self.set('input', fileset, filetype, source)
-                    self.logger.info(f'Source {source} inferred as {fileset}/{filetype}')
-                else:
-                    self.logger.warning('Unable to infer input type for '
-                        f'{source} based on file extension, ignoring. Use the '
-                        '-input flag to provide it explicitly.')
+                self.input(source, iomap=input_map)
```
nmoroze, line 439: Very nice!

nmoroze (approved):
LGTM, thanks for the cleanup!

aolofsson:
I love this change! The only concern I would have is that the really important naming methodology is in the header of the input function, but the fileset names also effects task/tool implementations. That "contract" should be clearly documented somewhere central (not just in the input description).

gadfort:
@aolofsson I've made a note of the documentation need, I think we have several other places where we need to fix documentation to reflect all the changes, so I can add it in then.

PR #1232, opened 2023-01-25
Schema: remove openroad specific parameters and move to pdk

gadfort:
See all_schema for associated changes.

asic,var used for now for some parameters since there isn't a better place yet.
names of openroad and yosys pdk/tool variables I'm open to suggestions.


aolofsson (approved):
Looks great!

PR #1233, opened 2023-01-26
update core to use tasks instead of step names

gadfort:
Fixes:
- Updates left over references to steps to be specific to tasks instead.

Changes:
- Changes names of builtin tasks to be a tool name of `builtin` with the associated task and adds `import` as a task which behaves like a `nop` but is needed for flow validation.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -103,7 +103,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         self.status = {}
 
         self.builtin = ['minimum','maximum',
-                        'nop', 'mux', 'join', 'verify']
+                        'nop', 'mux', 'join', 'verify', 'import']
 
         # We set 'design' and 'loglevel' by directly calling the schema object
         # because of a chicken-and-egg problem: self.set() relies on the logger,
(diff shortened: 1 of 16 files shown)


siliconcompiler/core.py
```diff
@@ -103,7 +103,7 @@ def __init__(self, design, loglevel=None):
         self.status = {}
 
         self.builtin = ['minimum','maximum',
-                        'nop', 'mux', 'join', 'verify']
+                        'nop', 'mux', 'join', 'verify', 'import']
```
aolofsson, line 106: Are we 100% committed to the 'import' function? Going once, twice,...it's just a string and I know there have been some places where it clashes with the python 'import' key word. It's a good word (which is why we chose it), just need to make sure there is no dead end with the new import module approach.

aolofsson (changes requested):
Confirmation that we are committed to 'import' before making it an official builtin task.

siliconcompiler/core.py
gadfort: Now is the time to correct it (if ever)... the name conflict with python does make some things a little 'strange', but it makes sense and I cannot think of a better name. If there is something else it's pretty each to correct it now (in this PR). Note that `import` is not implemented as a function (which would fail), but as a "alias" to "nop" so we have symmetry with the requirement of an import task in each job.

@WRansohoff  found that by using `importlib` something called `import` can be imported. We just can't have `def`'s with that name.

WRansohoff: It is a bit awkward to have to do `importlib.import_module('tools.surelog.import')` instead of `from tools.surelog import import`.

There are other cases where we need to use `importlib` though, like tests which iterate over a list of strings representing module names.

I wouldn't object to renaming it (`pack`, `pickle`, `prepare`, `init`, ...?), but I don't feel strongly that it needs to be done.

gadfort: While `import` is awkward, it's also intuitive and provides some symmetry with `export`. Most users will not need to reference the `import` module by name but as a task (so we handle the lookup), so I don't think it will cause much trouble. We can discuss tomorrow.

aolofsson (approved):

PR #1234, opened 2023-01-26
Adding enum to schema type

gadfort:
Changes:
- Add enum as type to the schema
- Changes `mode`/`loglevel`/`status` to enum, I'm open to suggestions on more or if `mode` should remain a string.


docs/_ext/schemagen.py
```diff
@@ -11,7 +11,7 @@
 class SchemaGen(SphinxDirective):
 
     def run(self):
-        self.env.note_dependency('../siliconcompiler/schema.py')
+        self.env.note_dependency('../siliconcompiler/schema_cfg.py')
```
aolofsson, line 14: Isn't this missing something, isn't it 'siliconcompiler/schema/schema_cfg.py'?

siliconcompiler/schema/schema_cfg.py
```diff
@@ -1038,7 +1043,8 @@ def schema_flowgraph(cfg, flow='default', step='default', index='default'):
             * "success": task ran successfully
             * "error": task failed with an error
 
-            An empty value indicates the task has not yet been completed.""")
+            An empty value indicates the task has not yet been completed.""",
```
aolofsson, line 1046: Can we put this right below type for readability? Feels like the enum list belongs with the type since it's static.

siliconcompiler/schema/schema_cfg.py
```diff
@@ -1971,7 +1977,7 @@ def schema_option(cfg):
 
     # Compilation
     scparam(cfg, ['option', 'mode'],
-            sctype='str',
+            sctype='enum',
```
aolofsson, line 1980: Since these mode strings get called out in the code, I agree that it should be an enum. Otherwise, the user ends up getting weird downstream errors.

siliconcompiler/schema/schema_cfg.py
```diff
@@ -2127,7 +2134,8 @@ def schema_option(cfg):
                 "api: chip.set('option', 'loglevel', 'INFO')"],
             schelp="""
             Provides explicit control over the level of debug logging printed.
-            Valid entries include INFO, DEBUG, WARNING, ERROR.""")
+            Valid entries include INFO, DEBUG, WARNING, ERROR.""",
+            enum=["INFO", "DEBUG", "WARNING", "ERROR"])
```
aolofsson, line 2138: Was that the last one, I seem to remember there were more special strings?

aolofsson (approved):
A really nice addition!

siliconcompiler/schema/schema_cfg.py
gadfort: I can double check, I just copied from the line above

siliconcompiler/schema/schema_cfg.py
gadfort: I'll move them up in the _cfg.py

docs/_ext/schemagen.py
gadfort: yes, you are correct

PR #1236, opened 2023-01-26
Adding job scheduler parameters to schema

aolofsson:
This PR adds user facing server schema parameters for driving driving job schedulers such as lsf, slurm, sge. Until now, we have treated the job scheduler as a black box, with only one switch to drive local vs remote. All server side settings were hidden from the user. While this worked well for simple batch level 'gcd' level processing, it is not configurable enough for production level SOC design. Pretty much every large chip since 1999 has used some kind of load sharing/job scheduling tool for distributing verification and implementation jobs across local on premises grids:

https://www.deepchip.com/posts/0312.html

The implementation of this PR is still to be done, but the features embedded in the schema should be non-controversial and "must haves" for expert chip designers looking to leverage on-site or cloud compute farms managed by slurm/lsf.

Key features (global or per step/index)
- request machines with N cores
- request machine with M memory
- message notification
- timeout
- queue selection
- defer job start

tbd:
- arrays
- others...


siliconcompiler/schema/schema_cfg.py
```diff
@@ -3506,6 +3484,177 @@ def schema_constraint(cfg):
...
+                "cli: -server_option '--pty bash'",
+                "api: chip.set('server','option, '--pty bash')"],
+            schelp="""
+            List of run time options passed through to the job scheduler for
+            expert users. See the job scheduler (slurm, lsf, gse) for
+            information about available options. The user specified options must
+            be compatible with the standardized job scheduler options (memory,
+            cores,...).""")
+
+    scparam(cfg, ['server', 'msgevent'],
```
WRansohoff, line 3628: I agree with what Peter said today, that this and `'msgcontact'` may be nice-to-haves which can be added in a future change.

siliconcompiler/schema/schema_cfg.py
```diff
@@ -3506,6 +3484,177 @@ def schema_constraint(cfg):
...
+            example= ["cli: -server_defer 16:00",
+                    "api: chip.set('server', 'defer', '16:00')"],
+            schelp="""
+            Defer initiation of job until the specified time. The parameter
+            is pass through string for remote job scheduler such as slurm.
+            For more information abotut the exact format specification, see
+            the job scheduler documentation. Examples of valid slurm specific
+            values include: now+1hour, 16:00, 010-01-20T12:34:00.""")
+
+    scparam(cfg, ['server', 'option'],
```
WRansohoff, line 3614: It feels a bit odd to accept an arbitrary list of options, when we have dedicated schema parameters for some options such as `--constraint`, resource limits, etc.

siliconcompiler/schema/schema_cfg.py
```diff
@@ -3506,6 +3484,177 @@ def schema_constraint(cfg):
...
+            switch="-server_timeout <str>",
+            example= ["cli: -server_timeout 16:00",
+                    "api: chip.set('server', 'timeout', '16:00')"],
+            schelp="""
+            Specifies the limit on the total job run time.
+            Acceptable slurm time formats include "minutes", "minutes:seconds",
+            "hours:minutes:seconds", "days-hours", "days-hours:minutes",
+            and "days-hours:minutes:seconds".""")
+
+    scparam(cfg, ['server', 'interactive'],
```
WRansohoff, line 3589: May be tricky to use this argument with SC since most of our current flows are designed around batch jobs. Could be a nice to have which can be added as we bring interactive workloads into the framework (analog, etc)

siliconcompiler/schema/schema_cfg.py
```diff
@@ -3506,6 +3484,177 @@ def schema_constraint(cfg):
...
+            schelp="""
+            Provides machine feature constraints to the job scheduler for
+            selecting compute nodes. Feature names are strings assigned to
+            computes nodes by the job scheduler administrator. The constraint
+            parameter is passed through to the job schedular unmodified. For
+            the slurm scheduler, this translates to the '--constraint'
+            switch. See the slurm documentation for examples of the types of
+            constraints supported.""")
+
+    scparam(cfg, ['server', 'queue'],
```
WRansohoff, line 3552: Should we call it `partition`? Could be confusing since Slurm has a command called `squeue`.

siliconcompiler/schema/schema_cfg.py
```diff
@@ -3506,6 +3484,177 @@ def schema_constraint(cfg):
 
     return cfg
 
+############################################
+# Server Configuration
+############################################
+
+def schema_server(cfg):
+
+    scparam(cfg, ['server', 'jobscheduler'],
```
WRansohoff, line 3493: Like Peter mentioned, may be worth keeping this under `option`. It may also be worth considering other names for the top-level key, as the job scheduling parameters will work the same in both local and server runs. (Depending on the server)

Maybe `cluster`, `grid`, or `scheduler`?

WRansohoff (changes requested):
I really like the idea of formalizing the `scheduler.py` configurations in the schema. Just some comments / questions about how we should structure the keys for an initial implementation.

gadfort (approved):
Looks fine. Only a couple of questions/comments
- Should we add `For more information, see the job scheduler documentation` to each of the to make it clear that this is driven by the scheduler. I think it's fine to have the slurm examples (since thats all that is supported).

siliconcompiler/schema/schema_cfg.py
aolofsson: Queue is the term that people are familiar with from chip design and to me a clearer name. If we are going to support all three, then we will have translation/confusion regardless. I consider queue a better word than partition.
https://slurm.schedmd.com/rosetta.html

gadfort: I would prefer `queue`

siliconcompiler/schema/schema_cfg.py
```diff
@@ -3506,6 +3484,177 @@ def schema_constraint(cfg):
...
+            sctype='str',
```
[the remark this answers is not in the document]
gadfort: enum is merged, so this could be updated if appropriate

siliconcompiler/schema/schema_cfg.py
```diff
@@ -3506,6 +3484,177 @@ def schema_constraint(cfg):
...
+                "api: chip.set('server','option, '--pty bash')"],
+            schelp="""
+            List of run time options passed through to the job scheduler for
+            expert users. See the job scheduler (slurm, lsf, gse) for
+            information about available options. The user specified options must
+            be compatible with the standardized job scheduler options (memory,
+            cores,...).""")
+
+    scparam(cfg, ['server', 'msgevent'],
+            sctype='str',
```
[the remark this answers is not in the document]
aolofsson: The slurm job scheduler has specific keywords for handling different types of events and do not support lists. If we want to support a list it would need to be outside of slurm and we would need to come up with an enumerated list of keywords for those events.

siliconcompiler/schema/schema_cfg.py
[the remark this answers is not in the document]
aolofsson: Yes! (this PR predates your enum PR.:-)

siliconcompiler/schema/schema_cfg.py
aolofsson: You are right, I forgot that we support running jobs on premises with slurm and without the -remote flag. Since slurm/lsf is such an important feature, I think it warrants continuing this path. I will change the name to 'scheduler'.

siliconcompiler/schema/schema_cfg.py
aolofsson: Having the ability to get access to a large work station on the fly for interactive digital work/gds viewing or analog view is a fundamental need. Getting this feature to work will be a bit tricky, but it's definitely a must have.

siliconcompiler/schema/schema_cfg.py
aolofsson: Perhaps, but what is the alternative? 1.) Be completely prescriptive and only allow the options that we specify 2.) Support all the slurm options (which would be too much). https://slurm.schedmd.com/srun.html 3.) support all options for all schedulers (lsf, sge, slurm). 

It may be that --constraint went too far?

Like with the pnr flow ('var') the idea here is to offer 'experts'/'advanced users' pass through their own variables at their own risk, while making the most common standardized (queue, cores, mem)

siliconcompiler/schema/schema_cfg.py
```diff
@@ -3514,6 +3475,168 @@ def schema_constraint(cfg):
 
     return cfg
 
+############################################
+# Scheduler Configuration
+############################################
+
+def schema_server(cfg):
+
+    scparam(cfg, ['scheduler', 'name'],
```
WRansohoff, line 3484: Maybe `type` or `engine`?

WRansohoff (approved):

siliconcompiler/schema/schema_cfg.py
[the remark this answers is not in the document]
gadfort: Type needs to be enum as well.

gadfort (changes requested):
fix sctype and place parameters based on discussion in meeting.

gadfort (approved):
Looks great

siliconcompiler/schema/schema_cfg.py
```diff
@@ -2649,6 +2604,137 @@ def schema_option(cfg):
             before a run.
             """)
 
+    scparam(cfg, ['option', 'timeout'],
```
WRansohoff, line 2607: I like the idea of providing this option for all jobs, not just scheduled ones. Should it have `pernode='optional'`, so we can specify per-task timeouts?

WRansohoff (approved):

PR #1237, opened 2023-01-26
Replace `load_[...]` functions with `use`

WRansohoff:
This PR removes the `load_pdk`, `load_lib`, `load_flow`, and `load_target` methods, replacing them with the `use` method which accepts a Python module rather than a string.

The underlying `setup` methods are still structured to modify a `Chip` object instead of returning a `Schema` sub-class for the `Chip` to incorporate. That will be addressed in a follow-up change.

### DIFF
--- a/examples/benchmark/benchmark.py
+++ b/examples/benchmark/benchmark.py
@@ -2,6 +2,8 @@ import os
 import sys
 import time
 import siliconcompiler
+from siliconcompiler.flows import asicflow
+from siliconcompiler.targets import skywater130_demo
 
 # Setting up the experiment
 
@@ -26,7 +28,7 @@ def main():
             wall_start = time.time()
             chip = siliconcompiler.Chip(design)
             chip.set('jobname', f"job{n}")
-            chip.load_target('skywater130_demo')
+            chip.use(skywater130_demo)
             chip.set('relax', True)
             chip.set('quiet', True)
             chip.set('remote',False)
@@ -41,7 +43,7 @@ def main():
             chip.set('flowarg', 'syn_np', n)
             chip.set('flowarg', 'place_np', n)
             chip.set('flowarg', 'cts_np', n)
-            chip.load_flow('asicflow')
+            chip.use(asicflow)
             chip.set('flow', 'asicflow')
 
             # Set router to 1 thread to not interfere with measurement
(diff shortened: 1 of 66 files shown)


siliconcompiler/core.py
```diff
@@ -426,7 +426,11 @@ def create_cmdline(self, progname, description=None, switchlist=None, input_map=
             if 'fpga_partname' in cmdargs.keys():
                 self.set('fpga', 'partname', cmdargs['fpga_partname'], clobber=True)
             # running target command
-            self.load_target(cmdargs['option_target'])
+            try:
+                target_module = importlib.import_module(f'targets.{cmdargs["option_target"]}')
```
gadfort, line 430: Add a TODO around this, I think we need to "find" this module in PYTHON_PATH or in targets.{}

siliconcompiler/core.py
```diff
@@ -426,7 +426,11 @@ def create_cmdline(self, progname, description=None, switchlist=None, input_map=
...
+                self.use(target_module)
+            except ModuleNotFoundError:
+                self.error(f'Target module {cmdargs["option_target"]} not found in $SCPATH or siliconcompiler/targets/.')
```
gadfort, line 433: This error message doesn't seem correct any more. Should probably be python_path and SCROOT

gadfort (changes requested):
Should we leave `load_target` in place for now since it's still a special case for use? That way we can implement the others first and do target last?

examples/benchmark/benchmark.py
```diff
@@ -41,7 +43,7 @@ def main():
             chip.set('flowarg', 'syn_np', n)
             chip.set('flowarg', 'place_np', n)
             chip.set('flowarg', 'cts_np', n)
-            chip.load_flow('asicflow')
+            chip.use(asicflow)
```
aolofsson, line 46: Isn't the skywater130_demo loading the asicflow already?

siliconcompiler/core.py
WRansohoff: Good point - users could have their own targets in different locations, so we might also want to import the command-line target string literally (e.g. `sc -target targets.freepdk45_demo` instead of `sc -target freepdk45_demo`)

examples/benchmark/benchmark.py
WRansohoff: Yes, but I think that this example might re-load it to have multiple parallel tasks for syn, place, and cts.

aolofsson: You are right!

gadfort (approved):

aolofsson (approved):

PR #1238, opened 2023-01-27
Implement configurable step/index overrides

nmoroze:
This PR adds the long-awaited step/index override functionality! Each schema parameter now has a "pernode" field that can be set to one of 3 values:

- "none": step and index are illegal arguments for this parameter
- "optional": step and/or index may be provided for this parameter, overriding the global value (for asic, tools, constraints, etc.)
- "mandatory": step and index must be provided for this parameter (for metrics/records)

I've enabled it for one parameter so far, `['asic', 'logiclib']`, as a test. We can come up with a plan for turning it on for more parameters in chunks.


aolofsson:
We have 'value', 'defvalue' right now.  What do you think about a name change to 'nodevalue' instead of current proposal of 'pernode_value'.

siliconcompiler/schema/schema_cfg.py
```diff
@@ -84,6 +86,9 @@ def scparam(cfg,
         cfg['help'] = schelp
         cfg['signature'] = signature
         cfg['notes'] = notes
+        # none, optional, mandatory
+        cfg['pernode'] = pernode
+        cfg['pernode_value'] = {}
```
aolofsson, line 91: Can we make pernode_value --> 'nodevalue'.
We have gone this far without an underscore in the fields. Feels out of place in terms of "lock and feel.", it's an outlier compared to all others. (we have 13 fields so far)

aolofsson (changes requested):
Very elegant! My only comment is really the name change.

siliconcompiler/schema/schema_cfg.py
nmoroze: Sure, easy change! I'm happy with that.

siliconcompiler/schema/schema_cfg.py
```diff
@@ -28,7 +28,8 @@ def scparam(cfg,
             switch=None,
             example=None,
             schelp=None,
-            enum=None):
+            enum=None,
+            pernode='none'):
```
WRansohoff, line 32: What do you think about 'never' or 'forbidden' as the opposite option to 'mandatory', instead of 'none'?

siliconcompiler/schema/schema_obj.py
```diff
@@ -78,8 +78,27 @@ def get(self, *keypath, field='value', job=None):
         if not Schema._is_leaf(cfg):
             raise ValueError(f'Invalid keypath {keypath}: get() must be called on a complete keypath')
 
+        if cfg['pernode'] == 'none' and (step is not None or index is not None):
+            raise ValueError(f'step and index are not valid for keypath {keypath}')
```
WRansohoff, line 82: "step and index cannot be used with keypath"?

"step and index are not valid" makes me think that some other combination of step/index might be valid. Ditto for the set() and add() error messages.

siliconcompiler/schema/schema_obj.py
```diff
@@ -500,7 +578,10 @@ def write_tcl(self, fout, prefix=""):
 
         for key in allkeys:
             typestr = self.get(*key, field='type')
-            value = self.get(*key)
+            if self.get(*key, field='pernode') != 'none':
+                value = self.get(*key, step=step, index=index)
```
WRansohoff, line 582: If a task-specific value has not been set on an 'optional' parameter, will this still return the correct "step=None, index=None" value?

tests/schema/test_schema.py
```diff
@@ -9,3 +12,16 @@ def test_list_of_lists():
     schema.set('test', [['foo']])
 
     assert schema.get('test') == [['foo']]
+
+def test_pernode_mandatory():
```
WRansohoff, line 16: Maybe `test_pernode_none`, `test_pernode_optional`?

WRansohoff (commented):
Looks good overall, I like the simplicity. Just a few minor questions and comments.

aolofsson (approved):

siliconcompiler/schema/schema_obj.py
```diff
@@ -151,9 +202,24 @@ def add(self, *args, field='value'):
...
+                    cfg['nodevalue'][step] = {}
+                if index not in cfg['nodevalue'][step]:
+                    cfg['nodevalue'][step][index] = []
+                cfg['nodevalue'][step][index].extend(value)
+            elif step is not None:
+                if step not in cfg['nodevalue']:
+                    cfg['nodevalue'][step] = {}
+                if 'default' not in cfg['nodevalue'][step]:
+                    cfg['nodevalue'][step]['default'] = []
+                cfg['nodevalue'][step]['default'].extend(value)
```
gadfort, line 217: why not set `set`?

siliconcompiler/schema/schema_obj.py
```diff
@@ -151,9 +202,24 @@ def add(self, *args, field='value'):
...
 
-        cfg[field].extend(value)
         if field == 'value':
-            cfg['set'] = True
+            if step is not None and index is not None:
+                if step not in cfg['nodevalue']:
+                    cfg['nodevalue'][step] = {}
+                if index not in cfg['nodevalue'][step]:
+                    cfg['nodevalue'][step][index] = []
+                cfg['nodevalue'][step][index].extend(value)
```
gadfort, line 211: why not set `set`?

gadfort (approved):
Looks great.

gadfort:
@nmoroze We'll merge this to get the ball rolling, anticipating there will be some fixes needed along the way

PR #1239, opened 2023-01-27
klayout: switch to using Chip() rather than JSON

nmoroze:
This PR modifies our KLayout driver so that we can use SiliconCompiler in the drivers rather than accessing the JSON manifest directly. Thanks to @gadfort, the trick is to provide the user's Python environment to KLayout via its own special environment variable so that it has access to SC and all of its dependencies.

This change will be a requirement for supporting the step/index abstraction being added in PR #1238.


gadfort (approved):
Looks good, we just need to merge the other PR first

nmoroze:
Alright, I think I figured out the timeout issue. There were a few layers that made this a bit tricky to debug.

The fundamental bug was in klayout_show.py, and was triggered in `test_show_nopdk`. This test clears out `SCPATH` to make it so that PDK/std. cell library files can't be resolved (this is meant to simulate a remote run where a user may not have cloned FreePDK45/Sky130 locally, but still wants to pop open a GDS). The issue was a call to `chip.get('library', sc_mainlib, 'output', sc_stackup, 'lef')`, which needed a `chip.valid(...)` guard, since this keypath doesn't exist in the manifest if the file can't be resolved.

One semi-annoying thing is that doing, for example, `chip.get('output', stackup, 'lef')` at the top-level would work and return an empty list, but that doesn't work for things under the `library` keypath since it's a weird case without any "default" structures underneath. In essence, `chip.get('output', stackup, 'lef')` would return `cfg['output']['default']['default']['defvalue']` if the keypath `['output', <stackup>, 'lef']` didn't exist, but the same doesn't apply to `'library'` since there is no such thing as `'library, 'default', ...`. We could potentially tweak this behavior, but that would be a larger change.

This test failure didn't occur locally since I have an editable install of SC that lets it resolve the PDK/std. cell library filepaths without having SCPATH be set. To make the root cause more visible in CI, I added timeouts to the show tests, and got rid of the quiet parameter so the KLayout output is visible without having to SSH in and view logs.

PR #1242, opened 2023-01-28
fix leftover import in oh_experiment

gadfort:


WRansohoff:
I don't think this is broken, is it? `load_target` is just a shortcut for `import` followed by `use` now.

gadfort:
> I don't think this is broken, is it? `load_target` is just a shortcut for `import` followed by `use` now.

The import was broken. But since we decided not to modify load_target I also swapped that back.

WRansohoff (approved):

PR #1245, opened 2023-01-30
Make breakpoint boolean and use step/index

gadfort:
Changes:
- bkpt -> breakpoint
- Can now specify as boolean and via new step/index methods
- Update OpenROAD to use


nmoroze (approved):
Looks good to me! I'll fix the CLI to accept step/index overrides this week, which will restore the functionality of setting breakpoints via CLI.

PR #1251, opened 2023-02-01
Clean up `summary()`/HTML report logic

nmoroze:
This PR cleans up the `summary()`/HTML report logic a bit in preparation for step/index-related changes. 

The primary change is to reduce the HTML report template's reliance on direct access to the schema so that we can abstract away the step/index-related schema changes. The Python logic in `summary()` is now responsible for directly passing relevant values to the template in a more concise form. 

Unfortunately we can't fully abstract away the manifest from the HTML report, since it also has functionality for viewing the entire tree/modifying checklist values and re-exporting the manifest. Eventually at least some of this logic will need to be updated for our new step/index abstraction scheme, but this is lower priority since it doesn't block switching over to the new scheme.

As part of this change, I also refactored `summary()` a bit to separate out the ASCII table display logic from the actual data extraction logic, since we want to have raw data to feed into the template, as opposed to formatted strings.

### DIFF
--- a/tests/core/test_summary.py
+++ b/tests/core/test_summary.py
@@ -1,5 +1,4 @@
 # Copyright 2020 Silicon Compiler Authors. All Rights Reserved.
-import os
 import siliconcompiler
 
 import pytest
@@ -31,6 +30,7 @@ def test_steplist(gcd_with_metrics, capfd):
 
     gcd_with_metrics.summary()
     stdout, _ = capfd.readouterr()
+    print(stdout)
 
     assert 'import0' not in stdout
     assert 'syn0' in stdout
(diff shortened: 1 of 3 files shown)


tests/core/test_summary.py
```diff
@@ -31,19 +30,19 @@ def test_steplist(gcd_with_metrics, capfd):
 
     gcd_with_metrics.summary()
     stdout, _ = capfd.readouterr()
+    print(stdout)
```
gadfort, line 33: is this left over?

gadfort (approved):
Looks good. I assume the template changes were tested.
Just need to double check on the print()

tests/core/test_summary.py
nmoroze: I think this should stay - this test captures stdout to check the output of summary(), but that also hides it. It's useful to print back out for debugging - I added an explanatory comment.

nmoroze:
Yup, tested the template changes by viewing the outputs of the HTML report for the GCD example in lieu of an extensive CI test for this.

WRansohoff (approved):
Thanks for cleaning this up! I appreciate the removal of the very long dictionary key paths from the template file.

PR #1252, opened 2023-02-01
return to load_target

gadfort:
Changes:
- since `load_target` will not use `.use`, this returns all instances of it back to the previous method


nmoroze (approved):
LGTM

WRansohoff (approved):
Makes sense, it didn't quite fit cleanly with the other setup functions.

PR #1253, opened 2023-02-02
Add an option to hide layers by default in show and screenshot tasks

WRansohoff:
I'm not a marketer, but I think that our auto-generated Skywater130 design screenshots would look better if the details weren't obscured by blocks of bright magenta. I could see an argument for only applying this to the `screenshot` task, and not `show`, but users can override the default visibility settings in the KLayout GUI.

Before (`picorv32`/`sky130hd`, all open-source):

picorv32_top_meh

After:

picorv32_top

### DIFF
--- a/siliconcompiler/pdks/skywater130.py
+++ b/siliconcompiler/pdks/skywater130.py
@@ -112,6 +112,9 @@ def setup(chip):
     chip.set('pdk', process, 'var', 'openroad', 'pin_layer_vertical', stackup, 'met2')
     chip.set('pdk', process, 'var', 'openroad', 'pin_layer_horizontal', stackup, 'met3')
 
+    # Hide the 81/4 'areaid.standardc' layer by default; it puts opaque purple over most core areas.
+    chip.set('pdk', process, 'var', 'klayout', 'hide_layers', stackup, ['81/4'])
+
     # PEX
     chip.set('pdk', process, 'pexmodel', 'openroad', stackup, 'typical',
         pdkdir + '/pex/openroad/typical.tcl')
--- a/siliconcompiler/pdks/skywater130.py
+++ b/siliconcompiler/pdks/skywater130.py
@@ -113,7 +113,7 @@ def setup(chip):
     chip.set('pdk', process, 'var', 'openroad', 'pin_layer_horizontal', stackup, 'met3')
 
     # Hide the 81/4 'areaid.standardc' layer by default; it puts opaque purple over most core areas.
-    chip.set('pdk', process, 'var', 'klayout', 'hide_layers', stackup, ['81/4'])
+    chip.set('pdk', process, 'var', 'klayout', 'hide_layers', stackup, ['areaid.standardc'])
 
     # PEX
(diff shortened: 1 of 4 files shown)


siliconcompiler/tools/klayout/screenshot.py
```diff
@@ -21,6 +21,11 @@ def setup(chip):
     chip.set('tool', tool, 'task', task, 'script', step, index, script, clobber=clobber)
     chip.set('tool', tool, 'task', task, 'option', step, index, option, clobber=clobber)
 
+    pdk = chip.get('option', 'pdk')
+    stackup = chip.get('option', 'stackup')
+    if chip.valid('pdk', pdk, 'var', 'klayout', 'hide_layers', stackup):
+        scrot_hide_layers = chip.get('pdk', pdk, 'var', 'klayout', 'hide_layers', stackup)
```
gadfort, line 27: scrot_hide_layers -> hide_layers?

siliconcompiler/pdks/skywater130.py
```diff
@@ -112,6 +112,9 @@ def setup(chip):
     chip.set('pdk', process, 'var', 'openroad', 'pin_layer_vertical', stackup, 'met2')
     chip.set('pdk', process, 'var', 'openroad', 'pin_layer_horizontal', stackup, 'met3')
 
+    # Hide the 81/4 'areaid.standardc' layer by default; it puts opaque purple over most core areas.
```
gadfort, line 115: can you use the `areaid.standardc` in the list instead? I think that would be clearer than the GDS layer/purpose

WRansohoff: We could accept either; it looks like `layer.name` returns a value in the format `"areaid.standardc - 81/4"`

siliconcompiler/tools/klayout/screenshot.py
WRansohoff: Yeahh, I followed the OpenROAD script naming conventions here, but it does seem like a bit of an odd choice for an abbreviation.

gadfort: If that came from the OR script, we should probably fix it there too. I don't think I picked a good name.

siliconcompiler/pdks/skywater130.py
gadfort: I think the layer name is clearer

siliconcompiler/tools/klayout/screenshot.py
WRansohoff: In any case, I made it `layers_to_hide` here and in `show.py`.

siliconcompiler/pdks/skywater130.py
WRansohoff: I put the layer name into the Sky130 PDK, but made the KLayout script accept either, in case someone wants to use a PDK without a clear layer properties file, or one that is being actively developed, or something like that.

gadfort (approved):

PR #1254, opened 2023-02-02
flows cleanup

gadfort:
Changes:
- Several flows were acting as quasi targets, this fixes the flows to just setup the flowgraph and moves settings into the appropriate targets.

Adds:
- a helper function to set the showtools in the targets so it's consistent

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -142,6 +142,9 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
             return codecs.ignore_errors(e)
         codecs.register_error('ignore_with_warning', log_error_handler)
 
+        # Setup default show tools
+        utils.set_common_showtools(self)
+
     ###########################################################################
     @property
     def design(self):
@@ -2976,13 +2979,15 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
             Creates a task with step='place' and index=0 and binds it to the 'openroad' tool.
         '''
 
+        index =str(index)
+
         # bind tool to node
-        self.set('flowgraph', flow, step, str(index), 'tool', tool)
-        self.set('flowgraph', flow, step, str(index), 'task', task)
+        self.set('flowgraph', flow, step, index, 'tool', tool)
+        self.set('flowgraph', flow, step, index, 'task', task)
 
         # set default weights
         for metric in self.getkeys('metric', 'default', 'default'):
-            self.set('flowgraph', flow, step, str(index), 'weight', metric, 0)
+            self.set('flowgraph', flow, step, index, 'weight', metric, 0)
 
     ###########################################################################
     def edge(self, flow, tail, head, tail_index=0, head_index=0):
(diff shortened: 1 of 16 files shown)


siliconcompiler/utils.py
```diff
@@ -302,3 +302,13 @@ def format_fileset_type_table(indent=12):
         table += f"{indent}{filetype:<10}| {fileset:<11}| {ext}\n"
 
     return table
+
```
aolofsson, line 305: It's ok, but this one feels awkward. I agree that we should try to not repeat ourselves, but this is coupling/hardcoding settings for specific tools inside the utils flow. How about putting this util function inside the target dir?

siliconcompiler/core.py
```diff
@@ -2973,13 +2976,15 @@ def node(self, flow, step, tool, task, index=0):
             Creates a task with step='place' and index=0 and binds it to the 'openroad' tool.
         '''
 
+        index =str(index)
```
nmoroze, line 2979: nit: space after =

siliconcompiler/utils.py
gadfort: sure, I can move it there.

siliconcompiler/utils.py
```diff
@@ -302,3 +302,13 @@ def format_fileset_type_table(indent=12):
...
+def set_common_showtools(chip):
+
+    # Physical only
+    chip.set('option', 'showtool', 'gds', 'klayout')
+    chip.set('option', 'showtool', 'lef', 'klayout')
+
+    # Design
+    chip.set('option', 'showtool', 'def', 'openroad')
+    chip.set('option', 'showtool', 'odb', 'openroad')
```
nmoroze: Perhaps using `clobber=False` for these would be useful to make sure any user-supplied settings don't get overwritten if they're performed before loading the target.

nmoroze (approved):
LGTM, just a few small comments.

I don't have super strong feelings about the location of the helper function, but I think Andreas' suggestion there seems reasonable.

siliconcompiler/core.py
gadfort: Thanks for catching

WRansohoff:
It looks like this PR may have broken the docs tests.

siliconcompiler/targets/asap7_demo.py
```diff
@@ -1,4 +1,5 @@
 import siliconcompiler
+from . import utils
```
WRansohoff, line 2: Syntax error. Try `import utils`.

gadfort: there was a separate issue with the docs. The current HEAD of main is green for the build docs. Are they still broken?

WRansohoff: They're failing on the tip of main if I run `make html` in the `docs/` directory; I think the recent merges were green because the tests ran on the branch pre-merge.

PR #1255, opened 2023-02-02
Fixes to facilitate mandatory step/index

nmoroze:
This PR makes a bunch of tweaks to fix oversights in the initial implementation, mainly regarding how mandatory step/index-type parameters are implemented.

Changes:

## Rename 'pernode' field values

- Previously `(none, optional, mandatory)`, now `(never, optional, required)`
- This sounds better to me after having a long weekend to sleep on it, credits to @WRansohoff for the suggestion of none->never
- I'd say this is last chance to change, so open to feedback on names :)

## Make clobber work on a per-node basis

- Initially, `clobber=False` would prevent a `set()` if any step/index has been set
- However, this is problematic in cases where per-node values are mandatory, the different values should be treated as independent 
- At the same time, we don't want to set per-node overrides for `pernode='optional'` keypaths when a global value has been set already and clobber=False - this would break tool drivers
- Solution:
  - `cfg['set']` only gets set when the global value is being modified
  - A value can only be modified when clobber=False is `cfg['set']` is False AND the supplied step/index hasn't been set
  - This gives us a solution that works generally

## Add `Schema.getvals(...)` helper function

There are various cases where we iterate over all keypaths in the schema and do something with the values (generally transforming them and stashing them back in). The presence of step/index values complicates this looping a bit, so I added a `getvals()` helper function that provides all values along with their associated step/index for a particular keypath.

For example:

```python
>>> chip.set('tool', 'openroad', 'task', 'floorplan', 'threads', 2)
>>> chip.set('tool', 'openroad', 'task', 'floorplan', 'threads', 4, step='fpstep', index=1)
>>> chip.getvals()
[(2, None, None), (4, 'fpstep', '1')]
```

The result can easily be iterated over and used by unpacking the tuple. 

## Miscellaneous cleanup

In using `getvals()` to fix up various methods, I also took a few opportunities to do more cleanup/refactoring. For example:

- `Schema._is_empty()` I renamed to `Schema._is_set()`, and `Schema._keypath_empty()` I renamed to `Schema._is_empty()` (these names better reflect what the methods mean). I also added step/index args to `Schema._is_empty()`
- Cleaned up `abspath()` and `prune()` a bit

A few other cleanup things:
- Added step/index to `find_files()`
- Improved step/index value checking and factored this out into a separate function
- Fixed a few tests to handle mandatory step/index parameters.

## To be done

For now `write_manifest()` will skip mandatory per-node values in the TCL dump (since these will be metrics and records, it isn't problematic for now). I'll fix this in an upcoming PR, after getting consensus on what exactly we want the TCL dump to look like.


tests/core/test_create_cmdline.py
```diff
@@ -93,6 +93,9 @@ def _mock_read_manifest(chip, manifest, clobber=False, clear=False):
     chip = siliconcompiler.Chip('test')
     for keypath in chip.allkeys():
         examples = chip.get(*keypath, field='example')
+        # TODO: undo this change once we support specifying pernode parameters using CLI
```
gadfort, line 96: What about optional?

gadfort (approved):
LGTM. I assume the TCL write and CLI TODOs are being tracked.
Please wait to merge until @aolofsson has weighed in.

aolofsson:
- Agree that "(never, optional, required)" is good
- Not sure I undestand how the 'set' solution would work based on the PR description. It's a single global value per parameter. not sure I get the part of "the step index" not having been set.  There is only one value, so are you looking for values in 'nodevalue'?
- Are you proposing that getvals() be a public function? This requires more discussion, so I suggest making it "_getvals()" to not block merging.

tests/core/test_create_cmdline.py
nmoroze: I don't think optional parameters would break this test since the CLI examples can just demonstrate how to set them globally (I'm envisioning that would look the same as it currently does)

siliconcompiler/core.py
```diff
@@ -848,7 +848,7 @@ def set(self, *args, field='value', clobber=True, step=None, index=None):
             self.error(e)
 
     ###########################################################################
-    def clear(self, *keypath):
+    def clear(self, *keypath, step=None, index=None):
```
aolofsson, line 851: I missed api name change earlier. At some point we had discussed 'unset'. With the parmeter field/state being 'set' and the value of 'set' being True/False. Not sure clear is the right term here...very subjective. Both clear/unset are a bit since we are reverting to the default value. 

Bash : unset delets a variable during program execution
Hardware: set ==set/preset (1), clear/reset(0)
css=unset resets property to defaut value


If I had to choose, I would say that set()/unset() is a more natural coupling thatn set()/clear(), but perhaps others can weigh in.

@gadfort

siliconcompiler/schema/schema_obj.py
```diff
@@ -235,16 +229,52 @@ def clear(self, *keypath):
...
+                index = 'default'
+            try:
+                del cfg['nodevalue'][step][index]
+            except KeyError:
+                # If this key doesn't exist, silently continue - it was never set
+                pass
 
         return True
 
+    def getvals(self, *keypath):
```
aolofsson, line 255: I recommend making this a hidden developer function until we work through use cases.

aolofsson (changes requested):
LGTM. Some questions around API and set/clobber.

siliconcompiler/core.py
gadfort: I agree with @aolofsson `unset` is a more natural pairing with `set`

siliconcompiler/schema/schema_obj.py
```diff
@@ -235,16 +229,52 @@ def clear(self, *keypath):
...
+        Returns a list of tuples of the form (value, step, index).
+        """
+        cfg = self._search(*keypath)
+
+        if not Schema._is_leaf(cfg):
+            raise ValueError(f'Invalid keypath {keypath}: getvals() must be called on a complete keypath')
+
+        vals = []
+        if cfg['pernode'] != 'required':
+            vals.append((self.get(*keypath, step=None, index=None), None, None))
```
WRansohoff, line 268: Could it make sense to order the output as `(step, index, value)` instead of `(value, step, index)`?

On the one hand, we are putting `step=`/`index=` at the end of function signatures since they're keyword arguments.

But in the schema, we usually do `chip.set([...], step, index, value)`.

siliconcompiler/schema/schema_obj.py
```diff
@@ -460,8 +490,15 @@ def error_msg(t):
         raise ValueError(f'Invalid field {field} for keypath {keypath}')
 
     @staticmethod
-    def _is_empty(cfg):
-        return not cfg['set']
+    def _is_set(cfg, step=None, index=None):
```
WRansohoff, line 493: If I'm reading this method correctly, it is supposed to indicate whether a parameter has been set to a value which is different from its default configuraiton?

Even though it's an internal method, it might be a good idea to add a docstring so that people don't assume `_is_set` means "has a non-empty value".

WRansohoff (approved):
Looks good, just a couple of formatting / docs questions.

nmoroze:
@aolofsson Example code for clobber behavior:

Required step/index:

```python
>>> chip.set('metric', 'errors', 0, step='syn', index=0) 
>>> chip.set('metric', 'errors', 1, step='syn', index=0, clobber=False) 
>>> chip.get('metric', 'errors', step='syn', index=0)
0 # previous set has no effect since it's same step/index

>>> chip.set('metric', 'errors', 1, step='floorplan', index=0, clobber=False) 
>>> chip.get('metric', 'errors', step='floorplan', index=0)
1 # previous set has an effect since it's a different step/index
```

Optional step/index:

```python
>>> chip.set('tool', 'openroad', 'task', 'floorplan', 'threads', 1)
>>> chip.set('tool', 'openroad', 'task', 'floorplan', 'threads', 4, step=step, index=index, clobber=False) 
>>> chip.get('tool', 'openroad', 'task', 'floorplan', 'threads', step=step, index=index)
1 # previous set has no effect since global already set
```

```python
>>> chip.set('option', 'param', 'N', 8, step='syn', index='0')
>>> chip.set('option', 'param', 'N', 4, clobber=False)
>>> chip.get('option', 'param', 'N', step='syn', index='0')
8 # first value returned since step/index set takes priority
>>> chip.get('option', 'param', 'N', step='floorplan', index='0')
4 # global value is still taken from the most recent set()
```

My only concern is that in a case where we are merging in a manifest with clobber=False, we won't get optional pernode values reflected if the global value has already been set:

```python
>>> chip1.set('option', 'param', 8, step='syn', index='0')
>>> chip1.write_manifest('chip2.pkg.json')

>>> chip2.set('option', 'param', 4)
>>> chip2.read_manifest('chip2.pkg.json', clobber=False)
>>> chip2.get('option', 'param', step='syn', index='0')
4 
```

Does that seem alright? It's a rare case, I'm not sure we have user code that does `read_manifest()` with clobber=False (just happens internally for merging manifests between nodes).

nmoroze:
@gadfort Opened an issue for tracking the TODOs: https://github.com/siliconcompiler/siliconcompiler/issues/1257

siliconcompiler/schema/schema_obj.py
nmoroze: It indicates whether a parameter has been set to any value by a user explicitly. Good suggestion - I added a docstring!

siliconcompiler/schema/schema_obj.py
nmoroze: That's a nice suggestion, I see both sides of this and don't have a particularly strong preference. @gadfort any thoughts?

I am making `_getvals()` private for the time being, so I do feel less urgent about getting this right now.

nmoroze:
Thanks all for feedback, I think I've addressed everything:

- `clear()` -> `unset()`
- `getvals()` -> `_getvals()`
  - Note this is a `Schema` method only that is used by `Chip`, so it's not great practice to make it private. However, we're already doing this in other cases so I'm okay with it for the sake of moving forward. We can revisit later perhaps!
- Add docstring to `_is_set()`

aolofsson (approved):

PR #1256, opened 2023-02-02
Refactor metrics group to use step/index abstraction

nmoroze:
This PR depends on PR #1255.

This PR refactors the entire metrics group to use our new step/index abstraction. These keypaths are marked as required per-node values, so a user must supply step/index kwargs when accessing them.

This is mostly a straightforward copy-replace, but I also eliminated some dead code, including a stale example and an unused `list_metrics()` function that was part of the public API and seems unnecessary now that iterating through the metrics is as simple as `chip.getkeys('metrics')`.


siliconcompiler/tools/openroad/openroad.py
```diff
@@ -345,7 +345,7 @@ def post_process(chip):
                     drvs += int(metrics[metric])
 
         if drvs is not None:
-            chip.set('metric', step, index, 'drvs', str(drvs), clobber=True)
+            chip.set('metric', 'drvs', str(drvs), clobber=True, step=step, index=index)
```
gadfort, line 348: just for readability, it might be good to swap step/index and clobber

gadfort (approved):
LGTM, just take a look at my comment.

examples/oh_experiments/parallel_adder_sweep.py
```diff
@@ -1,60 +0,0 @@
-# Copyright 2020 Silicon Compiler Authors. All Rights Reserved.
-
```
aolofsson, line 2: Motivation for deleting example?

aolofsson (changes requested):
The example deleted is a potential use case, unless we can no longer support it?

examples/oh_experiments/parallel_adder_sweep.py
nmoroze: The example has been broken for a long time (initializes Chip object without design) and isn't referenced in docs/CI. 

I'm also not sure it adds much since we already have a multiprocessing example (examples/heartbeat/parallel.py) and a sweep example (examples/oh_experiments/adder_sweep.py) that are both up-to-date and tested.

nmoroze: Although we could rewrite this very nicely using a multi-index approach once we have ['option', 'param'] be overridable on a step/index basis.

aolofsson: Yes, you are right, the heartbeat/parallel.py illustrates the pattern I was looking for.

aolofsson (approved):

siliconcompiler/tools/openroad/openroad.py
nmoroze: Nice catch, ended up getting rid of the `clobber=True` on the metrics from here and in Yosys driver since it's unnecessary (default is True, plus metrics shouldn't be set by anything else-don't think the explicit setting communicates much).

PR #1261, opened 2023-02-02
Early fixes for post-refactor remote client

WRansohoff:
A few small SiliconCompiler updates which are needed to get the minimal `heartbeat` example running remotely with our recent `Chip`/`Schema` updates.

I like the new `Schema.unset` method - it's more clear than overwriting the value to `None` or `[]`.

There might be a couple more PRs like this as I move on to testing the server-side scheduler functionality.

### DIFF
--- a/siliconcompiler/client.py
+++ b/siliconcompiler/client.py
@@ -37,7 +37,7 @@ def get_base_url(chip):
     return remote_protocol + remote_host
 
 ###################################
-def remote_preprocess(chip):
+def remote_preprocess(chip, steplist):
     '''Helper method to run a local import stage for remote jobs.
     '''
 
@@ -48,7 +48,7 @@ def remote_preprocess(chip):
 
     # Fetch a list of 'import' steps, and make sure they're all at the start of the flow.
     flow = chip.get('option', 'flow')
-    remote_steplist = chip.getkeys('flowgraph', flow)
+    remote_steplist = steplist.copy()
     import_tasks = chip._get_steps_by_task()['import']
     import_steps = []
     for task_tuple in import_tasks:
--- a/siliconcompiler/client.py
+++ b/siliconcompiler/client.py
@@ -46,17 +46,31 @@ def remote_preprocess(chip):
         job_hash = uuid.uuid4().hex
         chip.status['jobhash'] = job_hash
 
-    # Setup up tools for all local functions
+    # Fetch a list of 'import' steps, and make sure they're all at the start of the flow.
     flow = chip.get('option', 'flow')
     remote_steplist = chip.getkeys('flowgraph', flow)
-    local_step = remote_steplist[0]
-    indexlist = chip.getkeys('flowgraph', flow, local_step)
-    for index in indexlist:
-        task = chip._get_task(local_step, index)
-        tool = chip.get('flowgraph', flow, local_step, index, 'tool')
-        # Setting up tool is optional (step may be a builtin function)
-        if tool and tool not in chip.builtin:
-            chip._setup_tool(tool, task, local_step, index)
+    import_tasks = chip._get_steps_by_task()['import']
+    import_steps = []
+    for task_tuple in import_tasks:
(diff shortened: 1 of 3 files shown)


siliconcompiler/client.py
```diff
@@ -52,10 +52,11 @@ def remote_preprocess(chip):
     local_step = remote_steplist[0]
     indexlist = chip.getkeys('flowgraph', flow, local_step)
     for index in indexlist:
+        task = chip._get_task(local_step, index)
         tool = chip.get('flowgraph', flow, local_step, index, 'tool')
         # Setting up tool is optional (step may be a builtin function)
         if tool and tool not in chip.builtin:
-            chip._setup_tool(tool, local_step, index)
+            chip._setup_tool(tool, task, local_step, index)
```
gadfort, line 59: I suggest using: https://github.com/siliconcompiler/siliconcompiler/blob/b0d36a2a5f6013dd7c4a185837114c2436c09b6b/siliconcompiler/core.py#L4534
Otherwise this builtin check is not correct

gadfort (changes requested):

siliconcompiler/client.py
WRansohoff: Hmm...this was written when a built-in `import` step seemed like a possibility, but I'm not sure if that is in the cards anymore.

Really, this pre-process method is intended to run `import` and then send the collected/linted inputs off for processing. Considering where we are now, it might make more sense to do something like error out if the task name is not `import`?

gadfort: you can use `_get_steps_by_task()['import']` to get the import steps (name of the step doesn't matter, but the task is import)

WRansohoff: Thanks, that's probably a good way to do it. I'm a little uneasy with the assumption that `import` tasks will always exist as the first N steps in a flow, but I guess I can add a check for that corner case.

nmoroze: I don't see a problem using the new built-in check -- I think built-in import tasks are common (I've written lots of flows with a nop import), and the check is just here because builtins don't have a setup() method, right?

Adding an additional check that the first step is an "import" task seems fine, but not necessarily related to the builtin check?

siliconcompiler/core.py
```diff
@@ -4008,6 +4008,8 @@ def run(self):
                 local_dir = self.get('option','builddir')
                 self.read_manifest(cfg, clobber=True, clear=True)
                 self.set('option', 'builddir', local_dir)
+                # Un-set steplist so 'show'/etc flows will work on returned results.
+                self.unset('option', 'steplist')
```
nmoroze, line 4012: Is this necessarily something we want to do? I think if the user sets `['option', 'steplist']`, the setting should persist after `run()` is called. Perhaps a better thing would be to store whether steplist has been set/its value before calling `remote_preprocess()` and then restore the state of the parameter at the end (unset() if it wasn't set, otherwise reset it to the value).

WRansohoff: This is all under an `if self.get('option','remote'):` block, so it won't get run for local builds with user-defined steplists.

I suppose it's an open question as to whether we should let people define a subset of steps to run in the remote flow. It might be useful if people want to experiment with early steps like floorplan/placement.

nmoroze: Got it - I forgot that steplists aren't supported with the remote run at the moment. This seems fine then!

siliconcompiler/client.py
WRansohoff: I'm just thinking through a more extensible remote workflow. It seems worthwhile to update the overall logic which finds and runs the `import` tasks, since I'm already fixing the syntax and we have more capable helper methods in the `Chip` object now.

gadfort (approved):
LGTM

aolofsson:
Very nice! Given the amount of changes made, it's great to see that the number of changes are limited.

PR #1263, opened 2023-02-02
Implementation of .use()

gadfort:
Changes:
- Implements .use() for PDKs/Libraries/Flows/Checklists
- both `.use` and `.load_target` can take a `**kwargs` to pass long parameters needed to configure flows, etc
- removal of ['arg', 'flow'/'pdk'] from schema'
- Updates docs to build with the new return types

Notes:
- many of the files in flows/, libs/ pdks/, targets/, and checklists/ are mostly modified to account for the change in use model, but should not have any functional changes.

### DIFF
--- a/siliconcompiler/__init__.py
+++ b/siliconcompiler/__init__.py
@@ -1,11 +1,5 @@
 from siliconcompiler.core import *
-#from siliconcompiler.floorplan import *
-#from siliconcompiler.leflib import *
-#from siliconcompiler.deflib import *
 
 from siliconcompiler._metadata import version as __version__
 
-from siliconcompiler.use import PDK
-from siliconcompiler.use import Library
-from siliconcompiler.use import Flow
-from siliconcompiler.use import Checklist
+from siliconcompiler.use import PDK, Library, Flow, Checklist
--- a/siliconcompiler/__init__.py
+++ b/siliconcompiler/__init__.py
@@ -4,3 +4,8 @@ from siliconcompiler.core import *
 #from siliconcompiler.deflib import *
 
 from siliconcompiler._metadata import version as __version__
+
+from siliconcompiler.use import PDK
+from siliconcompiler.use import Library
+from siliconcompiler.use import Flow
+from siliconcompiler.use import Checklist
--- a/siliconcompiler/checklists/oh_tapeout/oh_tapeout.py
+++ b/siliconcompiler/checklists/oh_tapeout/oh_tapeout.py
@@ -1,35 +1,40 @@
 import siliconcompiler
 
+from siliconcompiler import Checklist
+
 def make_docs():
     '''Subset of OH! library tapeout checklist.
 
     https://github.com/aolofsson/oh/blob/master/docs/tapeout_checklist.md
     '''
     chip = siliconcompiler.Chip()
-    setup(chip)
-    return chip
+    return setup(chip)
 
 def setup(chip):
     standard = 'oh_tapeout'
 
+    checklist = Checklist(chip, standard)
+
     # Automated
-    chip.set('checklist', standard, 'drc_clean', 'description',
-             'Is block DRC clean?')
-    chip.set('checklist', standard, 'drc_clean', 'criteria', 'drvs==0')
+    checklist.set('checklist', standard, 'drc_clean', 'description',
+                  'Is block DRC clean?')
+    checklist.set('checklist', standard, 'drc_clean', 'criteria', 'drvs==0')
 
-    chip.set('checklist', standard, 'lvs_clean', 'description',
-             'Is block LVS clean?')
-    chip.set('checklist', standard, 'lvs_clean', 'criteria', 'drvs==0')
+    checklist.set('checklist', standard, 'lvs_clean', 'description',
+                  'Is block LVS clean?')
+    checklist.set('checklist', standard, 'lvs_clean', 'criteria', 'drvs==0')
 
-    chip.set('checklist', standard, 'setup_time', 'description',
-             'Setup time met?')
-    chip.set('checklist', standard, 'setup_time', 'criteria', 'setupslack>=0')
+    checklist.set('checklist', standard, 'setup_time', 'description',
+                  'Setup time met?')
+    checklist.set('checklist', standard, 'setup_time', 'criteria', 'setupslack>=0')
 
-    chip.set('checklist', standard, 'errors_warnings', 'description',
-             'Are all EDA warnings/errors acceptable?')
-    chip.set('checklist', standard, 'errors_warnings', 'criteria',
-              ['errors==0', 'warnings==0'])
+    checklist.set('checklist', standard, 'errors_warnings', 'description',
+                  'Are all EDA warnings/errors acceptable?')
+    checklist.set('checklist', standard, 'errors_warnings', 'criteria',
+                  ['errors==0', 'warnings==0'])
 
     # Manual
-    chip.set('checklist', standard, 'spec', 'description',
-        'Is there a written specification?')
+    checklist.set('checklist', standard, 'spec', 'description',
+                  'Is there a written specification?')
+
+    return checklist
--- a/siliconcompiler/checklists/oh_tapeout/oh_tapeout.py
+++ b/siliconcompiler/checklists/oh_tapeout/oh_tapeout.py
@@ -1,7 +1,5 @@
 import siliconcompiler
 
-from siliconcompiler import Checklist
-
 def make_docs():
     '''Subset of OH! library tapeout checklist.
 
@@ -13,7 +11,7 @@ def make_docs():
 def setup(chip):
     standard = 'oh_tapeout'
 
-    checklist = Checklist(chip, standard)
+    checklist = siliconcompiler.Checklist(chip, standard)
 
     # Automated
     checklist.set('checklist', standard, 'drc_clean', 'description',
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -85,10 +85,6 @@ class Chip:
             self.error("""SiliconCompiler must be run from a directory that exists.
 If you are sure that your working directory is valid, try running `cd $(pwd)`.""", fatal=True)
 
-        # Indicates that the chip object is marked as read only and should not
-        # allow modifications to schema
-        self._readonly = False
-
         self.schema = Schema()
 
         # The 'status' dictionary can be used to store ephemeral config values.
@@ -593,14 +589,8 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         from siliconcompiler import Library
         from siliconcompiler import Checklist
 
-        # Mark self as read only
-        self._readonly = True
-
         # Call the module setup function.
         use_modules = module.setup(self, **kwargs)
-
-        # Reset readonly flag
-        self._readonly = False
 
         # Make it a list for consistency
         if not isinstance(use_modules, list):
@@ -873,9 +863,6 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         if keypath == ['option', 'loglevel'] and field == 'value':
             self.logger.setLevel(value)
 
-        if self._readonly:
-            raise SiliconCompilerError("Chip object is read only.")
-
         try:
             if not self.schema.set(
                 *keypath, value, field=field, clobber=clobber, step=step, index=index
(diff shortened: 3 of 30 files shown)


siliconcompiler/core.py
```diff
@@ -563,40 +567,84 @@ def load_target(self, name):
...
         '''
         Loads a SiliconCompiler module into the current chip object by calling
         a module.setup() method.
         '''
 
+        # Load supported types here to avoid cyclic import
+        from siliconcompiler import PDK
+        from siliconcompiler import Flow
+        from siliconcompiler import Library
+        from siliconcompiler import Checklist
```
nmoroze, line 594: Seems funky, what exactly is causing this issue? (is it related to the imports in __init__.py that bring things into the module scope?)

siliconcompiler/core.py
```diff
@@ -825,6 +873,9 @@ def set(self, *args, field='value', clobber=True, step=None, index=None):
         if keypath == ['option', 'loglevel'] and field == 'value':
             self.logger.setLevel(value)
 
+        if self._readonly:
+            raise SiliconCompilerError("Chip object is read only.")
+
```
nmoroze, line 878: Should this be raised above the loglevel block?

nmoroze (commented):
Nice change! This seems pretty clean.

A few higher-level questions - I think a lot of this got hashed out while I was away, sorry if this is repeating discussions that were already had:

1: Why did we go for inheriting from Chip rather than Schema? Is it logger related? One benefit of the Schema object is that it can be initialized with a subset of the schema, preventing someone from setting keypaths outside a certain group. It seems like as-is, the user can set any schema parameters they want, and then the ones outside of the proper group would be silently dropped (I could see that causing confusion).

If the main issue is the logger, I think that could be passed in on its own and supplied to the Schema obj (which we'll want to do anyways): 

```python
from schema_cfg import schema_pdk

class PDK(Schema):
   def __init__(self, name, logger=None):
        super().__init__(cfg=schema_pdk(name), logger=logger)
```

2: What is the purpose of the read-only Chip object? Are we accepting dependence on values already set in the Chip? I think it would be ideal to move away from that and rely on kwargs to `setup()` instead.

WRansohoff:
Re: Noah's first question, I think we were concerned about asking people to understand two very similar objects: `Chip` and `Schema`. It's probably easier to only ask that people know how to use the `Chip` (and now `PDK`/etc) pass-through methods.

+1 to his second question: it seems like we might be able to get by with either `kwargs` or a read-only `Chip` in `setup`, but not both.

siliconcompiler/core.py
WRansohoff: I think the custom `__init__.py` causes circular import issues if you try to use anything that it imports (or `import siliconcompiler`) within `core.py`.

You might be able to put something like this at the top of the file instead:

`from siliconcompiler.use import PDK, Flow, Library, Checklist`

siliconcompiler/core.py
```diff
@@ -563,40 +567,84 @@ def load_target(self, name):
...
-        elif 'libs.' in module.__name__:
-            lib_name = module.__name__[module.__name__.rfind('.')+1:]
-            self._loaded_modules['libs'].append(lib_name)
-            # TODO: Stdcell libs may call 'import_library' of their own volition, and not return a Chip.
-            if new_schema:
-                self.import_library(module.setup(self))
+        use_modules = module.setup(self, **kwargs)
+
+        # Reset readonly flag
+        self._readonly = False
```
WRansohoff, line 603: Could the `PDK.setup` methods also call this to modify the parent `Chip`, if a package author got frustrated with the limitation?

nmoroze:
@WRansohoff I see, that makes sense. That being said, given the `Schema` object has an identical API for all the relevant methods, it doesn't seem like a huge burden (and if they're not identical, that could be fixed). It also seems a little odd to me that things like `PDK.run()` exist. Even though it would just fail out, I like the idea of being more restrictive with the interface.

I guess there's the question of higher-level helper functions like `node()` and `edge()`. On the one hand, we could make those members of `Flow`, which seems like a nice way to enforce some more organization. On the other hand, I guess that breaks our ability to easily hack up a flow on a Chip object within a build script.

WRansohoff:
I wouldn't mind revisiting the `Schema` sub-class idea, with `Flow` owning (possibly static) methods like `node` and `edge`. That could let us pull a bit more code out of the large `core.py` file, and the `get`/`set`/`add` interface is probably what matters most to the user.

But I think there may have been other reasons to prefer `Chip` as a super-class - it might merit a more in-depth discussion.

aolofsson:
@nmoroze @gadfort @WRansohoff We should take this discussion off line tomorrow. I shared a spreadsheet with all the public api methods to give an overview of where different methods might be useful. As a HW guy, my SW instincts are questionable, but I would think that we could come up with a clean class architecture with modest refactoring...

aolofsson:
What do you think about the mockup below for a Class structure? The schema is the base class used by all, chip is the complete superset, and pdk/flow/others have a small number of functions to avoid repeating code. Feels like an organized and natural structure based on the schema and api functions, let me know...apologies if this was already suggested and I didn't get the idea. 

```python
class Schema:
    def set(self,a):
        print(f"Calling set({a}).")

class Flow(Schema):
    def __init__(self):
        self.flowgraph={"hello":"flow"}
    def node(self, a):
        print(f"Calling node({a})...")
    def edge(self, a):
        print(f"Calling node({a})...")
    def pipe(self, a):
        print(f"Calling node({a})...")

class PDK(Schema):
    def __init__(self):
        self.pdk={"hello":"pdk"}
    pass

class Chip(Flow, PDK, Schema):
    def __init__(self):
        self.cfg={"hello":"chip"}
    def run(self,a):
        print(f"Calling run({a})...")

# Creating Different Objects

f = Flow()
f.set("from flow")
f.node("from flow")
print(f.flowgraph)

p = PDK()
p.set("from PDK")
print(p.pdk)

c = Chip()
c.set("from chip")
c.node("from chip")
c.run("from chip")
print(c.cfg)


```

siliconcompiler/core.py
gadfort: @WRansohoff is right. I can try to see if there is a good method to avoid this.

siliconcompiler/core.py
gadfort: I'm not sure I follow about the `PDK.setup`?

WRansohoff: I mean, is there anything that makes the `_readonly` value immutable, or could a PDK module author call `chip._readonly = False` in the PDK `setup()` method, and then call `chip.set(...)`?

gadfort: Ahh, yes, for the same reason that anyone could do `chip.schema.cfg` that would only really be private if it was `self.__readonly`

nmoroze (approved):
After offline discussion today, happy to merge! Just 1 minor comment RE: the location of the guard in chip.set() still.

In follow-on PRs, let's revisit implementation/where the keypaths are rooted/chip vs logger in the setup() signature.

siliconcompiler/use.py
```diff
@@ -0,0 +1,79 @@
+# Copyright 2023 Silicon Compiler Authors. All Rights Reserved.
+
+from siliconcompiler import Chip
+
+class PDK(Chip):
```
WRansohoff, line 5: I think it would be a good idea to make these `Schema` sub-classes, especially if we want them to eventually be persistent within the `Chip` object. But like we discussed, that can come in a future PR because of the complications with `Flow.node` / `Flow.edge`.

siliconcompiler/__init__.py
```diff
@@ -4,3 +4,8 @@
 #from siliconcompiler.deflib import *
 
 from siliconcompiler._metadata import version as __version__
+
+from siliconcompiler.use import PDK
```
WRansohoff, line 8: Nit: these can probably go on one line: `from siliconcompiler.use import Checklist, Flow, Library, PDK`

siliconcompiler/checklists/oh_tapeout/oh_tapeout.py
```diff
@@ -1,35 +1,40 @@
...
     '''
     chip = siliconcompiler.Chip()
-    setup(chip)
-    return chip
+    return setup(chip)
 
 def setup(chip):
     standard = 'oh_tapeout'
 
+    checklist = Checklist(chip, standard)
```
WRansohoff, line 16: Does `siliconcompiler.Checklist` work without the `from siliconcompiler import Checklist` from above? Ditto for Flow/PDK/Library modules.

siliconcompiler/core.py
```diff
@@ -85,6 +85,10 @@ def __init__(self, design, loglevel=None):
             self.error("""SiliconCompiler must be run from a directory that exists.
 If you are sure that your working directory is valid, try running `cd $(pwd)`.""", fatal=True)
 
+        # Indicates that the chip object is marked as read only and should not
+        # allow modifications to schema
+        self._readonly = False
```
WRansohoff, line 90: Could we remove the attribute-based `_readonly` flag, and leave the `Chip` object mutability in place until we can land on a read-only Chip implementation (/decide if we want one)?

siliconcompiler/targets/asap7_demo.py
```diff
@@ -21,7 +21,7 @@ def setup(chip):
     from flows import asicflow
     from libs import asap7sc7p5t
     chip.use(asap7)
-    chip.use(asicflow)
+    chip.use(asicflow, syn_np=syn_np, floorplan_np=floorplan_np, physyn_np=physyn_np, place_np=place_np, cts_np=cts_np, route_np=route_np, dfm_np=dfm_np)
```
WRansohoff, line 24: Nit: Multi-line? Also in freepdk45/skywater130 targets

WRansohoff (changes requested):
Looks good to me overall!

Most of these comments are small and quick, and we can punt on the `Chip`/`Schema` super-class changes. I do think it might be a good idea to remove the `_readonly` flag until we're sure how/if that should be implemented.

siliconcompiler/targets/asap7_demo.py
aolofsson, line 24: With such a large number of arguments, clarity argument goes away. For something like this, a dictionary seems more natural. Something to think about...

aolofsson (approved):
LGTM! (with the caveat that class arch change needs to be implemented at some point)

siliconcompiler/targets/asap7_demo.py
gadfort: I can convert this to be a dict that is expanded and passed in so it's a much shorter line to read.

siliconcompiler/core.py
gadfort: Sounds reasonable. With the larger refactor making a ReadOnly schema would be easy

siliconcompiler/checklists/oh_tapeout/oh_tapeout.py
gadfort: it should, I can update them to do `siliconcompiler.???`

WRansohoff: Up to you - it's one less line, but the only important change here is the `_readonly` attribute.

siliconcompiler/targets/asap7_demo.py
WRansohoff: I was considering proposing a clever solution to this, but I'm not sure if it would be worth it. You could probably use `inspect.signature`:

```
import inspect

def setup(chip, syn_np=1, floorplan_np=1, physyn_np=1, place_np=1, cts_np=1, route_np=1, dfm_np=1):
    for arg in inspect.signature(setup).parameters:
        if arg != 'chip':
            kwargs[arg] = locals()[arg]
    [...]
    chip.use(asicflow, **kwargs)
```

But again, I think it's probably too complex to save a very small amount of repetition.

WRansohoff (approved):
Looks good! Maybe let the tests finish before hitting merge, just in case

siliconcompiler/targets/asap7_demo.py
gadfort: I think we can explore that in a separate PR if it's worth it, but I would worry that users looking at these modules will not be able to follow easily.

gadfort:
> Looks good! Maybe let the tests finish before hitting merge, just in case

@WRansohoff thanks. I also ran it earlier on the daily CI (which didn't show any problems), so hopefully nothing new comes up.

siliconcompiler/targets/asap7_demo.py
WRansohoff: Yeah, and it would be hard to debug later.

PR #1265, opened 2023-02-03
Support partial steplists in remote flow.

WRansohoff:
Like @nmoroze suggested in #1261, there's not much reason to exclude custom `steplist` parameters from the remote flow. It may not be useful for generic ASIC / bitstream builds, but it could help with design exploration, and the functionality is already built into `Chip.run` if we pass it through.

This could also work with a `from` / `to` specification, if we implement that.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3989,7 +3989,10 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                 self.error('Cannot pass "-step" parameter into remote flow. A steplist including '
                            'an "import" task and at least one EDA task is required.',
                            fatal=True)
-            pre_remote_steplist = self.get('option', 'steplist')
+            pre_remote_steplist = {
+                'steplist': self.get('option', 'steplist'),
+                'set': self.get('option', 'steplist', field='set'),
+            }
             remote_preprocess(self, steplist)
 
             # Run the job on the remote server, and wait for it to finish.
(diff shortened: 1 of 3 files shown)


aolofsson:
Do you have a test for this functionality, or will that have to wait. Since this is new functionality, I think it should come with a test.

WRansohoff:
Yes, you're right. I had only done a quick manual test, and in the process of adding CI tests, I found a small bug.

tests/flows/test_server.py
```diff
@@ -55,8 +56,264 @@ def mocked_runtask(*args, **kwargs):
...
+
+    # Ensure that klayout doesn't open its GUI after results are retrieved.
+    os.environ['DISPLAY'] = ''
+
+    # Create the temporary credentials file, and set the Chip to use it.
+    tmp_creds = '.test_remote_cfg'
+    with open(tmp_creds, 'w') as tmp_cred_file:
+        tmp_cred_file.write(json.dumps({'address': 'localhost', 'port': 8081}))
+    gcd_chip.set('option', 'remote', True)
+    gcd_chip.set('option', 'credentials', os.path.abspath(tmp_creds))
```
nmoroze, line 95: Using a local test fixture might be helpful here to reduce some of this duplicated setup (and improve the clarity of the tests).

I think this can wait for a future PR next time you happen to touch this file, though.

nmoroze (commented):
Nice to have this feature! Just one thing I noticed regarding preserving the steplist setting, left a comment inline.

siliconcompiler/core.py
```diff
@@ -4000,7 +4005,7 @@ def run(self):
                 self.read_manifest(cfg, clobber=True, clear=True)
                 self.set('option', 'builddir', local_dir)
                 # Un-set steplist so 'show'/etc flows will work on returned results.
-                self.unset('option', 'steplist')
+                self.set('option', 'steplist', pre_remote_steplist)
```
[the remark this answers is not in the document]
WRansohoff: Good point, thanks!

tests/flows/test_server.py
WRansohoff: That's a good idea - and it's probably better to do it now. I think we should try to be more proactive about test hygiene with the higher volume of changes that we're dealing with.

nmoroze (approved):
Looks great, thanks!

PR #1266, opened 2023-02-03
Make record schema use step/index

nmoroze:
This PR switches over the record group to use the new step/index scheme. Also took the chance to fix a few set(..., None)s to unset()s where I noticed them.


gadfort (approved):
LGTM

PR #1267, opened 2023-02-03
openroad constraint pins/partname

gadfort:
Changes:
- Adds pin constraint support (initial), still needs to be refined
- Support for partname in component placement

### DIFF
--- a/siliconcompiler/tools/openroad/sc_floorplan.tcl
+++ b/siliconcompiler/tools/openroad/sc_floorplan.tcl
@@ -72,23 +72,119 @@ if { 0 } {
 # Pin placement
 ###########################
 
-if {[dict exists $sc_cfg constraints pin]} {
-  foreach name [dict get $sc_cfg constraints pin] {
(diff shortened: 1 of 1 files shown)


siliconcompiler/tools/openroad/sc_floorplan.tcl
```diff
@@ -56,49 +56,155 @@ if {[dict exists $sc_cfg pdk $sc_pdk aprtech openroad $sc_stackup $sc_libtype tr
   make_tracks
 }
 
+set do_automatic_pins 1
 if { 0 } {
```
WRansohoff, line 60: Is `if { 0 }` different from `if False:`?

gadfort: It's the same in tcl. The `if {0}` is a placeholder for the padring code to make sure it doens't get included for now

aolofsson (approved):
Encouraging. Looks like the new schema constraints fit in pretty naturally.

PR #1268, opened 2023-02-03
openroad: enable export with SPEF

gadfort:
Changes:
- modifies asicflow to split export into two indexes, with klayout handling the gds generation, and openroad generating SPEF and timing libraries
- Adds export task to openroad


aolofsson:
- Can you add some documentation to the code (I had to read the PR comments to understand what was going on.
- Naming/consistency. I could imagine two paths: 1.) export = def, gds, spef, lib, lef , or 2.) specific tasks for pex, abstract (bad name), etc. Probably easiest to just add comment for 1.) for now.

gadfort:
@aolofsson the naming consistency is a problem with things like the .lib and .spef because they can't all be named `??.lib` and we need a way to differentiate between them if we want to bring them back together into a single step.
I settled on spef getting the pexcorner appended (that seemed natural) and the lib getting the libcorner appended.

aolofsson:
My question was at a higher level. By export we mea: siliconcompiler/tools/openroad/sc_export.tcl

ie. takin in odb (or def?) and write out lef, spef, lib, cdl 

Is my understanding correct?

gadfort:
Yes, that is correct. If export is reserved for GDS then we can rename it, but it makes sense to me since this is the information needed to build a hierarchical flow

aolofsson (approved):

aolofsson:
Makes sense to me

PR #1269, opened 2023-02-03
fix daily ci

gadfort:
Fixes failing tests on daily ci. (one is disabled until routing can be resolved).


gadfort:
@WRansohoff I turned off the RAM example, it's not passing routing, so it's either too sensitive to the OpenROAD version or there is a different issue. We can turn it back on once thats been sorted out.

WRansohoff (approved):

nmoroze (approved):
LGTM, gonna merge so I can rebase my KL fix onto it and see if that gets daily CI green.

PR #1270, opened 2023-02-03
klayout: import Schema directly to avoid dependencies

nmoroze:
- Need to refactor Schema a bit to avoid dependencies elsewhere in SC project
- Need to pass path via CLI options so that it can be resolved at runtime (otherwise remote runs won't work)

testing with manual daily CI run: https://github.com/siliconcompiler/siliconcompiler/actions/runs/4086206108

### DIFF
--- a/siliconcompiler/schema/utils.py
+++ b/siliconcompiler/schema/utils.py
@@ -1,0 +1,83 @@
+# Copyright 2023 Silicon Compiler Authors. All Rights Reserved.
+
+# NOTE: this file cannot rely on any third-party dependencies, including other
+# SC dependencies outside of its directory, since it may be used by tool drivers
+# that have isolated Python environments.
+
+import os
+import re
+import sys
+
+PACKAGE_ROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), '..')
+
+def escape_val_tcl(val, typestr):
+    '''Recursive helper function for converting Python values to safe TCL
+    values, based on the SC type string.'''
+    if val is None:
+        return ''
+    elif typestr.startswith('('):
+        # Recurse into each item of tuple
+        subtypes = typestr.strip('()').split(',')
+        valstr = ' '.join(escape_val_tcl(v, subtype.strip())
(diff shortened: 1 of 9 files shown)


nmoroze:
Alright, looks like this is working properly: https://github.com/siliconcompiler/siliconcompiler/actions/runs/4109065559. 

I had to modify the Daily CI tests a bit - the way that setup-python was modifying the environment (particularly, the fact that it sets `LD_LIBRARY_PATH`) would cause KLayout to run a Python 3.8.12 interpreter, while instantiating sys.path to the system-installed Python 3.8.10. This happened to break importing the built-in CSV module. After talking offline with @gadfort, it seems like fixing this issue in the test runner is appropriate, since it's a weird Python setup that probably wouldn't occur on a user's machine.

gadfort (approved):
LGTM. Thanks for taking a look

siliconcompiler/schema/utils.py
```diff
@@ -0,0 +1,83 @@
...
+
+    This is important for properly parsing complex RST in our docs.
+
+    Source:
+    https://www.python.org/dev/peps/pep-0257/#handling-docstring-indentation'''
+    if not docstring:
+        return ''
+    # Convert tabs to spaces (following the normal Python rules)
+    # and split into a list of lines:
+    lines = docstring.expandtabs().splitlines()
```
WRansohoff, line 65: If you call `.strip()` before `.splitlines()`, you might be able to skip the logic below which removes leading/trailing newlines

WRansohoff (commented):
Sorry, I'm a bit behind on code reviews. Looks like a good solution!

nmoroze:
No worries, thanks for taking a look! Sorry I jumped the gun on merging.

siliconcompiler/schema/utils.py
nmoroze: Seems like a good catch - I'll probably keep the logic as-is since it came verbatim from the linked PEP, though!

PR #1271, opened 2023-02-03
Fix small `Schema` bug: `set` field is not processed in `merge_manifest`

WRansohoff:
(Edit: I was wrong about the root cause described below)

If we write out a Schema with `prune=True`, and then re-read that manifest file to re-create a Chip object, the `defvalue` keys may get populated with `None` in the new schema dictionary.

This can cause confusing errors if a Schema value has not been changed from its default settings: in my case, `chip.get('tool', 'openroad', 'exe')` was returning `None` because the default value of `openroad` had not been changed, and that caused odd errors in the `_runtask` method.

@nmoroze: Does this seem like a reasonable solution, or can you think of cases where we might not want to return the contents of `value` if `set == False`?


WRansohoff:
It sounds like the root cause may not be as simple as I'd thought. I'll try to make a minimal test case which demonstrates the bug.

WRansohoff:
Okay, it looks like the real issue is that `set` isn't included in the list of fields to merge during the `merge_manifest` function. So when we merge a parameter containing:

`set: True, value: 'something', defvalue: None`

The merged parameter ends up containing:

`set: False, value: 'something', defvalue: None`

The `Schema` class then returns a value of `None` for the parameter, because the `set` field is `False`.

nmoroze (approved):
LGTM, thanks for working through this with me offline! This should prevent parameters with 'set'=False from overriding other parameters when merging the schema in. 

I'd still be curious to know what the cause of the OpenROAD exe parameter's 'set' flag going False in the Yosys JSON dump is, but this should definitely get merged!

WRansohoff:
Thanks - and I'm curious about why the failure is so hard to reproduce, too!

I would really like a test case for this - I have one more idea in mind, so I might wait until tomorrow to merge this.

WRansohoff:
Okay, I think that we can actually use the PicoRV32 + RAM example to test this! It passes with the changes you suggested, and fails if I remove `'set'` from the list of keys to ignore in `_merge_macro`.

I do still need to submit some changes to upload the imported `Library`'s GDS/LEF files with the job, and support for that may not make it into the minimal dev server immediately. But these pytests are run on the same machine over `localhost`, so I think this gives us a way to reproduce the bug locally.

PR #1272, opened 2023-02-03
remove relative imports as those break the docs generation

gadfort:
Changes:
- No functional changes, this just removes relative imports as these are not supported in the docs generation.

This PR is a step towards generating task specific documentation for the tools.


WRansohoff (approved):
Ah, whoops. Maybe relative imports were a bad idea, they are less precise than absolute ones.

Thanks for cleaning up the tool imports, too!

PR #1274, opened 2023-02-06
Daily test fixes

WRansohoff:
* Increase picorv32 + RAM example's die height by 100um so that it consistently passes routing.
* Fix typo in `sc_floorplan.tcl`.
* Delete duplicate `picorv32.v` file.
* Update heartbeat w/ padring test for the new `Chip.use` implementation.

For better or worse, the `Chip.use` method can accept anything with a valid `setup` method. Since the `heartbeat_padring` example only sets three parameters in its `Library` object, we can define an inline class for the macro instead of creating a new module/file for four lines of code.

It might not be a bad time to discuss adding a test coverage requirement for new changes, or a process which ensures that new changes are run before they are submitted for review. It shouldn't be too difficult to add a coverage report to our CI suite, and I know that I sometimes need the reminder when I'm excited to push a new feature.

The full test suite is [currently only passing on Python 3.8](https://github.com/siliconcompiler/siliconcompiler/actions/runs/4099471916/jobs/7069411923). I believe that's because of a version conflict triggered by importing siliconcompiler within KLayout's bundled Python interpreter, which Noah figured out and is investigating.


gadfort (approved):
Please wait to merge until github actions are green again

PR #1275, opened 2023-02-06
Minor docs fixes

gadfort:
Changes:
- Repairs the target references to libraries that broke after switching to `.Library`
- Fix openroad links in docs
- Add reference to the PDK associated with the skywater130 pdk


nmoroze (approved):
LGTM, thanks!

PR #1277, opened 2023-02-07
Add tasks to documentation

gadfort:
Adds:
- initial support for loading the tasks into the documentation. This simply adds a subsection to each tool with one section per task.

The tasks are currently not well documented, so next we need to fix the doc strings there.


nmoroze (approved):
LGTM, thanks for delving into the gnarliness of the docs generation!

PR #1279, opened 2023-02-07
Update `sc-configure`, and add tests

WRansohoff:
* Adds a simple one-line configuration for public servers (`sc-configure https://...`)
* Omits the `username` and `password` fields in `client.py` if either one is an empty string.
* Adds some tests for the `sc-configure` app.

### DIFF
--- a/siliconcompiler/apps/sc_configure.py
+++ b/siliconcompiler/apps/sc_configure.py
@@ -33,10 +33,17 @@ def main():
             elif (oin == 'y') or (oin == 'Y'):
                 overwrite = True
 
-    # Get parameters from user input.
-    srv_addr = input('Remote server address: ').replace(" ","")
-    username = input('Remote username: ').replace(" ","")
-    user_pass = input('Remote password: ').replace(" ","")
+    # If a command-line argument is passed in, use that as a public server address.
+    if len(sys.argv) > 1:
+        print(f'Creating remote configuration file for public server: {sys.argv[1]}')
+        with open(cfg_file, 'w') as f:
+            f.write('{"address": "%s"}'%(sys.argv[1]))
+        return
+
+    # If no arguments were passed in, interactively request credentials from the user.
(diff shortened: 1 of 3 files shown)


tests/apps/test_sc_configure.py
```diff
@@ -0,0 +1,108 @@
...
+    server_name = 'https://example.com'
+    username = 'ci_test_user'
+    password = 'ci_test_password'
+    monkeypatch.setattr('sys.argv', ['sc-configure'])
+
+    # Use sys.stdin to simulate user input.
+    with open('cfg_stdin.txt', 'w') as wf:
+         wf.write(f'y\n{server_name}\n{username}\n{password}\n')
+    with open('cfg_stdin.txt', 'r') as rf:
+         sys.stdin = rf
```
nmoroze, line 68: Could something weird happen from overwriting `sys.stdin` without restoring it? If so, maybe there's a way to patch this safely.

siliconcompiler/apps/sc_configure.py
```diff
@@ -33,10 +33,17 @@ def main():
...
+    if len(sys.argv) > 1:
+        print(f'Creating remote configuration file for public server: {sys.argv[1]}')
+        with open(cfg_file, 'w') as f:
+            f.write('{"address": "%s"}'%(sys.argv[1]))
+        return
+
+    # If no arguments were passed in, interactively request credentials from the user.
+    srv_addr = input('Remote server address:\n').replace(" ","")
+    username = input('Remote username (leave blank for public servers):\n').replace(" ","")
+    user_pass = input('Remote password (leave blank for public servers):\n').replace(" ","")
```
nmoroze, line 46: I realize this is carried over from previous behavior, but looking at it now - do you think it would make sense to validate whether there are spaces in the name rather than silently replacing them? I could see that tripping someone up when they go to log in later.

nmoroze (approved):
Looks good, thanks! Left a few minor comments.

PR #1281, opened 2023-02-08
Fix polarity on builtin check

nmoroze:
Looks like this got flipped during the switch to builtins: https://github.com/siliconcompiler/siliconcompiler/commit/cab52cdce87490637d643b727c15d72a3a46951c#diff-0a7b686d579b6f3a831a79a61fda4755be2fc2f5cfefc0d848da015810cb3f54L1453

Didn't get caught due to another bug in the test, but it should be captured by tests now.


gadfort (approved):
LGTM

Issue #1257, opened 2023-02-02
step/index cleanup


nmoroze:
This issue tracks various TODOs that have arisen while cleaning up the step/index abstraction change.

- [ ] Change over all schema parameters to reflect our goal for per-node support
- [x] Allow providing step/index overrides via CLI
- [ ] Fix file hashing (and author/date/signature fields)
- [ ] Add additional parameter for tracking input/output files so that tool input/output keypaths don't need to be step/index "required"
- [x] Fix `test_create_cmdline` 
- [ ] Dump step/index values in TCL manifest [wontfix]
- [ ] Fix automatic docs gen to display step/index values [wontfix]
- [ ] Fix HTML report manifest view to handle step/index

PR #1282, opened 2023-02-08
Make 'tool' schema group configurable per-node

nmoroze:
This PR updates all the "tool" group parameters to match our desired per-node configuration settings. Sorry for the huge PR, this touches a lot of files! It's mostly mechanical refactor, but some things to highlight:

- This change removes `if valid('tool', ...)` guards in a lot of places. The `_search()` refactor (##1222) actually removes the need for this in most cases, since `get()` will now return the default value for a parameter if the supplied wildcards don't exist. The instances modified in this PR were caught because `valid()` doesn't accept step/index kwargs.
- Some new functionality is included, since the 'path', 'version', and 'licenseserver' parameters were not previously set on a step/index basis. I've added a test that exercises this new functionality.
- While most of these parameters use the 'optional' pernode configuration, I made 'input', 'output', and 'report' take in the required setting, since these parameters all refer to files in specific step directories (and `find_files` resolves them accordingly)
- Working on this PR revealed the file-hashing logic needs to be fixed (tracked in #1257), since the hashes are stored in a plain list that maps 1-to-1 with the values in a [file] type. This means we don't have a way to store the hashes for per-node files. Fixing this will require changes to the schema structure, so I'm deferring any fixes until we have a chance to discuss.
- TCL manifests still only include the value for a parameter associated with the current step/index. This doesn't break anything even with this PR, since the current show implementations don't access 'show_step'/'show_index' in scripts.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3509,7 +3509,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
 
         # License file configuration.
         for item in self.getkeys('tool', tool, 'licenseserver'):
-            license_file = self.get('tool', tool, 'licenseserver', item)
+            license_file = self.get('tool', tool, 'licenseserver', item, step=step, index=index)
             if license_file:
                 os.environ[item] = ':'.join(license_file)
 
@@ -3528,7 +3528,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
 
         vercheck = not self.get('option', 'novercheck')
         veropt = self.get('tool', tool, 'vswitch')
-        exe = self._getexe(tool)
+        exe = self._getexe(tool, step, index)
         version = None
         toolpath = exe # For record
         if exe is not None:
@@ -3544,7 +3544,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                 version = parse_version(proc.stdout)
 
                 self.logger.info(f"Tool '{exe_base}' found with version '{version}' in directory '{exe_path}'")
-                if vercheck and not self._check_version(version, tool):
+                if vercheck and not self._check_version(version, tool, step, index):
                     self._haltstep(step, index)
             else:
                 self.logger.info(f"Tool '{exe_base}' found in directory '{exe_path}'")
@@ -4356,8 +4356,8 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
     ############################################################################
     # Chip helper Functions
     ############################################################################
-    def _getexe(self, tool):
-        path = self.get('tool', tool, 'path')
+    def _getexe(self, tool, step, index):
+        path = self.get('tool', tool, 'path', step=step, index=index)
         exe = self.get('tool', tool, 'exe')
         if exe is None:
             return None
@@ -4384,7 +4384,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
             command arguments (list)
         '''
 
-        fullexe = self._getexe(tool)
+        fullexe = self._getexe(tool, step, index)
 
         options = []
         is_posix = (sys.platform != 'win32')
@@ -4409,11 +4409,13 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         for key in self.getkeys('option','env'):
             envvars[key] = self.get('option','env', key)
         for item in self.getkeys('tool', tool, 'licenseserver'):
-            license_file = self.get('tool', tool, 'licenseserver', item)
+            license_file = self.get('tool', tool, 'licenseserver', item, step=step, index=index)
             if license_file:
                 envvars[item] = ':'.join(license_file)
-        if self.get('tool', tool, 'path'):
-            envvars['PATH'] = self.get('tool', tool, 'path') + os.pathsep + os.environ['PATH']
+
+        path = self.get('tool', tool, 'path', step=step, index=index)
+        if path:
+            envvars['PATH'] = path + os.pathsep + os.environ['PATH']
         else:
             envvars['PATH'] = os.environ['PATH']
 
@@ -4623,7 +4625,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
 
         return f'{filename}_{pathhash}{ext}'
 
-    def _check_version(self, reported_version, tool):
+    def _check_version(self, reported_version, tool, step, index):
         # Based on regex for deprecated "legacy specifier" from PyPA packaging
         # library. Use this to parse PEP-440ish specifiers with arbitrary
         # versions.
@@ -4642,7 +4644,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
 
         normalize_version = self.find_function(tool, 'normalize_version', 'tools')
         # Version is good if it matches any of the specifier sets in this list.
-        spec_sets = self.get('tool', tool, 'version')
+        spec_sets = self.get('tool', tool, 'version', step=step, index=index)
         if not spec_sets:
             return True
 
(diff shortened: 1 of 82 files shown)


siliconcompiler/schema/schema_cfg.py
```diff
@@ -1165,38 +1168,41 @@ def schema_task(cfg, tool='default', task='default', step='default', index='defa
     key = 'default'
     suffix = 'default'
 
-    scparam(cfg, ['tool', tool, 'task', task, 'warningoff', step, index],
```
gadfort, line 1168: Update schema version

siliconcompiler/tools/bambu/import.py
```diff
@@ -16,15 +16,15 @@ def setup(chip):
     chip.set('tool', tool, 'vswitch', '--version')
     chip.set('tool', tool, 'version', '>=0.9.6', clobber=False)
 
-    chip.set('tool', tool, 'task', task, 'refdir', step, index, refdir, clobber=False)
-    chip.set('tool', tool, 'task', task, 'threads', step, index, os.cpu_count(), clobber=False)
-    chip.set('tool', tool, 'task', task, 'option', step, index, [])
+    chip.set('tool', tool, 'task', task, 'refdir', refdir, step=step, index=index, clobber=False)
+    chip.set('tool', tool, 'task', task, 'threads', os.cpu_count(), step=step, index=index, clobber=False)
+    chip.set('tool', tool, 'task', task, 'option', [], step=step, index=index)
```
gadfort, line 21: is this needed?

siliconcompiler/tools/bambu/import.py
```diff
@@ -16,15 +16,15 @@ def setup(chip):
...
 
     # Input/Output requirements
-    chip.add('tool', tool, 'task', task, 'output', step, index, chip.top() + '.v')
+    chip.add('tool', tool, 'task', task, 'output', chip.top() + '.v', step=step, index=index)
 
     # Schema requirements
-    chip.add('tool', tool, 'task', task, 'require', step, index, 'input,hll,c')
+    chip.add('tool', tool, 'task', task, 'require', 'input,hll,c', step=step, index=index)
```
gadfort, line 27: is step/index needed? This is true for all import tasks for bambu

siliconcompiler/tools/bluespec/import.py
```diff
@@ -21,15 +21,15 @@ def setup(chip):
     chip.set('tool', tool, 'vswitch', '-v')
     chip.set('tool', tool, 'version', '>=2021.07', clobber=False)
 
-    chip.set('tool', tool, 'task', task, 'refdir', step, index,  refdir, clobber=False)
-    chip.set('tool', tool, 'task', task, 'threads', step, index,  os.cpu_count(), clobber=False)
-    chip.set('tool', tool, 'task', task, 'option', step, index,  [], clobber=False)
+    chip.set('tool', tool, 'task', task, 'refdir', refdir, step=step, index=index, clobber=False)
+    chip.set('tool', tool, 'task', task, 'threads', os.cpu_count(), step=step, index=index, clobber=False)
+    chip.set('tool', tool, 'task', task, 'option', [], step=step, index=index, clobber=False)
```
gadfort, line 26: same

siliconcompiler/tools/bluespec/import.py
```diff
@@ -21,15 +21,15 @@ def setup(chip):
...
 
     # Input/Output requirements
-    chip.add('tool', tool, 'task', task,'output', step, index, chip.top() + '.v')
+    chip.add('tool', tool, 'task', task,'output', chip.top() + '.v', step=step, index=index)
 
     # Schema requirements
-    chip.add('tool', tool, 'task', task,'require', step, index, 'input,hll,bsv')
+    chip.add('tool', tool, 'task', task, 'require', 'input,hll,bsv', step=step, index=index)
```
gadfort, line 32: same

siliconcompiler/tools/ghdl/import.py
```diff
@@ -17,14 +17,14 @@ def setup(chip):
     chip.set('tool', tool, 'vswitch', '--version')
     chip.set('tool', tool, 'version', '>=2.0.0-dev', clobber=clobber)
 
-    chip.set('tool', tool, 'task', task, 'threads', step, index, '4', clobber=clobber)
-    chip.set('tool', tool, 'task', task, 'option', step, index, '', clobber=clobber)
-    chip.set('tool', tool, 'task', task, 'stdout', step, index, 'destination', 'output')
-    chip.set('tool', tool, 'task', task, 'stdout', step, index, 'suffix', 'v')
+    chip.set('tool', tool, 'task', task, 'threads', '4', step=step, index=index, clobber=clobber)
```
gadfort, line 20: why 4?

siliconcompiler/tools/netgen/lvs.py
```diff
@@ -20,34 +20,34 @@ def setup(chip):
     chip.set('tool', tool, 'version', '>=1.5.192', clobber=False)
     chip.set('tool', tool, 'format', 'tcl')
 
-    chip.set('tool', tool, 'task', task, 'threads', step, index, 4, clobber=False)
-    chip.set('tool', tool, 'task', task, 'refdir', step, index, refdir, clobber=False)
-    chip.set('tool', tool, 'task', task, 'script', step, index, script, clobber=False)
+    chip.set('tool', tool, 'task', task, 'threads', 4, step=step, index=index, clobber=False)
```
gadfort, line 23: 4?

gadfort (changes requested):
In my first pass, I cannot see anything that is obviously an issue.
Please update the schema version
It would be nice to allow input/output to be set with default. report is tied with metric so should be required.
Also please run the daily CI on this

siliconcompiler/core.py
```diff
@@ -2305,7 +2287,18 @@ def hash_files(self, *keypath, algo='sha256', update=True):
         if 'file' not in self.get(*keypath, field='type'):
             self.error(f"Illegal attempt to hash non-file parameter [{keypathstr}].")
         else:
-            filelist = self.find_files(*keypath)
+            vals = self.schema._getvals(*keypath)
+            # TODO: fix this. will require changing schema structure.
+            if len(vals) > 1:
+                self.logger.warning(
```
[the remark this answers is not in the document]
nmoroze: From a practical standpoint, not a whole lot since I don't think we really enable hashing in current use cases. But I think the tracking/provenance is still considered an important feature, so should definitely get fixed soon.

siliconcompiler/core.py
```diff
@@ -3739,13 +3732,13 @@ def read(fd):
         # Hash files
         if (not is_builtin) and self.get('option', 'hash'):
             # hash all outputs
-            self.hash_files('tool', tool, 'task', task, 'output', step, index)
+            # TODO: only hash outputs from this step
```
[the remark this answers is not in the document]
nmoroze: Yeah, I think this will fall under the scope of the other hashing-related fixes.

nmoroze:
@gadfort thanks for the review!

- Bumped schema version
- Cleaned up the odd tool driver things you mentioned. There's probably a lot more of those sorts of things, and in particular step/index is set in tons of places I don't think it has to be. I figure those issues can be cleaned up in more focused review of each tool driver at some point.
- Daily test run queued, good suggestion: https://github.com/siliconcompiler/siliconcompiler/actions/runs/4146378633
- Tracking the input/output thing as another task in issue #1257

nmoroze:
Alright, got a daily test run that ultimately timed out, but gave me useful failing tests for a couple Python versions https://github.com/siliconcompiler/siliconcompiler/actions/runs/4147175495/jobs/7173735411#step:4:557. 

I pushed fixes for those tests that I verified manually.

gadfort (approved):
Assuming the Daily CI passes, this looks good.

PR #1283, opened 2023-02-09
docs: add variable and file sections to allow for documenting free variables and files

gadfort:
Adds:
- Variable and File sections to the tools (currently only yosys has this done).


nmoroze (approved):
LGTM, thanks!

PR #1286, opened 2023-02-13
Fix and update scheduler.py

WRansohoff:
* Use the top-level 'scheduler' keypath in scheduler.py
* Run tasks on compute nodes using Python build scripts, instead of the `sc` command line.

Unfortunately, we don't have a way to un-set an enum Schema value from the command line. I can report that edge case bug, but it's probably not a priority to fix it right this moment, so for now we can use the Python API's `Chip.unset()` call to mark when a task has reached the node that it should run on.

This is probably a good change to make anyways, as it will give us more flexibility going forward.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3394,7 +3394,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         # (Run the initial 'import' stage[s] locally)
 
         if self.get('option', 'scheduler', 'name', step=step, index=index) and \
-           self.get('flowgraph', flow, step, index, 'input') and \
+           self.get('flowgraph', flow, step, index, 'input'):
             # Note: The _deferstep method blocks until the compute node
             # finishes processing this step, and it sets the active/error bits.
             _deferstep(self, step, index, status)
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3393,9 +3393,8 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         # and send it to a compute node for deferred execution.
         # (Run the initial 'import' stage[s] locally)
 
-        if self.get('option', 'scheduler', 'name') and \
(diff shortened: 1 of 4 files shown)


WRansohoff:
My original plan was to add a `none` or `local` enum value, but I was discouraged from doing that in initial feedback.

New feedback seems to be that the extra Python script is too complicated, so the original idea is back in play.

siliconcompiler/core.py
```diff
@@ -4029,6 +4030,7 @@ def run(self):
                     self.set('option', 'steplist', pre_remote_steplist['steplist'])
                 else:
                     self.unset('option', 'steplist')
+                self.unset('option', 'scheduler', 'name')
```
nmoroze, line 4033: Why the need to unset this?

nmoroze (approved):
One small question, but otherwise looks good!

siliconcompiler/core.py
gadfort: Could that be used in `remote_run`?

gadfort (approved):

siliconcompiler/core.py
WRansohoff: We need to un-set it because the local `Chip` object will pick up the `('option', 'scheduler', 'name')` value from the remote run's manifest, and try to run the screenshot step which happens later on a cluster.

If we unset the value during the remote run, it would get written out as `None` in each task's output manifest, and following tasks wouldn't get the correct value to run the job through the scheduler. It would also be risky to have the 'runner' node modify the output manifests after each task is completed, because NFS can be a bit slow to register new files which another machine created, and re-mounting the drive is the only reliable solution I've found for that so far.

That said, I have one new idea after sleeping on this: if we let the `('option', 'scheduler', 'name')` parameter be set on a per-node basis, the server could specify 'slurm' in only those tasks which are in the remote steplist, and leave the default value untouched.

nmoroze (approved):
Per-node scheduler change looks good, thanks!

PR #1287, opened 2023-02-13
docs: switch to use setup() or module for documentation

gadfort:
Changes:
- uses `setup()` instead of `make_docs()` for documentation + associated changes in tools/pdks/libs/checklists/flows

Note:
- This is a step towards making `make_docs()` or a similar function optional

### DIFF
--- a/siliconcompiler/checklists/oh_tapeout.py
+++ b/siliconcompiler/checklists/oh_tapeout.py
@@ -1,14 +1,14 @@
 import siliconcompiler
 
 def make_docs():
-    '''Subset of OH! library tapeout checklist.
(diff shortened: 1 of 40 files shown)


aolofsson:
Looks good, but why not just remove make_docs?

gadfort:
@aolofsson I decided to do that in a separate PR. Some of the tools/flows need the `Chip` object to contain some information, so for those, we'll need to have a way to configure the `Chip`.

nmoroze (approved):
Looks good, thanks!

aolofsson (approved):
LG!

PR #1288, opened 2023-02-13
Make 'constraint' parameters specifiable on a per-node basis

nmoroze:
- This entails having to tack on step/index to any instances where we get from a "constraint" keypath in Python. A few opportunities came up to simplify tool driver logic while I was working on this
- I'm not sure a great way to test this functionality (although I think the physical constraints schema could be easy to test), so it's currently untested -- would be open to suggestions though (I'm also nervous adding more tests that run the flow given it seems like we've pushing the runtime a bit lately)


siliconcompiler/tools/yosys/yosys.py
```diff
@@ -342,14 +342,13 @@ def get_synthesis_corner(chip):
...
-            corner = chip.get('constraint', 'timing', constraint, 'libcorner')
+        if "setup" in chip.get('constraint', 'timing', constraint, 'check', step=step, index=index) and not corner:
+            corner = chip.get('constraint', 'timing', constraint, 'libcorner', step=step, index=index)
 
     if corner is None:
         # try getting it from first constraint with a valid libcorner
         for constraint in chip.getkeys('constraint', 'timing'):
-            if chip.valid('constraint', 'timing', constraint, 'libcorner') and not corner:
-                corner = chip.get('constraint', 'timing', constraint, 'libcorner')
+            corner = chip.get('constraint', 'timing', constraint, 'libcorner', step=step, index=index)
```
gadfort, line 351: Lost the `if not corner:` check

gadfort (approved):
Please fix the check in yosys for corner, otherwise looks good.

nmoroze:
Fixed the check, nice catch!

PR #1289, opened 2023-02-13
Make input/output parameters specifiable on a per-node basis

nmoroze:
Same deal as most recent PR.

One thing that's making me nervous as I work on this - tool driver authors need to be careful to tack on the step/index for these keypaths when calling get(), since if they forget there might be a silent error if a user tries to supply a step/index-specific setting and it doesn't go through.

It might be worth going back and making step/index required when doing get/find_files for "optional" per-node parameters, any thoughts on that?


nmoroze:
@gadfort I updated this PR to add a check to `get()` that complains if you don't supply step/index for "optional"-pernode params. 

I'd say it's medium invasive. I had to fix some things in the core, but they were probably important to tweak anyways. A big problem is that we have a lot of places where we loop over all keypaths and do something, and this change makes these loops even more complicated since we're introducing more irregularity into how different types of parameters are implemented.

I also realized that these sorts of loops require some way to explicitly get the "global" value. I decided to implement this by allowing the string "default" to be used for the step/index keywords:

```python
>>> chip.set('asic', 'logiclib', ['mylib'])
>>> chip.set('asic', 'logiclib', ['synlib'], step='syn', index=0)
>>> chip.get('asic', 'logiclib')
# raises exception
>>> chip.get('asic', logiclib, step='default', index='default')
['mylib']
>>> chip.get('asic', 'logiclib', step='syn', index=0)
['synlib']
```

I'm not 100% happy about this, but I think it works out okay - it requires the author of the code to be explicit about bypassing the step/index override values, it should be invisible to users, and it's safe since "default" isn't a legal step value (would conflict with the schema placeholders).

nmoroze:
I decided to move the check logic to #1296 to make it easier to see the impact of those changes. Will mark this as draft and rebase on top of it once that one is merged, to make sure that I had good coverage adding step/index.

aolofsson:
Looking at this, I am starting to see the problem with the per node value, mostly around the user having to remember the priority and value of the pernode value vs global. Having the get below raise an exception feels like a step backwards, and adding the 'default' step does not help. Without a step/index, wouldn't we expect that to return the global value? Sorry, I am getting lost in the new logic. This may be my own fault (or not..), user

```python
>>>chip.get('asic', 'logiclib')
```

gadfort:
@aolofsson in speaking with @nmoroze there doesn't appear to be any cases where the `getter` needs to get the global value (they would still get it if the value is not set for that step/index). By generating an error if the parameter has a step/index that is optional it prevents users from accidentally getting the global when they were intending to get the step/index value.

nmoroze:
Yeah, the thinking is that this mostly affects tool driver writers - a designer-user probably doesn't have to call get() much unless they're trying to do some sort of design introspection. Adding the exception revealed a lot of places where I had forgotten to add step/index (which would have resulted in buggy drivers), so it seems value in that regard.

I think the best alternative to the verbosity would be to do the "magic" getter that determines step/index from [arg, step] and [arg, index], but that has its own pitfalls as we've discussed.

gadfort (approved):
LGTM, you may want to merge the strict PR first and then just go a merge to make sure nothing was missed.

PR #1290, opened 2023-02-13
Make certain "option" parameters specifiable on a per-node basis

nmoroze:
This PR implements most of the "option" group parameters we wanted to make specifiable on a per-node basis. 

There were a few issues I ran into while going through the list:

- The spreadsheet wasn't updated for the new scheduler-related items, but since 'jobscheduler', 'msgevent', and 'msgcontact' were all meant to be optional pernode parameters, I set `pernode='optional'` for the entire scheduler group for now (although I haven't touched the implementation yet -- so might have to revert this depending on if we add checks to `get()`)
- I didn't change the following items yet, since I'm not sure if it makes much sense to do so:
  - 'skipstep' and 'indexlist' - - 'skipstep' is unused, and, if I recall correctly, we are planning to replace this entire mechanism altogether
  - 'showtool': unsure if we want to make it per-"show step", or per-step in terms of the step name in e.g. the showflow
  - 'copyall': seems like a global parameter - it gets checked once during the `_collect_paths()` call at import. We could potentially use it to affect the behavior of which paths we collect, but doing this per-node seems tough. We could do it based on the 'require' keypaths, but I'm not sure those are used consistently enough at the moment.
  - 'show': is unused, and also meant to be a global parameter I think (whether or not show() automatically gets called at the end of the run is how I read it)


WRansohoff (approved):
Looks great, thanks for doing this!

I think we can simplify #1286 by setting `('option', 'scheduler', 'name')` on a per-node basis, so great timing.

aolofsson:
LGTM!

nmoroze:
Rebased this on latest main, and caught a couple more things thanks to the strict mode test. Should be good for review again.

gadfort (approved):
LGTM

PR #1291, opened 2023-02-13
add back option,dir which got lost

gadfort:


aolofsson (approved):
Makes sense. Nice catch!

PR #1293, opened 2023-02-14
fix docs build after step/index change

gadfort:


nmoroze (approved):
Looks good, thanks for fixing! I guess this will only show one of the values assigned to a parameter, but I can't think of a scenario where we'd likely need to document more than one value at this point.

PR #1294, opened 2023-02-15
Sc 13 asap7 corners

### LINKED ISSUE #1100: Hard coded ASAP7 libs
The asap7 timing libraries are currently hard codede as "typical" with ff timing based on the library setup file.
This should be fixed to use the correct scenarios and ff/ss/tt libraries (with .gz suffix).

gadfort:
Closes #1100

Changes:
- timing libraries needed to be fixed
- enables all three timing corners on asap7 target


nmoroze (approved):
Code changes look good, thanks. Any reason for deleting the particular libraries that got removed?

gadfort:
@nmoroze the file removed was just an unzipped version of the others, the tools can read gz files so no need to have a text version

PR #1295, opened 2023-02-15
Support step/index overrides in command line

nmoroze:
This PR adds support for providing step/index for per-node parameters via the CLI. 

The syntax is to add the step and index name prior to the value. These must always be included for "required" pernode parameters:

- `sc -record_distro 'import 0 ubuntu'` (good)
- `sc -record_distro ubuntu` (error)

You are still allowed to supply string values with spaces for required pernode parameters, as before:

- `sc -record_distro 'import 0 my distro name'` -> stores "my distro name" in [record, distro]

For optional parameters, you can omit the step and index, provide a step only, or provide a step and an index:

- `sc -asic_logiclib mylib`
- `sc -asic_logiclib 'syn synlib'`
- `sc -asic_logiclib 'syn 1 syn1lib'`

Values with spaces for optional parameters must be wrapped in quotes now:

- `sc -asic_logiclib 'syn "syn lib"'`

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -463,21 +463,41 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                     self.error(f'Invalid value {item} for switch {switchstr}. Expected format {metavar}.', fatal=True)
 
                 # We replace 'default' in keypath with first N words in provided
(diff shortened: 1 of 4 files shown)


gadfort (approved):
LGTM. Thanks

PR #1296, opened 2023-02-15
Require step/index when reading optional pernode params

nmoroze:
This PR splits out the new check for `get()` and `find_files()` from PR #1289 for ease of review. With this PR, all optional per-node parameters require a user to specify a step/index, to make sure that related features are implemented correctly.

For the internal cases where we need to access the global value for one of these parameters, this PR adds private `_get()` and `_find_files()` methods that do not have this additional check. The need for this mainly comes up with:

- patterns where we need to loop over all values of a parameter and operate on them one at a time
- for building up more complex functionality (e.g. the call to `get()` within `find_files()`)


nmoroze:
@gadfort I think this is once again ready for review.

- Implemented ['option', 'strict'] per Andreas' suggestion
  - This is implemented at the Chip.get() level, since I want to preserve the behavior that the Schema object doesn't rely on any particular parameter being available (could be helpful for implementing schema subsets)
  - This lets us use Schema.get() as a bypass of the check for internal code
  - Kept _find_files() as another internal bypass
- Added a pytest fixture to always set 'strict' in our tests, with a marker that we can use to get rid of it on a per-test basis

siliconcompiler/core.py
```diff
@@ -783,6 +783,15 @@ def get(self, *keypath, field='value', job=None, step=None, index=None):
         self.logger.debug(f"Reading from {keypath}. Field = '{field}'")
 
         try:
+            strict = self.schema.get('option', 'strict')
+            if field == 'value' and strict:
+                pernode = self.schema.get(*keypath, field='pernode')
+                if pernode == 'optional' and (step is None or index is None):
+                    raise ValueError(
+                        f'Invalid args to get() of keypath {keypath}: step and '
+                        'index are required for reading from this parameter'
```
gadfort, line 792: Maybe add 'while strict is True'

siliconcompiler/core.py
```diff
@@ -1124,6 +1133,19 @@ def find_files(self, *keypath, missing_ok=False, job=None, step=None, index=None
             the schema.
 
         """
+        strict = self.get('option', 'strict')
+        pernode = self.get(*keypath, field='pernode')
+        if strict and pernode == 'optional' and (step is None or index is None):
+            self.error(f'Invalid args to find_files() of keypath {keypath}: step and index '
+                'are required for reading from this parameter')
```
gadfort, line 1140: Maybe add 'while strict is True'

siliconcompiler/schema/schema_cfg.py
```diff
@@ -2654,6 +2654,18 @@ def schema_option(cfg):
             if an operation should continue. The timeout value is also
             useed by the jobscheduler to automatically kill jobs.""")
 
+    scparam(cfg, ['option', 'strict'],
```
gadfort, line 2657: Bump version of schema

gadfort (changes requested):
Looks good, just need to correct the schema version and see about the error messages.

gadfort (approved):

PR #1297, opened 2023-02-15
schema_cfg: add import fallback for generate_defaults

nmoroze:
I noticed that the relative import required for our new KLayout scheme breaks the generate_defaults script - this PR fixes that.

### DIFF
--- a/siliconcompiler/schema/schema_cfg.py
+++ b/siliconcompiler/schema/schema_cfg.py
@@ -4,9 +4,9 @@ import copy as pycopy
 import json
 import re
 
-# Default import must be relative, to facilitate KLayout's direct import of the
-# schema package. However, the fallback allows running this script directly to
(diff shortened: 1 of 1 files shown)


siliconcompiler/schema/schema_cfg.py
```diff
@@ -4,7 +4,13 @@
 import json
 import re
 
-from .utils import trim
+# Default import must be relative, to facilitate KLayout's direct import of the
```
gadfort, line 7: While klayout is the tool at the moment with this issue, others will come. Maybe change the comment to reflect that it's for the tools with python interfaces (like klayout) that need this.

gadfort (approved):
Please address the comment, but otherwise looks good. Thanks for fixing this I got caught with this issue the other day and didn't have time to fix it.

PR #1299, opened 2023-02-15
Remove stdin requirement from yosys setup script

WRansohoff:
Currently, the yosys setup script requires you to enter 'y' in response to the apt-get install command. This is a bit inconvenient for headless install methods.


gadfort (approved):
Thanks

PR #1300, opened 2023-02-16
schema_obj: update write_csv to handle per-node values

nmoroze:
This wasn't on my formal todo list, but I had the code for this already so I figured we might as well get it in (was going to be bundled with TCL manifest changes we're punting).

This PR adds some logic to the `write_csv` method to dump per-node values as well as the global values. Previous behavior was that per-node values get ignored.

Example:

```python
    chip.add('asic', 'logiclib', 'mainlib')
    chip.add('asic', 'logiclib', 'synlib', step='syn')
    chip.add('asic', 'logiclib', 'syn1lib', step='syn', index=1)
```

after


gadfort (approved):
LGTM

PR #1301, opened 2023-02-16
Remove make docs SC-49

gadfort:
Changes:
- Makes `make_docs` optional and removes `make_docs` from most pdks/libs/flows/targets/checklists
- The interface to `make_docs` changes from `make_docs()` to `make_docs(chip)` so the docs generator can pass in chip object (useful for tools).
- If `make_docs` does not return anything it's simply configuring the chip object (which the docs generator will pass into the setup function itself), otherwise it behaves as it did previously by returning a fully configured chip object for the docs to use.

### DIFF
--- a/siliconcompiler/checklists/oh_tapeout.py
+++ b/siliconcompiler/checklists/oh_tapeout.py
@@ -1,9 +1,5 @@
 import siliconcompiler
 
-def make_docs():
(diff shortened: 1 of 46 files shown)


aolofsson (approved):
Looks great!

PR #1303, opened 2023-02-19
Add tools to tools.json and fix surelog SC-169

gadfort:


aolofsson:
Noticed the dependence on "flopoco" in the panda build. Can you remove it, it's not a proper open source project.

gadfort:
> Noticed the dependence on "flopoco" in the panda build. Can you remove it, it's not a proper open source project.

Remove the option flopoco in the build or remove the panda build all together?

aolofsson:
Just the flopoco. Panda should be fine, although I'm am not sure we have
reviewed the licenses for all the install tools..?

setup/install-klayout.sh
```diff
@@ -1,9 +1,12 @@
 #!/bin/sh
 
+# Get directory of script
+src_path=$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)
+
 mkdir -p deps
 cd deps
 
-pkg_version="0.27.8-1"
+pkg_version=$(python3 ${src_path}/_tools.py --tool klayout --field version)
```
WRansohoff, line 9: My gut says that sed or awk might be better for extracting strings in a bash script, but JSON isn't fun to parse that way, and this *is* a Python project. Who knows, maybe someday we'll move these setup scripts from .sh to .py.

WRansohoff (approved):

setup/install-klayout.sh
gadfort: sed or awk would be pretty hard to manage. The point of the _tools.json file is to have the versions in a single place and the python script here helps get the information out. We could look at making python scripts, but given how finicky these install scripts are to begin with I don't know if we want to be the single source of install scripts.

PR #1304, opened 2023-02-19
General fixes

gadfort:


siliconcompiler/tools/yosys/yosys.py
```diff
@@ -368,6 +368,11 @@ def get_dff_liberty_file(chip):
     if dff_liberty:
         return dff_liberty[0]
 
+    mainlib = chip.get('asic', 'logiclib')[0]
```
WRansohoff, line 371: At some point in the future, we may want to add a Chip helper method for fetching this 'mainlib' value, and clarify if 'first entry' is a good criteria.

tests/examples/test_gcd.py
```diff
@@ -43,11 +43,11 @@ def test_py(setup_example_test):
     # "Creating fake entries in the LUT"
     assert chip.get('metric', 'warnings', step='cts', index='0') == 2
 
-    # Missing route to pin (x77)
-    assert chip.get('metric', 'warnings', step='route', index='0') == 77
+    # Missing route to pin (x71)
+    assert chip.get('metric', 'warnings', step='route', index='0') == 71
```
WRansohoff, line 47: Good to see the number of warnings going down, but I've also had a recent change get tripped up on this test's number-of-warnings checks. Should we make a note to update this test to be less rigid, maybe looking for a list of expected error classes instead of an exact number?

WRansohoff (approved):
Just a couple of non-blocking comments.

siliconcompiler/tools/yosys/yosys.py
gadfort: Helper functions would be great, but those can wait.

tests/examples/test_gcd.py
gadfort: This test is way too sensitive, these are a "bug" in OpenROAD. Them going up or down doesn't mean a lot for our flow implementation.

PR #1305, opened 2023-02-20
Make .use() take in a chip object

gadfort:
Changes:
- Currently to use a previous Chip (such as in a hierarchical flow), a separate class had to be created to be able to use this. This change allows for the chip object to passed into .use() directly as a new library.

Previous flow:
```
module = siliconcompiler.Chip(“sub”)
…
module.run()
class ModuleClass():
  def setup(chip):
    lib = siliconcompiler.Library(chip, “sub-lib”)
    lib.set(“output”, stackup, “lef”, module.find_result(“lef”, step=”export”))
    lib.set(“output”, corner, “nldm”, module.find_result(“lib”, step=”export”))

chip = siliconcompiler.Chip(“top”)
chip.use(ModuleClass)
chip.add('asic', 'macrolibs', “sub-lib”)
chip.run()
…
```

New flow (without the indirection of creating a new class):
```
module = siliconcompiler.Chip(“sub”)
…
module.run()
module.set(“output”, stackup, “lef”, module.find_result(“lef”, step=”export”))
module.set(“output”, corner, “nldm”, module.find_result(“lib”, step=”export”))

chip = siliconcompiler.Chip(“top”)
chip.use(module)
chip.add('asic', 'macrolibs', module.top())
chip.run()
…
```


WRansohoff (approved):
Looks good as long as we can easily change `Chip` to `Schema` when we make that the base class for the `PDK`/`Flow`/etc objects, but I don't see anything that would get in the way of that.

aolofsson:
Looks good to me. Only recommendation would be to make the support for object/module orthogonal. ie. make sure a design/pdk/tool/flow can all be passed in as a python module (in which case setup() is called) or as a direct object.

PR #1306, opened 2023-02-20
Remove floorplan API

gadfort:
Removes the floorplan API, this will be replaced with the new OpenROAD ICeWall (https://github.com/The-OpenROAD-Project/OpenROAD/pull/2909) once it's been merged in.


examples/gcd/sky130_floorplan.py
```diff
@@ -73,6 +71,9 @@ def place_pdn(fp, vdd, vss, hwidth, hspacing, hlayer, vwidth, vspacing,
    fp.place_wires([vss] * ngnd, core_left, bottom, 0, spacing,
                     core_w, stripe_w, stripe_layer, 'followpin')
 
+
+   # move here until test can be rebuilt
+   from siliconcompiler.floorplan import _layer_i
```
WRansohoff, line 76: Should we maybe delete these examples, or are we planning to try re-implementing the Floorplan API in the near future? They'll still be in the git history if we remove them. Ditto for the other tests.

WRansohoff (approved):
Looks good to me, thanks for cleaning up the core.py imports. I could go either way on fully removing the Floorplan example scripts, but it might be a little confusing to see import statements for modules that don't exist.

examples/gcd/sky130_floorplan.py
gadfort: @WRansohoff they will be re-implemented in the near future, so it seemed easier to keep them around as a starting point.

aolofsson:
Deleting the code means we should be sure that we never want to use this functionality. In terms of APIs, is there a way to every drive a floor plan manually and exactly in Python without this. I would have for us to go down the path of reinventing the wheel again. One approach might be to emit TCL openroad code instead of emitting DEF, but to keep the python API shell. Thoughts? @nmoroze

nmoroze (approved):
LGTM, thanks for the cleanup!

Do you think ZeroSoC will stick around or will be replaced by a different example? If not, it might be good to take out `docs/tutorials/zerosoc.rst` as well, that's pretty heavily coupled to this API.

nmoroze:
@aolofsson I was looking at it the other way round - if we delete this code we can always revive the functionality in the future, but if we keep it around we're (somewhat) committing to keep the functionality. Re-implementing it seems like a decent bit of work, so it seems best to delete now (given it's completely broken), and then reintroduce later if we want to?

aolofsson:
Ok, works for me.

PR #1307, opened 2023-02-20
Update pdks SC-131

gadfort:
Updates the currently available PDKs:
- Mirrors options where appropriate from ORFS
- Adds pex and timing corners to sky130


aolofsson (approved):
Looks great, an important step to synching with orfs.

WRansohoff (approved):

PR #1308, opened 2023-02-20
sc: fix inference of design name from sources

nmoroze:
It looks like we broke this functionality when adding the extra layer of nesting to the 'input' keypath. Added a test for it to prevent regressions in the future.

Also switched from using `get()` to `_getvals()` to make this compatible with PR #1289

### DIFF
--- a/siliconcompiler/apps/sc.py
+++ b/siliconcompiler/apps/sc.py
@@ -46,10 +46,12 @@ def main():
         topfile = None
         for sourceset in ('rtl', 'hll'):
             for filetype in chip.getkeys('input', sourceset):
-                sources = chip.schema._getvals('input', sourceset, filetype)
-                if sources:
-                    # triple index to get first of this structure: ([val], <step>, <index>)]
-                    topfile = sources[0][0][0]
+                all_vals = chip.schema._getvals('input', sourceset, filetype)
+                if all_vals:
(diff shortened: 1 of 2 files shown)


siliconcompiler/apps/sc.py
```diff
@@ -45,12 +45,12 @@ def main():
...
-                for filetype in chip.getkeys('input', sourceset):
-                    sources = chip.get('input', sourceset, filetype)
-                    if sources:
-                        topfile = sources[0]
-                        break
+            for filetype in chip.getkeys('input', sourceset):
+                sources = chip.schema._getvals('input', sourceset, filetype)
+                if sources:
+                    # triple index to get first of this structure: ([val], <step>, <index>)]
+                    topfile = sources[0][0][0]
```
gadfort, line 52: Could you break this into two accesses into the array? Something like files=sources[0] and topfile = files[0][0] that way the comment is a little more clear.

gadfort (approved):
Looks good. Please fix the triple into to make is more readable. I'll approve so you can merge when it's been fixed.

siliconcompiler/apps/sc.py
nmoroze: fixed, thanks - I should know better :)

PR #1309, opened 2023-02-21
Fix Wheels build

nmoroze:
The builds themselves seemed to be fine, it was just a few tests that were failing:

- new [option, nice] code had an undefined variable in Windows
- bug in special case in `find_files()`: I think related to the fact that ['tool', tool, 'path'] is set in KLayout specifically for MacOS/Windows
- at some point show changes made `test_show_lyp` run OpenROAD instead of KLayout
- `yaml` is not actually a built-in package

Validated working: https://github.com/siliconcompiler/siliconcompiler/actions/runs/4227188616


gadfort (approved):
LGTM. Thanks

PR #1310, opened 2023-02-21
Fix daily CI

gadfort:
Changes:
- Restores daily CI to passing
- Adds a default timeout of 60s for all tests to ensure the CI finishes
- Adds special timeouts for tests requiring more than 60 seconds
- Extends the overall timeout to 60mins


nmoroze (approved):
LGTM, thanks!

PR #1311, opened 2023-02-21
Refactor schema structure to store global values alongside per-node values

nmoroze:
In preparation for fixing per-node file hashing, this PR refactors the schema structure per Andreas' suggestion to make the way we store values more consistent. Now, rather than having their own key, global values are stored alongside the per-node values in a field that has been renamed from `nodevalue` to `nodefield`. 

**Before**

```python
{
  'shorthelp': ASIC logiclib',
  'value': ['mylib'],
  'set': True,
  'nodevalue': {
    'syn': {
      'default': ['synlib'],
      '1':  ['syn1lib']
  },
  # ...
}
```

**After**

```python
{
  'shorthelp': ASIC logiclib',
  'nodevalue': {
    'global': {
      'global': {
        'value': ['mylib']
      }
    },
    'syn': {
      'global': {
        'value': ['synlib']
      }
      '1': {
        'value': ['syn1lib']
      }
    }
  },
  # ...,
}
```

The tradeoffs of this change -

Pros:
- More consistent handling of values, which simplifies the implementation somewhat

Cons:
- The nesting makes the JSON dump a bit harder to read
- Using 'global' as a keyword restricts the user from using this as a step name

The rename from `nodevalue` to `nodefield` and the additional layer of nesting introduced with the `value` key are both forward-looking to the filehash fix, which I'll make in a follow-on PR.


aolofsson:
- "nodefield" seems a bit redundant, since they are all fields based on the docs. What about just calling it 'node'. https://docs.siliconcompiler.com/en/stable/user_guide/data_model.html

- The 'set' flag is removed in the new approach (replaced with looking for the key). Do we have a test to verify that the logic around set/unset works with the new approach?

- Will the hash, author,.. be changed later?

nmoroze:
- Makes sense, I think 'node' sounds good.
- Yeah, we have a handful of tests of the behavior in https://github.com/siliconcompiler/siliconcompiler/blob/main/tests/core/test_setget.py (`test_unset()`, `test_no_clobber_false()`). This was changed to make things consistent with how the status of node values was checked previously.
- Yep, planning to change hash, author, etc. later.

aolofsson (approved):

siliconcompiler/core.py
```diff
@@ -3065,6 +3065,9 @@ def node(self, flow, step, tool, task, index=0):
             >>> chip.node('asicflow', 'place', 'openroad', index=0)
             Creates a task with step='place' and index=0 and binds it to the 'openroad' tool.
         '''
+        if step == Schema.GLOBAL_KEY:
```
gadfort, line 3068: check for all reserved words

siliconcompiler/core.py
```diff
@@ -3099,6 +3102,9 @@ def edge(self, flow, tail, head, tail_index=0, head_index=0):
             >>> chip.edge('place', 'cts')
             Creates a directed edge from place to cts.
         '''
+        if head == Schema.GLOBAL_KEY or tail == Schema.GLOBAL_KEY:
```
gadfort, line 3105: same

siliconcompiler/schema/schema_obj.py
```diff
@@ -38,6 +38,10 @@ class Schema:
         manifest (str): Initial manifest.
     """
 
+    # Special key in node dict that represents a value correponds to a
+    # global default for all steps/indices.
+    GLOBAL_KEY = 'global'
```
gadfort, line 43: Add default

gadfort (changes requested):

nmoroze:
I added checks for default in all places where I've got a global key check. I can't make it part of the same `GLOBAL_KEY` variable since that is meant to be a scalar used for indexing into the 'node' field.

gadfort (approved):
Great

PR #1313, opened 2023-02-23
add help string to configure so it can be included in docs

gadfort:


gadfort:
@WRansohoff if you think the help string should be different let me know and I'll update it

WRansohoff (approved):
Looks good, thanks - I'll probably update this docstring once we decide on the messaging for a brief ToS blurb.

PR #1319, opened 2023-02-23
Implement per-node fields and fix file hashing

nmoroze:
This PR implements storing the following fields inside the 'node' dict within a schema parameter: `author`, `date`, `filehash`, `signature`. This is necessary since all of these fields are meant to have a 1:1 correspondence with values.

I also took the chance to clean up the correspondence between `hash_files()`'s actual implementation and its docstring, and added more test cases for file-hashing and set-get logic.


gadfort:
Could you also update: https://github.com/siliconcompiler/siliconcompiler/blob/main/docs/user_guide/data_model.rst to reflect all the changes? I'll review tomorrow

siliconcompiler/core.py
```diff
@@ -1360,20 +1360,36 @@ def _merge_manifest(self, src, job=None, clobber=True, clear=True, check=False,
                 if not key_valid:
                     self.logger.warning(f'Keypath {keylist} is not valid')
             if key_valid and 'default' not in keylist:
-                # update value, handling scalars vs. lists
                 typestr = src.get(*keylist, field='type')
                 for val, step, index in src._getvals(*keylist, return_defvalue=False):
+                    # update value, handling scalars vs. lists
                     if re.match(r'\[', typestr) and not clear:
                         dest.add(*keylist, val, step=step, index=index)
+                        add = True
```
gadfort, line 1368: Wouldn't this be cleaner with a test for `is_list = re.match...`

siliconcompiler/core.py
```diff
@@ -1360,20 +1360,36 @@ def _merge_manifest(self, src, job=None, clobber=True, clear=True, check=False,
...
                     else:
                         dest.set(*keylist, val, step=step, index=index, clobber=clobber)
+                        add = False
+
+                    # update other pernode fields
+                    # TODO: only update these if clobber is successful
```
gadfort, line 1374: What is missing to complete this TODO?

siliconcompiler/core.py
```diff
@@ -1360,20 +1360,36 @@ def _merge_manifest(self, src, job=None, clobber=True, clear=True, check=False,
...
+                            dest.set(*keylist, v, step=step, index=index, field=field)
 
                 # update other fields that a user might modify
                 for field in src.getdict(*keylist).keys():
                     if field in ('node', 'switch', 'type', 'require', 'defvalue',
                                  'shorthelp', 'example', 'help'):
-                        # skip these fields (value handled above, others are static)
+                        # skip these fields (node handled above, others are static)
                         continue
+                    # TODO: should we be taking into consideration clobber for these fields?
```
gadfort, line 1392: same

siliconcompiler/schema/schema_obj.py
```diff
@@ -462,6 +473,9 @@ def _normalize_field(value, sc_type, field, keypath):
         def error_msg(t):
             return f'Invalid value {value} for field {field} of keypath {keypath}: expected {t}'
 
+        if field in ('author', 'filehash', 'date', 'hashalgo', 'copy') and ('file' not in sc_type):
```
gadfort, line 476: This field check if almost the same as PERNODE_FIELDS, would it be possible to use that here?

siliconcompiler/schema/schema_obj.py
```diff
@@ -41,6 +41,7 @@ class Schema:
     # Special key in node dict that represents a value correponds to a
     # global default for all steps/indices.
     GLOBAL_KEY = 'global'
+    PERNODE_FIELDS = ('value', 'filehash', 'date', 'author', 'signature')
```
gadfort, line 44: Update schema version

gadfort (changes requested):

siliconcompiler/schema/schema_obj.py
nmoroze: These are fields that are valid only for file-type values, which happen to overlap a bit but are distinct meaning-wise.

siliconcompiler/core.py
nmoroze: It's actually a little tricky, I opened a ticket SC-187 to track with a test case and a couple proposed solutions.

nmoroze:
Updated data model and added step/index to docstrings while I was at it, bumped schema version, and cleaned up logic where suggested.

gadfort (approved):
LGTM

PR #1323, opened 2023-02-24
fix links reported by make linkcheck as broken or redirecting

gadfort:


docs/resources/glossary.rst
```diff
@@ -186,7 +186,7 @@ Manufacturing
 * `Silicon on insulator <https://en.wikipedia.org/wiki/Silicon_on_insulator>`_: Layered silicon–insulator–silicon with reduced parasitic capacitance.
 * `Stepper <https://en.wikipedia.org/wiki/Stepper>`_: Machine that passes light through reticle onto the silicon wafer being processed.
 * `TSV <https://en.wikipedia.org/wiki/Through-silicon_via>`_: Vertical electrical connection (via) passing completely through a silicon wafer or die.
-* `Wafer <https://en.wikipedia.org/wiki/Wafer_(electronics>`_:: Thin slice of semiconductor material used in electronics for the fabrication of integrated circuits.
+* `Wafer <https://en.wikipedia.org/wiki/Wafer_(electronics)>`_:: Thin slice of semiconductor material used in electronics for the fabrication of integrated circuits.
```
WRansohoff, line 189: Should this also have `%28`/`%29` instead of `(`/`)`?

WRansohoff (approved):

docs/resources/glossary.rst
gadfort: It doesn't seem to care, but for the flipflops wikipedia seems to want %. (I went with the URL wikipedia generated, so I think it's okay)

PR #1325, opened 2023-02-24
Fix schema key links

gadfort:
Changes:
- make keypaths in docs a generated link instead of hardcoded to the website

### DIFF
--- a/siliconcompiler/sphinx_ext/dynamicgen.py
+++ b/siliconcompiler/sphinx_ext/dynamicgen.py
@@ -791,6 +791,7 @@ class SCDomain(sphinx.domains.std.StandardDomain):
     name = 'sc'
 
     # Override in StandardDomain so xref is literal instead of inline
+    # https://github.com/sphinx-doc/sphinx/blob/ba080286b06cb9e0cadec59a6cf1f96aa11aef5a/sphinx/domains/std.py#L789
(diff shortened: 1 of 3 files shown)


siliconcompiler/sphinx_ext/dynamicgen.py
```diff
@@ -775,16 +775,47 @@ def document_module(self, module, modname, path):
...
     except ValueError as e:
         msg = inliner.reporter.error(f'{rawtext}: {e}', line=lineno)
         prb = inliner.problematic(rawtext, rawtext, msg)
         return [prb], [msg]
 
+class SCDomain(sphinx.domains.std.StandardDomain):
+    name = 'sc'
+
+    # Override in StandardDomain so xref is literal instead of inline
+    def build_reference_node(self, fromdocname, builder, docname, labelid, sectname, rolename, **options):
```
nmoroze, line 794: Is the body of this function derived from the original in StandardDomain? Maybe include link to the source if so.

nmoroze (approved):
That's some Sphinx magic right there. Thanks for fixing!

siliconcompiler/sphinx_ext/dynamicgen.py
gadfort: Good point

PR #1326, opened 2023-02-24
setup: fix Surelog install script

nmoroze:


setup/install-surelog.sh
```diff
@@ -6,16 +6,21 @@ src_path=$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)
...
 cd surelog
+
 git checkout $(python3 ${src_path}/_tools.py --tool surelog --field git-commit)
 git submodule update --init --recursive
 
+# Use clean env for build
+python3 -m venv .venv
+PYTHON_ROOT=$(realpath .venv)
+$PYTHON_ROOT/bin/python -m pip install orderedmultidict
+export ADDITIONAL_CMAKE_OPTIONS="-DPython3_ROOT_DIR=$PYTHON_ROOT -DPython3_FIND_STRATEGY=LOCATION"
```
gadfort, line 22: Why build in a venv?

nmoroze: The issue Aulihan reported seemed related to cmake being inconsistent with regards to the Python installation it picked up depending on whether or not it's run with sudo - this makes it explicit + has the benefit of not installing orderedmultidict into a user's environment.

aulihant: Yup; I was able to get past by doing a `sudo apt-get install python3-orderedmultidict`
This is actually in the Surelog INSTALL.md, which brings up the real issue of hard-coding dependencies for Surelog into our shell script.
Does anyone have an objection with getting that command from the INSTALL.md from our setup script? I'm testing out my version of that.

nmoroze: I think the only problem might be that even if that works, you could still have the issue I noticed in your install log where Surelog gets built twice (which is fine, but takes up extra time). A side benefit of grabbing the dependency via pip+a venv is that it's less heavy handed in terms of modifying the user's environment, but it already installs a bunch of stuff from apt anyways so perhaps that's moot.

It'd be helpful to see the install log after making this change, since I wasn't able to replicate the double-build issue on my machine.

aulihant: @nmoroze Peter and I synced on this. The summary is that since we are now pointing to a specific commit of Surelog, we can make sure that the `sudo apt-get install ...` is is up-to-date. I'll submit a PR with the changes.

nmoroze: Sounds good, I'll close this PR in favor of yours!

PR #1329, opened 2023-02-24
Fix surelog install script + add exit on error to all install scripts

aulihant:


nmoroze (approved):
Looks good to me, thanks!

setup/install-surelog.sh
```diff
@@ -1,12 +1,12 @@
...
+
 # Get directory of script
 src_path=$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)
 
-sudo apt-get install -y build-essential cmake git pkg-config tclsh swig uuid-dev libgoogle-perftools-dev python3 python3-dev
-sudo apt-get install -y default-jre
-
-pip3 install orderedmultidict
+# These dependencies are up-to-date with instructions from the INSTALL.md from the commit we are pinned to below
+sudo apt-get install -y build-essential cmake git pkg-config tclsh swig uuid-dev libgoogle-perftools-dev python3 python3-orderedmultidict python3-psutil python3-dev default-jre lcov
```
WRansohoff, line 9: I'm not sure about installing PyPi packages using the Ubuntu package repositories - is there a reason that we shouldn't be consistent about installing Python package dependencies through Python's package manager?

gadfort: This makes this script consistent with the install instructions

WRansohoff: Okay, makes sense - and I guess that we wouldn't want the Surelog installation to depend on a Python virtual environment, anyways.

gadfort (approved):
Looks good to me

setup/install-surelog.sh
aulihant: We need these from: https://github.com/chipsalliance/Surelog/blob/master/INSTALL.md

aolofsson:
Nice. Does everyone agree that "-e" is always a good idea in these scripts? I am not a bash person.

http://mywiki.wooledge.org/BashFAQ/105
https://stackoverflow.com/questions/821396/aborting-a-shell-script-if-any-command-returns-a-non-zero-value

setup/install-surelog.sh
aulihant: Unfortunately, without the Ubuntu packages, Surelog will not install. I've tried with and without venvs and Surelog is still looking for the `orderedmultidict` package from the official python package area.

gadfort:
@aolofsson I think it's a good idea, we can revisit on a script by script basis, but this helps us prevent errors from getting ignored

aolofsson:
Ok, sounds good to me.

PR #1332, opened 2023-02-25
openroad fixes

gadfort:
Fixes:
- incorrect check if a value was set

Changes:
- disable write_cdl by default for now as the PDKs only have partial support for this (mainly the macros and IOs). Once those are updated we can reenable.

### DIFF
--- a/siliconcompiler/tools/openroad/openroad.py
+++ b/siliconcompiler/tools/openroad/openroad.py
@@ -234,7 +234,7 @@ def setup(chip, mode='batch'):
     for libvar, openroadvar in [('openroad_pdngen', 'pdn_config'),
                                 ('openroad_global_connect', 'global_connect')]:
         if chip.valid('tool', tool, 'task', task, 'var', openroadvar) and \
-           not chip.get('tool', tool, 'task', task, 'var', openroadvar, step=step, index=index):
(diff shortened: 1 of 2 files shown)


siliconcompiler/tools/openroad/openroad.py
```diff
@@ -234,7 +234,7 @@ def setup(chip, mode='batch'):
     for libvar, openroadvar in [('openroad_pdngen', 'pdn_config'),
                                 ('openroad_global_connect', 'global_connect')]:
         if chip.valid('tool', tool, 'task', task, 'var', openroadvar) and \
-           not chip.get('tool', tool, 'task', task, 'var', openroadvar, step=step, index=index):
+           chip.get('tool', tool, 'task', task, 'var', openroadvar, step=step, index=index):
```
WRansohoff, line 237: Could we have caught this with a test?

Also, should we have a place to record test cases which we believe could be helpful, but may not have time to implement immediately?

WRansohoff (approved):

siliconcompiler/tools/openroad/openroad.py
gadfort: @WRansohoff I'd say open a JIRA for them. I think once we fill out our CI with more actual designs these things would be easier to catch.

PR #1336, opened 2023-02-27
Add a default remote server address, if one has not been configured.

WRansohoff:
This change adds some logic to set the default server address to our public beta, if `sc-configure` has not been run.

The warning string is easy to miss if `('option', 'quiet')` is not set, since it gets printed before the local import step. I think this change should be paired with the more prominent "uploading to a public server" dialogs that we've discussed, once we agree on the phrasing of those messages.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -4134,12 +4134,13 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                 # Use the default config file path.
                 cfg_dir = os.path.join(Path.home(), '.sc')
                 cfg_file = os.path.join(cfg_dir, 'credentials')
-            if (not os.path.isdir(cfg_dir)) or (not os.path.isfile(cfg_file)):
-                self.error('Could not find remote server configuration - '
-                    'please run "sc-configure" and enter your server address and '
-                    'credentials.', fatal=True)
-            with open(cfg_file, 'r') as cfgf:
-                self.status['remote_cfg'] = json.loads(cfgf.read())
+            if os.path.isdir(cfg_dir) and os.path.isfile(cfg_file):
+                with open(cfg_file, 'r') as cfgf:
+                    self.status['remote_cfg'] = json.loads(cfgf.read())
+            else:
+                self.logger.warning('Could not find remote server configuration: defaulting to '
+                                    'https://server.siliconcompiler.com')
+                self.status['remote_cfg'] = { "address": "https://server.siliconcompiler.com" }
             if (not 'address' in self.status['remote_cfg']):
                 self.error('Improperly formatted remote server configuration - '
                     'please run "sc-configure" and enter your server address and '


siliconcompiler/core.py
```diff
@@ -4134,12 +4134,13 @@ def run(self):
...
-                    'credentials.', fatal=True)
-            with open(cfg_file, 'r') as cfgf:
-                self.status['remote_cfg'] = json.loads(cfgf.read())
+            if os.path.isdir(cfg_dir) and os.path.isfile(cfg_file):
+                with open(cfg_file, 'r') as cfgf:
+                    self.status['remote_cfg'] = json.loads(cfgf.read())
+            else:
+                self.logger.warning('Could not find remote server configuration: defaulting to '
+                                    'https://server.siliconcompiler.com')
+                self.status['remote_cfg'] = { "address": "https://server.siliconcompiler.com" }
```
gadfort, line 4143: Can we store this url in the meta data and import it so we only need to track it one place?

WRansohoff: Yes, good idea.

siliconcompiler/apps/sc_configure.py
```diff
@@ -3,6 +3,24 @@
...
+In particular, please ensure that you have the right to distribute any IP which is contained in designs that you upload to the service. This public service, provided by SiliconCompiler, is not intended to process proprietary IP.
+'''
+
+def confirm_dialog(message):
+    confirmed = False
+    while not confirmed:
+        oin = input(message)
+        if (not oin) or (oin == 'n') or (oin == 'N'):
+            print('Exiting.')
+            return 1
```
gadfort, line 19: Please use True / False

siliconcompiler/apps/sc_configure.py
```diff
@@ -3,6 +3,24 @@
...
+        elif (oin == 'y') or (oin == 'Y'):
+            confirmed = True
+    return 0
```
gadfort, line 22: same

siliconcompiler/apps/sc_configure.py
```diff
@@ -30,18 +48,14 @@ def main():
...
-                return
-            elif (oin == 'y') or (oin == 'Y'):
-                overwrite = True
+        if confirm_dialog('Overwrite existing remote configuration?'):
+            return
 
     # If a command-line argument is passed in, use that as a public server address.
     if len(sys.argv) > 1:
         print(f'Creating remote configuration file for public server: {sys.argv[1]}')
+        if 'server.siliconcompiler.com' in sys.argv[1] and confirm_dialog(tos_str):
```
gadfort, line 57: use the metadata url

siliconcompiler/apps/sc_configure.py
```diff
@@ -51,6 +65,9 @@ def main():
     username = input('Remote username (leave blank for public servers):\n').replace(" ","")
     user_pass = input('Remote password (leave blank for public servers):\n').replace(" ","")
 
+    if 'server.siliconcompiler.com' in srv_addr and confirm_dialog(tos_str):
```
gadfort, line 68: same

siliconcompiler/client.py
```diff
@@ -174,6 +174,13 @@ def request_remote_run(chip):
                    cwd=local_build_dir)
     upload_file = os.path.abspath(os.path.join(local_build_dir, 'import.tar.gz'))
 
+    # Print a reminder for public beta runs.
+    if 'server.siliconcompiler.com' in remote_run_url:
```
gadfort, line 178: same

siliconcompiler/client.py
```diff
@@ -174,6 +174,13 @@ def request_remote_run(chip):
...
+        chip.logger.warning("Your job will be uploaded to a public beta server for processing in 5 seconds.\n" \
+                            "WARNING: Please remember that the SiliconCompiler beta is not intended to process proprietary intellectual property. SiliconCompiler beta is not responsible for any proprietary intellectual property that may be uploaded." \
+                           f"Your job's reference ID is: {chip.status['jobhash']}")
```
gadfort, line 181: Maybe we should make the reference UD an `.info` instead of the warning.

siliconcompiler/client.py
```diff
@@ -174,6 +174,13 @@ def request_remote_run(chip):
                    cwd=local_build_dir)
     upload_file = os.path.abspath(os.path.join(local_build_dir, 'import.tar.gz'))
 
+    # Print a reminder for public beta runs.
+    if 'server.siliconcompiler.com' in remote_run_url:
+        chip.logger.warning("Your job will be uploaded to a public beta server for processing in 5 seconds.\n" \
+                            "WARNING: Please remember that the SiliconCompiler beta is not intended to process proprietary intellectual property. SiliconCompiler beta is not responsible for any proprietary intellectual property that may be uploaded." \
```
gadfort, line 180: remove beta

siliconcompiler/core.py
```diff
@@ -4156,7 +4156,7 @@ def run(self):
                            fatal=True)
             pre_remote_steplist = {
                 'steplist': self.get('option', 'steplist'),
-                'set': self.get('option', 'steplist', field='set'),
+                'set': self.schema._is_set(self.schema._search('option', 'steplist')),
```
gadfort, line 4159: I think this should be: `self.schema._is_set('option', 'steplist'),`

gadfort: nevermind, I see what it's doing

WRansohoff: `Schema._is_set` is a private method which accepts a config dictionary, not a keypath. `Schema._search` returns the config dictionary for a given keypath.

WRansohoff:
Like we talked about today, I'll also update the server to return a more explanatory message for successful job submissions, probably:

"Job started. Please be patient: progress will be checked and reported every 30 seconds."

gadfort:
I think you can leave out the "Please be patient"

siliconcompiler/_metadata.py
```diff
@@ -1,6 +1,10 @@
 # Version number following semver standard.
 version = '0.9.6'
 
+# Default server address for remote runs, if unspecified.
+default_server = 'https://server.siliconcompiler.com'
+default_server_name = default_server.replace('https://', '').strip('/')
```
gadfort, line 6: Maybe we can make a utility function that strips the url instead of having this in the metadata. I think it's okay to have the real URL, but the derivative seems like that should not live here.

WRansohoff: Putting it here means less code in client.py and the configure app, and we don't really need the logic to strip off these leading/trailing characters anywhere else. It's only there so that it will pick up on "server.siliconcompiler.com" and "https://server.siliconcompiler.com/", both of which would be valid entries in the credentials file.

siliconcompiler/apps/sc_configure.py
```diff
@@ -3,6 +3,26 @@
 import sys
 from pathlib import Path
 
+from siliconcompiler._metadata import default_server_name
+
+tos_str = '''Please review the SiliconCompiler cloud beta's terms of service:
```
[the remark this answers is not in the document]
WRansohoff: I think we have been calling this the "SiliconCompiler cloud beta", and as I understand it the ToS only refers to the cloud side of things, not all of SC.

siliconcompiler/client.py
```diff
@@ -174,6 +175,14 @@ def request_remote_run(chip):
                    cwd=local_build_dir)
     upload_file = os.path.abspath(os.path.join(local_build_dir, 'import.tar.gz'))
 
+    # Print a reminder for public beta runs.
+    if default_server_name in remote_run_url:
+        chip.logger.warning("Your job will be uploaded to a public server for processing in 5 seconds.\n")
+        chip.logger.warning("WARNING: Please remember that the SiliconCompiler public servers are not intended to process proprietary intellectual property.")
+        chip.logger.warning("SiliconCompiler is not responsible for any proprietary intellectual property that may be uploaded.\n")
```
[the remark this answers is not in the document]
WRansohoff: It's hard to notice this text in the log, because it comes right after the local import step. The newlines add a break in the logger output which increases the odds that people will notice that they're supposed to read the message.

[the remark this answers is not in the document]
gadfort: I don't think they should be there. The timeout should serve as enough of a break to notice it

siliconcompiler/apps/sc_configure.py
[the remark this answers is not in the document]
gadfort: I see

siliconcompiler/client.py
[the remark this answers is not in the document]
WRansohoff: Respectfully, I disagree. I know that the message is supposed to be there, and I have trouble noticing the message when it is printed as a block, even with the delay. It just feels like the import step is taking a bit longer.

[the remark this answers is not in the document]
aulihant: I think it's totally fair that, in order to be readable, warning messages should be short and to the point - i.e. one line.
However, I think the main sentence that should be prefixed with the `WARNING` message is actually the second sentence.

So either of the following two would be fine for me:
Option 1
`WARNING: Please remember that the SiliconCompiler public servers are not intended to process proprietary intellectual property.`
`WARNING: SiliconCompiler is not responsible for any proprietary intellectual property that may be uploaded.`

or

Option 2
`Please remember that the SiliconCompiler public servers are not intended to process proprietary intellectual property.`
`WARNING: SiliconCompiler is not responsible for any proprietary intellectual property that may be uploaded.`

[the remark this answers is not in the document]
WRansohoff: Okay, here is what the current message looks like. How should it be modified, if at all?

```
[...surelog import task messages....]
[ SYNTAX] : 0
[  ERROR] : 0
[WARNING] : 1
[   NOTE] : 5
| WARNING | job0  | import     | 0  | [WRN:PA0205] /[...filepath...].v:1: No timescale set for "heartbeat".
| INFO    | job0  | import     | 0  | Finished task in 0.76s
| WARNING | job0  | import     | 0  | Your job will be uploaded to a public server for processing in 5 seconds.

| WARNING | job0  | import     | 0  | Please remember that the SiliconCompiler public servers are not intended to process proprietary intellectual property.
| WARNING | job0  | import     | 0  | SiliconCompiler is not responsible for any proprietary intellectual property that may be uploaded.

| INFO    | job0  | import     | 0  | Your job's reference ID is: [uuid]```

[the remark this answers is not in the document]
gadfort: I think it's sufficient as is. We can always circle back later if we decide to update the language.

gadfort (approved):

siliconcompiler/_metadata.py
gadfort: okay, we'll address this in a followon PR

PR #1337, opened 2023-02-28
Switch tools to be handled by python imports

gadfort:
Changes:
- adds a 'module' parameter to the schema to track the actual python module for the tool
- `.node()` takes in a module for the tool parameter, when using builtin tools, it will take in 'siliconcompiler'
- ensures `load_target` can be used outside of siliconcompiler modules


siliconcompiler/flows/showflow.py
```diff
@@ -29,16 +31,22 @@ def setup(chip, flowname='showflow', filetype=None, screenshot=False, np=1):
...
-    flow.node(flowname, 'import', 'builtin', 'import')
+    flow.node(flowname, 'import', siliconcompiler, 'import')
 
+    if filetype not in chip.getkeys('option', 'showtool'):
+        raise SiliconCompilerError(f'Show tool for {filetype} is not defined.')
     show_tool = chip.get('option', 'showtool', filetype)
 
+    show_tool_module = chip._lookup_toolmodule_by_name(show_tool)
+    if not show_tool_module:
+        raise SiliconCompilerError(f'Cannot determine tool module for {show_tool}.')
```
nmoroze, line 42: I think one pitfall is this requires the chip object already has a flow set up that includes a node using the show tool. Maybe fall back to doing a target-style search of `[name, f'siliconcompiler.tools.{name}', f'siliconcompiler.tools.{name}.{name}']`?

tests/core/test_check_manifest.py
```diff
@@ -5,6 +5,11 @@
 
 import pytest
 
+from tests.core.tools import foo
+from tests.core.tools import bar
+from tests.core.tools import baz
+from tests.core.tools.echo import echo
```
nmoroze, line 11: Can probably just move echo out of its containing directory to simplify path. (also it's nice that we can finally remove this from our actual tools directory!)

tests/core/test_check_manifest.py
```diff
@@ -102,13 +106,13 @@ def merge_flow_chip():
     chip = siliconcompiler.Chip('test')
 
     flow = 'test'
-    chip.node(flow, 'import', 'builtin', 'import')
-    chip.node(flow, 'parallel1', 'foo', 'parallel1')
-    chip.node(flow, 'parallel2', 'bar', 'parallel2')
+    chip.node(flow, 'import', siliconcompiler, 'import')
+    chip.node(flow, 'parallel1', foo, 'parallel1')
+    chip.node(flow, 'parallel2', bar, 'parallel2')
```
nmoroze, line 111: I'm pretty sure we could use "echo" here as well in place of foo/bar/baz in order to avoid creating the empty files

siliconcompiler/core.py
```diff
@@ -3106,29 +3048,45 @@ def node(self, flow, step, tool, task, index=0):
...
             return
 
         index = str(index)
 
+        if (not inspect.ismodule(tool)):
+            self.error(f"{tool} is not a module and cannot be used to setup a tool.", fatal=True)
+
+        module_name = tool.__name__
+
+        import siliconcompiler as sc
```
nmoroze, line 3080: I assume putting this at the top causes cyclic import issues, but might be good to tack on a comment since mid-body imports always seem weird.

siliconcompiler/flows/showflow.py
nmoroze: Although we might need to invert the priority on that search. Otherwise, if you happen to have the KLayout standalone package installed, "klayout" would match on that rather than the SC tool driver.

nmoroze (changes requested):
I really like where this is going! I noted a couple small things, a few high level comments: 

I think it would be helpful to support task modules that live outside of the tool module's package. Maybe the signature for `node()` could be adjusted to support a string or module for task, and if the task is a module we can just use that (and add a place in the schema to stash the full import path of the task). I'm also totally fine with this functionality being added in a separate PR. 

We may also want to make `node()` accept a full import path string for tools as an alternative to supplying a module. In theory, I think you could have a setup with a remote server where you don't have the tool modules installed locally, since the setup() method for non-import tools only gets run on the server. Allowing a string here would let us continue to support that sort of setup (granted, we can also punt on this since one could always set the schema directly rather than using `node()`, but I think it would be nice to support this in node as well).

siliconcompiler/core.py
gadfort: yes

aolofsson:
So do we actually need both the tool and tool module? Looks like [-1] index is used to resolve he tool from the toolmodule. Why not go all in on modules and resolve toolname with an internal function when needed?

aolofsson:
On second thought, this is an example of Python leaking into the schema, honestly this doesn't feel right.

gadfort:
> On second thought, this is an example of Python leaking into the schema, honestly this doesn't feel right.

lets plan to discuss this tomorrow

gadfort:
Closed in favor of #1426 1426

PR #1340, opened 2023-02-28
ensure workflow scripts point at the base siliconcompiler directory

### LINKED ISSUE #1338: suspected bug in .github/workflows/bin/install_surelog_macos.sh
https://github.com/siliconcompiler/siliconcompiler/blob/main/.github/workflows/bin/install_surelog_macos.sh#L9

`git checkout` will fail because after `cd surelog` working directory changed ==> `python3 setup/_tools.py` file not found.

```sh
# Install Surelog
git clone $(python3 setup/_tools.py --tool surelog --field git-url) surelog
cd surelog
git checkout $(python3 setup/_tools.py --tool surelog --field git-commit)
```

gadfort:
Closes #1338 

We should trigger the surelog build before merging, probably next week

### DIFF
--- a/.github/workflows/bin/install_surelog_macos.sh
+++ b/.github/workflows/bin/install_surelog_macos.sh
@@ -1,12 +1,15 @@
 #!/bin/bash
 
+# Get directory of setup scripts
+src_path=$(cd -- "$(dirname "$0")/../../../" >/dev/null 2>&1 ; pwd -P)
+
 # Install dependencies
 pip3 install orderedmultidict
 
 # Install Surelog
-git clone $(python3 setup/_tools.py --tool surelog --field git-url) surelog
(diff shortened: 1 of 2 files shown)


nmoroze:
BTW, the macOS half of this change has effectively been tested by https://github.com/siliconcompiler/siliconcompiler/actions/runs/4309851810, which has the caching logic removed. The Linux build doesn't actually cache Surelog (for reasons...), so I'm running a workflow to test that the Linux build still works: https://github.com/siliconcompiler/siliconcompiler/actions/runs/4315760176. 

I think if that passes we're pretty safe to merge this, although I'll leave that up to you.

gadfort:
We'll merge this and the other PR assuming it passes the wheels build today.

PR #1341, opened 2023-03-01
Self-test target

WRansohoff:
Like we've been discussing, it would be nice to have a quick self-test to make sure that everything is installed and configured correctly.

I copied the heartbeat sources into a subdirectory under `targets/` so that they should be included in the wheels build. I didn't delete the files under `examples/heartbeat/`, because they are referenced in many places. This change could also use a CI test or two when we have time.

But this should let us run an easy power-on test (along with #1336 for the remote flow):

    pip install siliconcompiler
    sc -target asic_demo [-remote]


siliconcompiler/targets/asic_demo.py
```diff
@@ -0,0 +1,31 @@
...
+    chip.load_target('skywater130_demo')
+
+    # Set die area.
+    chip.set('constraint', 'outline', [(0, 0), (50, 50)])
+    chip.set('constraint', 'corearea', [(5, 5), (45, 45)])
+
+    # Set design name and source files.
+    chip.set('design', design)
+    src_prefix = os.path.join(os.path.dirname(__file__), 'asic_demo', design)
+    for suffix in ['.v', '.sdc']:
```
gadfort, line 24: if all we are setting is the clock in the SDC, I don't see why we don't just use the `.clock(` instead?

siliconcompiler/targets/asic_demo.py
```diff
@@ -0,0 +1,31 @@
+import os
```
gadfort, line 1: This needs be added to the quick CI

gadfort (changes requested):

siliconcompiler/core.py
```diff
@@ -4153,9 +4153,10 @@ def run(self):
                 self.error('Cannot pass "-step" parameter into remote flow. A steplist including '
                            'an "import" task and at least one EDA task is required.',
                            fatal=True)
+            cur_steplist = self.get('option', 'steplist')
             pre_remote_steplist = {
-                'steplist': self.get('option', 'steplist'),
-                'set': self.get('option', 'steplist', field='set'),
+                'steplist': cur_steplist,
+                'set': self.get('option', 'steplist', field='set') if cur_steplist else False,
```
nmoroze, line 4159: The "set" field no longer exists as of https://github.com/siliconcompiler/siliconcompiler/pull/1311 (based on some feedback to make the handling of values more consistent). I missed this when refactoring though, my bad.

For a replacement, it's probably best to use the `Schema._is_set(...)` helper function. That takes in a cfg dictionary right now, but I think it would make sense to add a wrapper around it to take in a keypath, call `_search()` to grab the dict, and pass that in.

siliconcompiler/targets/asic_demo.py
```diff
@@ -0,0 +1,31 @@
...
+    "Self-test" target which builds a small 8-bit counter design as an ASIC,
+    targeting the Skywater130 PDK.
+
+    This target is not intended for general-purpose use.
+    It is intended to quickly verify that SiliconCompiler is installed and configured correctly.
+    '''
+
+    # Load the Sky130 PDK/standard cell library target.
+    design = 'heartbeat'
+    chip.load_target('skywater130_demo')
```
nmoroze, line 15: This makes perfect sense to me so there's no actionable item for this PR, but loading a target-within-a-target does cause ['option', 'target'] to get overwritten with the last one. I just wanted to flag that I opened a ticket to track how we want to handle that, for future discussion (SC-194).

nmoroze (changes requested):
Flagged a couple small things, but overall looking good! It'll be nice to have the easy self-test.

siliconcompiler/core.py
WRansohoff: Okay, I'll give that a try.

siliconcompiler/targets/asic_demo.py
WRansohoff: We're coming down to the wire on release timing, but sure, shouldn't take too long.

siliconcompiler/targets/asic_demo/heartbeat.v
```diff
@@ -0,0 +1,13 @@
+module heartbeat #(parameter N = 8)
```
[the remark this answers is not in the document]
WRansohoff: Okay, thanks. I'll definitely test this in a wheels build before merging.

siliconcompiler/targets/asic_demo.py
WRansohoff: Yeah, it's a little bit of a workaround, but that ends up working out in this case because the sky130 target is what we want for the build.

siliconcompiler/targets/asic_demo.py
WRansohoff: It worked, thanks. I should probably also update the picorv32 example/tutorial to do this.

nmoroze (approved):
LGTM!

gadfort (approved):

PR #1342, opened 2023-03-01
Add a LaTeX-based PDF report, improve reports for failed runs. 

WRansohoff:
It's nice that our current HTML reports are interactive, but their styling could use some work. With L<sup>a</sup>T<sub>e</sub>X, it's a bit easier to generate clean-looking static PDF reports. Here's an example report for a PicoRV32 design on skywater130:

image

I also moved the `_deferstep` call a bit further down in the `_runtask` method, so that it comes after the manifest check. This lets the clustered runs fail fast when a task fails partway through a run, and makes it easier to include metrics up to the point of failure in the reports. I don't think any of the logic in between is particularly consequential for the purposes of distributing tasks to a scheduler.

Going up to a 4k resolution for screenshots didn't seem to cause any problems with headless KLayout, and we can generate PDF reports with large page sizes to show that level of detail.

I added some kwargs to the `chip.summary()` method to set whether it should generate a screenshot, HTML report, and/or PDF report. The PDF report is disabled by default, because it requires some texlive packages. Some CI unit tests would probably be a good idea for this new functionality.


gadfort:
@WRansohoff lets discuss this

WRansohoff:
Okay, but it's a very easy win for server-side rendering and has almost no changes to how the core SC code is executed.

WRansohoff:
After talking offline, we're going to omit the LaTeX template for now, and only keep the logic to open PDF files in the results.

For the upcoming release, we'll implement PDF generation on the server side only. That will give us more time to look into RST as a format for PDF generation, while still letting us provide a cleaned-up static report.

WRansohoff:
Okay, the generation logic in [f71d6a0](https://github.com/siliconcompiler/siliconcompiler/pull/1342/commits/f71d6a045c7e22e5d06e8866bf364f0a0be32d8c) will only be present on the servers, so that we don't enshrine LaTeX as the project's officially blessed PDF generator without further review.

siliconcompiler/core.py
```diff
@@ -2952,57 +2953,80 @@ def summary(self, steplist=None, show_all_indices=False):
...
-                except webbrowser.Error:
-                    # Python 'webbrowser' module includes a limited number of popular defaults.
-                    # Depending on the platform, the user may have defined their own with $BROWSER.
-                    if 'BROWSER' in os.environ:
-                        subprocess.run([os.environ['BROWSER'], results_page])
+            # Prioritize PDF, followed by HTML.
+            self.logger.warning(results_pdf)
+            if (not self.get('option', 'nodisplay')) and ('DISPLAY' in os.environ):
+                if os.path.isfile(results_pdf):
+                    self.logger.warning('EXISTS')
```
nmoroze, line 3007: Log left over from debugging?

siliconcompiler/core.py
```diff
@@ -4308,7 +4327,7 @@ def run(self):
                 index_succeeded = False
                 for index in indexlist[step]:
                     stepstr = step + index
-                    if status[stepstr] != TaskStatus.ERROR:
+                    if stepstr in status and status[stepstr] != TaskStatus.ERROR:
```
nmoroze, line 4330: What prompted this extra check? Seems like this could only be False if a step in the steplist or index in the indexlist is not part of the flowgraph, which would be problematic.

siliconcompiler/core.py
```diff
@@ -2771,7 +2771,8 @@ def _find_leaves(self, steplist):
         return all_tasks.difference(non_leaf_tasks)
 
     ###########################################################################
-    def summary(self, steplist=None, show_all_indices=False):
+    def summary(self, steplist=None, show_all_indices=False,
+                generate_screenshot=True, generate_html=True, generate_pdf=False):
```
nmoroze, line 2775: Perhaps remove `generate_pdf` for now if it's currently non-functional, to prevent user confusion.

nmoroze (commented):
Just a few small comments. Server-side PDF rendering seems nice given Latex is a fairly heavy dependency.

siliconcompiler/core.py
```diff
@@ -2952,57 +2953,76 @@ def summary(self, steplist=None, show_all_indices=False):
...
-                    pruned_cfg = pruned_cfg,
-                    metric_keys = metrics_to_show,
-                    img_data = img_data,
-                ))
+            # Generate reports by passing the Chip manifest into the Jinja2 template.
+            results_html = os.path.join(web_dir, 'report.html')
+            results_pdfsrc = os.path.join(web_dir, 'report.tex')
+            results_pdf = os.path.join(web_dir, 'report.pdf')
+            if generate_html or generate_pdf:
+                env = Environment(loader=FileSystemLoader(templ_dir))
```
gadfort, line 2971: remove PDF

siliconcompiler/core.py
```diff
@@ -2952,57 +2953,76 @@ def summary(self, steplist=None, show_all_indices=False):
...
+                    self.logger.warning('EXISTS')
+                    # Open results with whatever application is associated with PDFs on the local system.
+                    if sys.platform == 'win32':
+                        os.startfile(results_pdf)
+                    elif sys.platform == 'darwin':
+                        subprocess.Popen(['open', results_pdf])
                     else:
-                        self.logger.warning('Unable to open results page in web browser:\n' +
-                                            os.path.abspath(os.path.join(web_dir, "report.html")))
+                        subprocess.Popen(['xdg-open', results_pdf])
```
gadfort, line 3014: why not use startfile for all? seems like python recommends that

siliconcompiler/core.py
```diff
@@ -2952,57 +2953,76 @@ def summary(self, steplist=None, show_all_indices=False):
...
-                except webbrowser.Error:
-                    # Python 'webbrowser' module includes a limited number of popular defaults.
-                    # Depending on the platform, the user may have defined their own with $BROWSER.
-                    if 'BROWSER' in os.environ:
-                        subprocess.run([os.environ['BROWSER'], results_page])
+            # Prioritize PDF, followed by HTML.
+            self.logger.warning(results_pdf)
+            if (not self.get('option', 'nodisplay')) and ('DISPLAY' in os.environ):
+                if os.path.isfile(results_pdf):
+                    self.logger.warning('EXISTS')
```
gadfort, line 3007: remove PDF stuff

siliconcompiler/core.py
```diff
@@ -3639,6 +3646,18 @@ def _runtask(self, step, index, status):
...
+        # If the job is configured to run on a cluster, collect the schema
+        # and send it to a compute node for deferred execution.
+        # (Run the initial 'import' stage[s] locally)
+        if self.get('option', 'scheduler', 'name', step=step, index=index) and \
+           self.get('flowgraph', flow, step, index, 'input'):
+            # Note: The _deferstep method blocks until the compute node
+            # finishes processing this step, and it sets the active/error bits.
+            _deferstep(self, step, index, status)
+            return
+
```
gadfort, line 3660: Update the docstring to reflect the change in order of operations

siliconcompiler/core.py
```diff
@@ -3525,19 +3545,6 @@ def _runtask(self, step, index, status):
...
-        # and send it to a compute node for deferred execution.
-        # (Run the initial 'import' stage[s] locally)
-
-        if self.get('option', 'scheduler', 'name', step=step, index=index) and \
-           self.get('flowgraph', flow, step, index, 'input'):
-            # Note: The _deferstep method blocks until the compute node
-            # finishes processing this step, and it sets the active/error bits.
-            _deferstep(self, step, index, status)
-            return
-
```
gadfort, line 3540: Whats the benefit of moving?

siliconcompiler/core.py
WRansohoff: Argh, I thought I caught that. Thanks.

siliconcompiler/core.py
WRansohoff: Sure, can't hurt.

siliconcompiler/core.py
WRansohoff: It's probably safe to remove this. I briefly tried adding a 'screenshot' task to the steplist on the server side, which raised some issues around here, but that was a dead end.

siliconcompiler/tools/klayout/screenshot.py
```diff
@@ -27,8 +27,8 @@ def setup(chip):
     chip.set('tool', tool, 'task', task, 'option', option, step=step, index=index, clobber=clobber)
 
     chip.add('tool', tool, 'task', task, 'output', design + '.png', step=step, index=index)
-    chip.set('tool', tool, 'task', task, 'var', 'show_horizontal_resolution', '1024', step=step, index=index, clobber=False)
-    chip.set('tool', tool, 'task', task, 'var', 'show_vertical_resolution', '1024', step=step, index=index, clobber=False)
+    chip.set('tool', tool, 'task', task, 'var', 'show_horizontal_resolution', '4096', step=step, index=index, clobber=False)
+    chip.set('tool', tool, 'task', task, 'var', 'show_vertical_resolution', '4096', step=step, index=index, clobber=False)
```
gadfort, line 31: Won't this generate a huge file? If more resolution is needed, maybe we can split the difference and do 2048x2048?
The HTML report size should be checked to make sure its an okay filesize

gadfort (changes requested):

siliconcompiler/core.py
WRansohoff: The call only exists on windows.

siliconcompiler/core.py
WRansohoff: We do want this logic, so that the client will open the PDF which was generated on the server side. I removed the messages though

siliconcompiler/tools/klayout/screenshot.py
WRansohoff: The picorv32+RAM example is <5MB.

gadfort: How big is the HTML?

WRansohoff: About the same size.

siliconcompiler/core.py
```diff
@@ -2952,57 +2953,69 @@ def summary(self, steplist=None, show_all_indices=False):
...
-                    metrics = metrics,
-                    reports = reports,
-                    manifest = self.schema.cfg,
-                    pruned_cfg = pruned_cfg,
-                    metric_keys = metrics_to_show,
-                    img_data = img_data,
-                ))
+            # Generate reports by passing the Chip manifest into the Jinja2 template.
+            results_html = os.path.join(web_dir, 'report.html')
+            results_pdfsrc = os.path.join(web_dir, 'report.tex')
```
gadfort, line 2968: left over

PR #1344, opened 2023-03-01
Implement OpenROAD padring

gadfort:
Changes:
- Adds padring script access to the floorplanning step in OR
- Bumps the required version of OpenROAD (@WRansohoff can we work on getting the CI running the new version)

### DIFF
--- a/siliconcompiler/libs/sky130io.py
+++ b/siliconcompiler/libs/sky130io.py
@@ -34,6 +34,11 @@ def setup(chip):
     lib.add('output', stackup, 'gds', os.path.join(libdir, 'sky130_fd_io.gds'))
     lib.add('output', stackup, 'gds', os.path.join(libdir, 'sky130_ef_io__gpiov2_pad_wrapped.gds'))
 
+    lib.set('asic', 'cells', 'filler', ['sky130_ef_io__com_bus_slice_1um',
+                                        'sky130_ef_io__com_bus_slice_5um',
+                                        'sky130_ef_io__com_bus_slice_10um',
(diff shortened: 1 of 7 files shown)


WRansohoff:
Congrats on getting this merged into OpenROAD! I started building the new version on the CI runner, and will retry the tests when it's done.

WRansohoff:
Update is done, looks like it is down to one test failure, on the number of warnings in `test_gcd` @gadfort

gadfort:
> Update is done, looks like it is down to one test failure, on the number of warnings in `test_gcd` @gadfort 

Awesome. I'll fix that this evening

siliconcompiler/libs/sky130io.py
```diff
@@ -34,6 +34,11 @@ def setup(chip):
     lib.add('output', stackup, 'gds', os.path.join(libdir, 'sky130_fd_io.gds'))
     lib.add('output', stackup, 'gds', os.path.join(libdir, 'sky130_ef_io__gpiov2_pad_wrapped.gds'))
 
+    lib.set('asic', 'cells', 'filler', ['sky130_ef_io__com_bus_slice_1um',
```
WRansohoff, line 37: Neat use of the fill cell parameter - I'm guessing these are the cells that fill the gaps between individual pads?

WRansohoff (approved):

siliconcompiler/libs/sky130io.py
gadfort: yeah, this will only work on older nodes where the fillers are the same for the different pad orientations

PR #1345, opened 2023-03-01
add release version to git clone commands

aulihant:
Per discussion with @gadfort, I added the release version to the `git clone` commands. Also a minor update to developer install instructions to be consistent with other install instructions

### DIFF
--- a/docs/user_guide/installation.rst
+++ b/docs/user_guide/installation.rst
@@ -168,13 +168,13 @@ installs packages on your machine.
 
 Finally, to clone and install SiliconCompiler, run the following:
 
-.. code-block:: bash
+.. parsed-literal::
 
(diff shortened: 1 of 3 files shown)


aulihant:
@gadfort the only other file which has `git clone` of siliconcompiler in the docs is `docs/tutorials/zerosoc.rst`, which I assume you're working with? If not, I can make the update in there too.

gadfort:
> @gadfort the only other file which has `git clone` of siliconcompiler in the docs is `docs/tutorials/zerosoc.rst`, which I assume you're working with? If not, I can make the update in there too.

I'll deal with zerosoc

gadfort (approved):
Awesome

PR #1348, opened 2023-03-02
adding generator to summarize number of parameters in schema groups

gadfort:


nmoroze (approved):
Seems much better to do it this way, thanks!

PR #1354, opened 2023-03-03
Remove recently-added DISPLAY check, and an extraneous file path def

WRansohoff:
Quick follow-up which lets the PDF results returned from the remote server be opened automatically on Windows machines after a successful run. The `$DISPLAY` check that I added in #1342  turned out to be unnecessary, and not applicable to Windows.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -2964,10 +2964,11 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                     with open(screenshot_path, 'rb') as img_file:
                         img_data = base64.b64encode(img_file.read()).decode('utf-8')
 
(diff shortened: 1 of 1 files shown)


siliconcompiler/core.py
```diff
@@ -2997,7 +2998,7 @@ def summary(self, steplist=None, show_all_indices=False,
 
             # Try to open the results and layout only if '-nodisplay' is not set.
             # Prioritize PDF, followed by HTML.
-            if (not self.get('option', 'nodisplay')) and ('DISPLAY' in os.environ):
+            if (not self.get('option', 'nodisplay')):
```
gadfort, line 3001: I think you still need this check for linux. Maybe the thing to do is to check that the os is windows or DISPLAY in os.environ

gadfort (approved):

PR #1360, opened 2023-03-04
Fix typo in description in schema_cfg.py

aulihant:


gadfort (approved):
Thanks for spotting that

PR #1363, opened 2023-03-06
Fix clock in asic_demo

gadfort:
Changes:
- clock getting set before design is set
- add checks in unit tests for setup and hold slack
- updates area since the server seems diverge in global place


WRansohoff (approved):
60/55 might be safer than 55/50, but both work for me and I trust your judgement on how deterministic the occurrence of replace divergences are.

PR #1367, opened 2023-03-06
move _collect call to after import to prevent multiple local files from getting sourced in surelog

gadfort:
Fixes:
- imported files getting found multiple times by surelog with include paths. Since import always runs locally, we do not need to run this until after the import step has completed.

### DIFF
--- a/siliconcompiler/client.py
+++ b/siliconcompiler/client.py
@@ -82,6 +82,12 @@ def remote_preprocess(chip, steplist):
         # step should have none.
         chip._runtask(local_step, index, {})
 
+        # Collect inputs into import directory only for remote runs, since
+        # we need to send inputs up to the server. Otherwise, it's simpler
+        # for debugging to leave inputs in place.
+        # TODO: if more than one import is present this might not work
+        chip._collect(local_step, index)
+
     # Set 'steplist' to only the remote steps, for the future server-side run.
     chip.unset('arg', 'step')
     chip.unset('arg', 'index')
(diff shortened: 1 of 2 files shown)


WRansohoff:
Ohh, I think I may have seen this problem (or something similar) a while back. We should probably also make an issue to track the case where a design includes multiple identical filenames (e.g. `adder.v`) in different levels of its hierarchy.

If I'm understanding the problem correctly, could we update `_collect` (which already gets called in the remote import `_runtask()` method) to only copy files if they are not already present? Something like changing this:

```
        for path in self._collect_paths():
            filename = self._get_imported_filename(path)
            abspath = self._find_sc_file(path)
            if abspath:
                self.logger.info(f"Copying {abspath} to '{indir}' directory")
                shutil.copy(abspath, os.path.join(indir, filename))
            else:
                self._haltstep(step, index)
```

To this:

```
        for path in self._collect_paths():
            filename = self._get_imported_filename(path)
            abspath = self._find_sc_file(path)
            destpath = os.path.join(indir, filename)
            if abspath:
                if os.path.isfile(destpath):
                    self.logger.warning("[duplicate filename message]")
                else:
                    self.logger.info(f"Copying {abspath} to '{indir}' directory")
                    shutil.copy(abspath, destpath)
            else:
                self._haltstep(step, index)
```

WRansohoff (approved):
After talking offline, this sounds like a reasonable solution.

PR #1370, opened 2023-03-07
Add summary image generation feature

nmoroze:
This PR adds a feature to `summary()` which generates an image including a layout screenshot + a subset of metrics (design name, node, area, fmax).

Per Peter's suggestion, this PR also removes the separate execution of a screenshot flow from `summary()`, and instead generates screenshots as part of the KLayout export step in the asicflow. I refactored the KL scripts a bit to make the show functionality importable from `klayout_export.py` (the screenshot is generated by re-reading the layout from GDS - I tried to use the actual Layout object in-memory directly, but I was having trouble getting a good result that corresponds with our normal show display).

I recall there being some concern that this wouldn't work on headless systems, but that appears to be fixed in KLayout 0.28: https://github.com/KLayout/klayout/issues/1082. I've updated the minimum version required by our driver correspondingly.

This PR adds a dependency on Pillow.

I expect the exact design parameters will want some discussion, I figure we can iterate on that offline.


siliconcompiler/core.py
```diff
@@ -2771,9 +2773,101 @@ def _find_leaves(self, steplist):
...
+        }
+
+        if self.get('option', 'pdk'):
+            metrics['Node'] = self.get('option', 'pdk')
+
+        # TODO: a bit hardcoded to asicflow assumptions... a way to query
+        # "final" metrics regardless of flow would be handy
+        totalarea = self.get('metric', 'totalarea', step='export', index='1')
+        if totalarea:
+            metrics['Area'] = f'{int(totalarea)} um^2'
```
gadfort, line 2794: can this be ``{:.2f}``

siliconcompiler/core.py
```diff
@@ -2771,9 +2773,101 @@ def _find_leaves(self, steplist):
...
+        fmax = self.get('metric', 'fmax', step='export', index='1')
+        if fmax:
+            # SI-ify
+            if fmax < 1e6:
+                fmax = f'{round(fmax / 1e3, 2)} KHz'
+            elif fmax < 1e9:
+                fmax = f'{round(fmax / 1e6, 2)} MHz'
+            else:
+                fmax = f'{round(fmax / 1e9, 2)} GHz'
+            metrics['Fmax'] = fmax
```
gadfort, line 2805: this should also be ``{:.2f}`` round(,2) doesn't always end up with just 2 digits

siliconcompiler/core.py
```diff
@@ -2771,9 +2773,101 @@ def _find_leaves(self, steplist):
...
+        BORDER = 16
+        LINE_SPACING = 4
+        TEXT_INDENT = 8
+
+        FONT_PATH = os.path.join(self.scroot, 'data', 'RobotoMono', 'RobotoMono-Regular.ttf')
+        FONT_SIZE = 24
+
+        # harcoded to match BG color configured in klayout_show.py
+        BG_COLOR = (33, 33, 33)
+        TEXT_COLOR = (224, 224, 224)
```
gadfort, line 2819: Can you add a comment about what the colors are (approx)

siliconcompiler/core.py
```diff
@@ -2771,9 +2773,101 @@ def _find_leaves(self, steplist):
...
+
+        original_layout = Image.open(input_path)
+        orig_width, orig_height =original_layout.size
```
gadfort, line 2822: ``=original_layout.size`` add space ``= original_layout.size``

siliconcompiler/core.py
```diff
@@ -2771,9 +2773,101 @@ def _find_leaves(self, steplist):
...
+
+        aspect_ratio = orig_height / orig_width
+
+        # inset by border left and right
+        thumbnail_width = WIDTH - 2 * BORDER
+        thumbnail_height = int(thumbnail_width * aspect_ratio)
```
gadfort, line 2828: round()

siliconcompiler/core.py
```diff
@@ -2771,9 +2773,101 @@ def _find_leaves(self, steplist):
...
+        layout_thumbnail = original_layout.resize((thumbnail_width, thumbnail_height))
+
+        font = ImageFont.truetype(FONT_PATH, FONT_SIZE)
+        _, descent = font.getmetrics()
+
+        # Get max height of any ASCII character in font, so we can consistently space each line
+        line_height = font.getmask(string.printable).getbbox()[3] + descent
```
gadfort, line 2835: comment why 4th entry? is 2nd entry always 0?

siliconcompiler/core.py
```diff
@@ -2771,9 +2773,101 @@ def _find_leaves(self, steplist):
...
+            # shorten line till it fits
+            cropped_line = line
+            while True:
+                line_width = font.getmask(cropped_line).getbbox()[2] + TEXT_INDENT
+                if x + line_width < (WIDTH - BORDER):
+                    break
+                cropped_line = cropped_line[:-1]
+
+            if cropped_line != line:
+                print(f'WARNING: cropped {line} to {cropped_line} to fit')
```
gadfort, line 2852: make this logger

siliconcompiler/core.py
```diff
@@ -2771,9 +2773,101 @@ def _find_leaves(self, steplist):
...
+            metrics['Fmax'] = fmax
+
+        # Generate design
+
+        WIDTH = 512
+        BORDER = 16
+        LINE_SPACING = 4
+        TEXT_INDENT = 8
+
+        FONT_PATH = os.path.join(self.scroot, 'data', 'RobotoMono', 'RobotoMono-Regular.ttf')
```
gadfort, line 2814: Do subfolders get included?

siliconcompiler/tools/klayout/klayout_export.py
```diff
@@ -268,3 +269,6 @@ def read_fills(top):
 
 gds_export(design, in_def, in_gds, out_gds, tech_file, foundry_lef, macro_lefs,
            config_file='', seal_file='', timestamps=sc_timestamps)
+
+# Generate screenshot
+show(schema, out_gds, f'outputs/{design}.png', screenshot=True)
```
gadfort, line 274: might be nice to wrap this with a generate_screenshot var (default to true), but allow us to control

siliconcompiler/tools/klayout/export.py
```diff
@@ -47,3 +46,10 @@ def setup(chip):
     # Export GDS with timestamps by default.
     chip.set('tool', tool, 'task', task, 'var', 'timestamps', 'true', step=step, index=index, clobber=False)
     chip.set('tool', tool, 'task', task, 'var', 'timestamps', 'Export GDSII with timestamps', field='help')
+
```
gadfort, line 49: Add png as output

siliconcompiler/tools/klayout/klayout_show.py
```diff
@@ -3,133 +3,147 @@
...
+        if (layer_name in sc_hide_layers) or (layer_ldt in sc_hide_layers):
+            layer.visible = False
+
+    # If 'screenshot' mode is set, save image and exit.
+    if screenshot:
+        # Save a screenshot. TODO: Get aspect ratio from sc_cfg?
+        # horizontal_resolution = int(schema.get('tool', 'klayout', 'task', task, 'var', 'show_horizontal_resolution', step=step, index=index)[0])
+        # vertical_resolution = int(schema.get('tool', 'klayout', 'task', task, 'var', 'show_vertical_resolution', step=step, index=index)[0])
+        horizontal_resolution = 4096
+        vertical_resolution = 4096
```
gadfort, line 114: why not get it from the schema

gadfort (changes requested):

siliconcompiler/tools/klayout/klayout_show.py
nmoroze: Left over from debugging, my bad - nice catch

siliconcompiler/core.py
nmoroze: apparently this breaks our current setup.py config (blocking CI). Should be fixable though

siliconcompiler/core.py
nmoroze: I don't actually have proof that it is, so I'll just do out the subtraction

siliconcompiler/tools/klayout/export.py
```diff
@@ -47,3 +46,10 @@ def setup(chip):
...
+    chip.set('tool', tool, 'task', task, 'var', 'show_horizontal_resolution', '4096', step=step, index=index, clobber=False)
+    chip.set('tool', tool, 'task', task, 'var', 'show_vertical_resolution', '4096', step=step, index=index, clobber=False)
+
+    # Help
+    chip.set('tool', tool, 'task', task, 'var', 'show_horizontal_resolution', 'Horizontal resolution in pixels', field='help')
```
gadfort, line 54: I think you also need to add: https://github.com/siliconcompiler/siliconcompiler/blob/eb8856b41df1bf096ad0fcb45c43c6c3b9fa650d/siliconcompiler/tools/klayout/show.py#L22
Otherwise the hide layers don't get hidden

siliconcompiler/core.py
```diff
@@ -2791,17 +2791,17 @@ def _generate_summary_image(self, input_path, output_path):
         # "final" metrics regardless of flow would be handy
         totalarea = self.get('metric', 'totalarea', step='export', index='1')
         if totalarea:
-            metrics['Area'] = f'{int(totalarea)} um^2'
+            metrics['Area'] = f'{totalarea:.2f} um^2'
```
gadfort, line 2794: I meant to ask, should we scale this to mm^2 when it gets too big?

siliconcompiler/tools/klayout/export.py
```diff
@@ -47,9 +47,19 @@ def setup(chip):
     chip.set('tool', tool, 'task', task, 'var', 'timestamps', 'true', step=step, index=index, clobber=False)
     chip.set('tool', tool, 'task', task, 'var', 'timestamps', 'Export GDSII with timestamps', field='help')
 
+    chip.set('tool', tool, 'task', task, 'var', 'screenshot', 'true', step=step, index=index, clobber=False)
+    chip.set('tool', tool, 'task', task, 'var', 'screenshot', 'Whether to generate a screenshot of layout', field='help')
```
gadfort, line 51: Maybe: true/false, Whether to generate a screenshot of layout

gadfort (changes requested):
Looks great, could you merge in the main branch so all the CIs can be run?

nmoroze:
Note: I undid the version bump on KLayout, since it seems like the change I highlighted affects the KL Python distribution only. We should be able to work around the issue on headless systems by setting `QT_QPA_PLATFORM=offscreen` instead.

siliconcompiler/core.py
```diff
@@ -2771,9 +2773,104 @@ def _find_leaves(self, steplist):
...
+                fmax = f'{(fmax / 1e3):.2f} KHz'
+            elif fmax < 1e9:
+                fmax = f'{(fmax / 1e6):.2f} MHz'
+            else:
+                fmax = f'{(fmax / 1e9):.2f} GHz'
+            metrics['Fmax'] = fmax
+
+        # Generate design
+
+        WIDTH = 512
```
WRansohoff, line 2809: Could we go with a higher-resolution? The 4K screenshots I looked at were only be in the range of 4-5MB before gzipping, but 1024 or 2048 could also provide better detail on large designs.

gadfort:
Looks awesome:
asic_flat

WRansohoff (approved):
This looks great! The only comment I have is that a higher resolution might be nice for the summary image.

siliconcompiler/core.py
nmoroze: yes

gadfort: I'll leave the trigger point up to you, in the past I've usually gone a little past the SI breakpoint so 10e6 um2 -> 10.00mm^2, instead of 1e6, but I'm not wed to either

siliconcompiler/core.py
nmoroze: Fair point, scaled it up to 1024.

siliconcompiler/core.py
```diff
@@ -2791,7 +2791,11 @@ def _generate_summary_image(self, input_path, output_path):
         # "final" metrics regardless of flow would be handy
         totalarea = self.get('metric', 'totalarea', step='export', index='1')
         if totalarea:
-            metrics['Area'] = f'{totalarea:.2f} um^2'
+            # SI-ify
+            if totalarea < 1e3:
+                metrics['Area'] = f'{totalarea:.2f} um^2'
+            else:
+                metrics['Area'] = f'{totalarea / 1e3:.2f} mm^2'
```
gadfort, line 2798: this is an area, so it should be 1e6 to scale to mm^2

nmoroze: oh yeah, math

that explains why GCD was coming back so large

siliconcompiler/core.py
```diff
@@ -2792,10 +2792,10 @@ def _generate_summary_image(self, input_path, output_path):
         totalarea = self.get('metric', 'totalarea', step='export', index='1')
         if totalarea:
             # SI-ify
-            if totalarea < 1e4:
+            if totalarea < 1e7:
                 metrics['Area'] = f'{totalarea:.2f} um^2'
             else:
-                metrics['Area'] = f'{totalarea / 1e4:.2f} mm^2'
+                metrics['Area'] = f'{totalarea / 1e7:.2f} mm^2'
```
gadfort, line 2798: should be 1e6

gadfort (approved):
LGTM. Wait to merge until CIs are all passing

nmoroze:
Full CI run queued: https://github.com/siliconcompiler/siliconcompiler/actions/runs/4359738242

aolofsson:
White background or black background?

WRansohoff:
Personally, I like the black background - it looks clean and professional, and the grid lines aren't very noticeable.

gadfort:
> Personally, I like the black background - it looks clean and professional, and the grid lines aren't very noticeable.

I concur

aolofsson:
Black background makes integration harder, unless everyone is doing dark
mode now?

WRansohoff:
Good point - for comparison, here's the heartbeat example with a white background:

image

PR #1372, opened 2023-03-07
Add a bit more exposition to the client-side messages

WRansohoff:
Like we discussed offline, it would be nice to have a final message pointing to where the results are located after a remote run finishes, in case the build script does not include a call to `chip.summary()`.

If the full build results are being returned, it can take a bit of time to retrieve and unzip the results, so I also added a quick mention of that call. The delay probably won't be noticeable when only the core reports are returned, though.


gadfort (approved):
LGTM

aulihant (approved):
LGTM

PR #1376, opened 2023-03-08
Update public-server warning message based on offline feedback

WRansohoff:
Modified slightly from the original version, and based on feedback from Ali, I added a line noting that the full GDS-II results are not currently being returned.


gadfort (approved):
LGTM

aulihant (changes requested):
Minor reword. Otherwise good.

aulihant (approved):
LGTM

PR #1379, opened 2023-03-09
allow overwriting the HTML report, while preventing repeat generating the result png

gadfort:
Fixes:
- server fails to generate HTML report correctly because its already generated in an earlier step.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3071,12 +3071,12 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
             # TODO: we should generalize where we look for the png (perhaps [output, layout, png])
             layout_img = self.find_result('png', step='export', index='0')
 
-            if generate_image and layout_img and not os.path.isfile(results_img):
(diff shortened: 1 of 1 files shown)


WRansohoff:
I'm not sure about this one - it adds a dependency on another file from the build.

Although, if we're okay with re-generating the HTML, it also seems like it would be okay to re-generate the summary PNG. Either way, it's a bit of duplicate work that doesn't need to be performed.

nmoroze:
What's the context for this one? Not sure I understand the problem this is solving.

gadfort:
@nmoroze @WRansohoff updated PR to come...

gadfort:
@WRansohoff @nmoroze this restores the functionality that was there before. The only thing a user may not be able to do is regenerate the GDS->PNG step which is not required with the changes to klayout/export

nmoroze (approved):
Looks good, could you just update the docstring before merging? Github won't let me comment on it, but the `generate_html` and `generate_report` args say the corresponding item won't be regenerated if it already exists.

PR #1380, opened 2023-03-10
Improve error messaging when results cannot be fetched.

WRansohoff:
This updates the client to print relevant error messaging when the server doesn't return any results. I tested it by asking for the wrong job ID, but the behavior should be the same when the server encounters expected failures such as refusing to return over-sized results:

```
| INFO    | job0  | import     | 0  | Remote job run completed! Fetching results...
| WARNING | job0  | import     | 0  | Error fetching results: Job not found.
  File "sc_env/bin/sc", line 33, in <module>
    sys.exit(load_entry_point('siliconcompiler', 'console_scripts', 'sc')())
  File "siliconcompiler/apps/sc.py", line 77, in main
    chip.run()
  File "siliconcompiler/core.py", line 4293, in run
    fetch_results(self)
  File "siliconcompiler/client.py", line 341, in fetch_results
    chip.error("Sorry, something went wrong and your job results could not be retrieved.", fatal=True)
  File "siliconcompiler/core.py", line 5005, in error
    raise SiliconCompilerError(msg) from None
siliconcompiler.core.SiliconCompilerError: Sorry, something went wrong and your job results could not be retrieved.
```

This should be better than the "gzip failed, archive does not exist" errors that we were previously seeing in this case.

### DIFF
--- a/siliconcompiler/client.py
+++ b/siliconcompiler/client.py
@@ -282,7 +282,7 @@ def fetch_results_request(chip):
 
        Returns:
        * 0 if no error was encountered.
-       * 1 if the results could not be retrieved.
+       * [response code] if the results could not be retrieved.
     '''
 
     # Set the request URL.
@@ -317,13 +317,14 @@ def fetch_results_request(chip):
                 shutil.copyfileobj(resp.raw, zipf)
                 return 0
             else:
-                msg = '.'
+                msg_json = {}
                 try: # (An unexpected server error may not return JSON with a message)
-                    msg = f': {resp.json()["message"]}'
-                except:
-                    pass
+                    msg_json = resp.json()
+                    msg = f': {msg_json["message"]}'
+                except requests.exceptions.JSONDecodeError:
+                    msg = '.'
                 chip.logger.warning(f'Could not fetch results{msg}')
-                return 1
+                return resp.status_code
 
 ###################################
 def fetch_results(chip):
@@ -331,7 +332,12 @@ def fetch_results(chip):
     '''
 
     # Fetch the remote archive after the export stage.
-    results_failed = fetch_results_request(chip)
+    results_code = fetch_results_request(chip)
+
+    # Note: the server should eventually delete the results as they age out (~8h), but this will
+    # give us a brief period to look at failed results.
+    if results_code:
+        chip.error(f"Sorry, something went wrong and your job results could not be retrieved. (Response code: {results_code})", fatal=True)
 
     # Call 'delete_job' to remove the run from the server.
     delete_job(chip)
@@ -340,9 +346,6 @@ def fetch_results(chip):
     top_design = chip.get('design')
     job_hash = chip.status['jobhash']
     local_dir = chip.get('option', 'builddir')
-
-    if results_failed:
-        chip.error("Sorry, something went wrong and your job results could not be retrieved.", fatal=True)
 
     # Unauthenticated jobs get a gzip archive, authenticated jobs get nested archives.
     # So we need to extract and delete those.
--- a/siliconcompiler/client.py
+++ b/siliconcompiler/client.py
@@ -320,7 +320,7 @@ def fetch_results_request(chip):
                 msg_json = {}
                 try: # (An unexpected server error may not return JSON with a message)
                     msg_json = resp.json()
-                    msg = f': {msg_json["message"]}'
+                    msg = f': {msg_json["message"]}' if 'message' in msg_json else '.'
                 except requests.exceptions.JSONDecodeError:
                     msg = '.'
                 chip.logger.warning(f'Could not fetch results{msg}')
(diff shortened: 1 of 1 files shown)


siliconcompiler/client.py
```diff
@@ -309,17 +309,25 @@ def fetch_results_request(chip):
             elif resp.status_code == 303:
                 redirect_url = resp.headers['Location']
                 can_redirect = True
-            else:
+            elif resp.status_code == 200:
                 shutil.copyfileobj(resp.raw, zipf)
-                return
+                return 0
```
gadfort, line 314: Invert the logic to be True on success and false on Failure

Use True/False instead of 0/1

siliconcompiler/client.py
```diff
@@ -309,17 +309,25 @@ def fetch_results_request(chip):
...
+            elif int(resp.status_code / 100) in [4, 5]:
+                msg = '.'
+                try: # (An unexpected server error may not return JSON with a message)
+                    msg = f': {resp.json()["message"]}'
+                except:
+                    pass
+                chip.logger.warning(f'Error fetching results{msg}')
+                return 1
```
gadfort, line 322: Update

siliconcompiler/client.py
```diff
@@ -309,17 +309,25 @@ def fetch_results_request(chip):
...
-            else:
+            elif resp.status_code == 200:
                 shutil.copyfileobj(resp.raw, zipf)
-                return
+                return 0
+            elif int(resp.status_code / 100) in [4, 5]:
+                msg = '.'
+                try: # (An unexpected server error may not return JSON with a message)
+                    msg = f': {resp.json()["message"]}'
+                except:
```
gadfort, line 319: except what? enumerate and generate appropriate messages

siliconcompiler/client.py
```diff
@@ -329,7 +337,10 @@ def fetch_results(chip):
     job_hash = chip.status['jobhash']
     local_dir = chip.get('option', 'builddir')
 
-    # Authenticated jobs get a zip file full of other zip files.
+    if results_failed:
+        chip.error("Sorry, something went wrong and your job results could not be retrieved.", fatal=True)
```
gadfort, line 341: I think it's fine to leave out the Sorry.
On failure should we avoid deleting the job?
Is there an alternative method they could retrieve the results?

siliconcompiler/client.py
```diff
@@ -329,7 +337,10 @@ def fetch_results(chip):
...
+
+    # Unauthenticated jobs get a gzip archive, authenticated jobs get nested archives.
```
gadfort, line 343: Why different?

siliconcompiler/client.py
```diff
@@ -309,17 +309,25 @@ def fetch_results_request(chip):
             elif resp.status_code == 303:
                 redirect_url = resp.headers['Location']
                 can_redirect = True
-            else:
+            elif resp.status_code == 200:
                 shutil.copyfileobj(resp.raw, zipf)
-                return
+                return 0
+            elif int(resp.status_code / 100) in [4, 5]:
```
gadfort, line 315: can't you just do `400 <= resp.status_code and resp.status_code < 600` that seems much easier to read

gadfort (changes requested):
Don't forget to update the function docstrings to indicate the return values and what they mean.

siliconcompiler/client.py
```diff
@@ -309,17 +309,25 @@ def fetch_results_request(chip):
...
+                msg = '.'
+                try: # (An unexpected server error may not return JSON with a message)
+                    msg = f': {resp.json()["message"]}'
+                except:
+                    pass
+                chip.logger.warning(f'Error fetching results{msg}')
```
gadfort, line 321: Make this an error since it's an error

siliconcompiler/client.py
WRansohoff: The comment explains: the exception may occur if the response is not in a JSON format. It is expected to be, but the server could also encounter an unexpected error and return a 500 response without content.

siliconcompiler/client.py
WRansohoff: I guess, or maybe just `else:`

siliconcompiler/client.py
WRansohoff: Update what?

siliconcompiler/client.py
WRansohoff: 0/1 are more common error code responses. 0 = no error, nonzero = error.

siliconcompiler/client.py
gadfort: True False

siliconcompiler/client.py
gadfort: True/False

siliconcompiler/client.py
WRansohoff: Replaced the raw `except:` with a specific exception.

gadfort (approved):

PR #1381, opened 2023-03-10
Add clock to picorv32 example

aulihant:
It's in for picorv32_top but not picorv32.


gadfort (approved):

WRansohoff (approved):
Thanks for catching this!

PR #1382, opened 2023-03-10
Move PNG report generation to the end of `.run()`

WRansohoff:
Since KLayout is now generating a design screenshot in its `export` task, the final report summaries no longer have any EDA tool dependencies and can be run locally.

Moving the image generation to the end of `.run()` will ensure that A) the image exists even if `.summary()` is not called, and B) the image is only re-generated when something is likely to have changed (after a new run/resume/etc).

HTML report generation is untouched for now, per offline discussion.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -2873,8 +2873,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         design_summary.save(output_path)
 
     ###########################################################################
-    def summary(self, steplist=None, show_all_indices=False,
-                generate_image=True, generate_html=True):
+    def summary(self, steplist=None, show_all_indices=False, generate_html=True):
         '''
         Prints a summary of the compilation manifest.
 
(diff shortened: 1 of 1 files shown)


nmoroze (approved):
Change seems fine, although I'm not sure I understand what problem this solves. It seems alright to generate the image at the end of `run()` if that's what we'd like, but it seemed fine to do it in `summary()` too. The extra work even if someone were to call `summary()` more than once seems negligible, as far as I recall the image generation is pretty fast. 

(my apologies if I'm missing context from offline discussion)

gadfort:
@nmoroze the call to summary is not always there and this ensures the file is always generated regardless and only pops up on calling summary(). It's more to make sure the file is there, the work to make the PNG is pretty light

gadfort (approved):

gadfort:
@WRansohoff lets wait to merge until we can check with @aolofsson

WRansohoff:
It will probably be easier to discuss the reasoning offline; there are a few intertwined issues.

In a nutshell, the fix in #1379 introduced a new failure, and revealed that the summary-based report generation logic is probably brittle enough to merit some reconsideration.

gadfort:
@WRansohoff closing this per the offline discussion and will update tutorials instead

PR #1386, opened 2023-03-12
Fix klayout export to handle empty cells and DBUs mismatch

gadfort:
Fixes: https://github.com/siliconcompiler/zerosoc/issues/22


nmoroze (approved):
LGTM, thanks!

PR #1388, opened 2023-03-14
vivado: update for SC 0.10

nmoroze:
Changes verified locally by running `examples/blinky/run_vivado.sh`.


siliconcompiler/tools/vivado/__init__.py
```diff
@@ -0,0 +1 @@
+from .vivado import *
```
gadfort, line 1: I see the reason for this, but I think you should do explicit imports so it's clear what exactly it is you want

siliconcompiler/tools/vivado/route.py
```diff
@@ -0,0 +1,17 @@
+from siliconcompiler.tools import vivado
+
+def setup(chip):
+    '''Performs routing.'''
+    tool = vivado.tool
```
gadfort, line 5: couldn't you just import this as `from sc.tools.vivado import tool`

gadfort (changes requested):
Why not split the compile.tcl into the different functions, otherwise you still have this `if task == ` which shouldn't be needed

siliconcompiler/tools/vivado/route.py
nmoroze: will do, although it's kind of a wash line count wise (I don't like the implicit import of post_process, so I still want to keep the first namespaced import)

nmoroze:
Thanks for the feedback, I refactored `compile.tcl` to use OpenROAD's pattern (which still uses a common entry point and switches on task name). Not sure if you have an opinion on that vs having each task have its own script entry point (which then sources a common pre/post script).

gadfort:
I think it would be cleaner to have each task associate their own script and then source a common startup script.

nmoroze:
Sounds good, done. It does end up adding some boilerplate to grab the refdir of each script.

gadfort (approved):
Just revert the script breakout

PR #1389, opened 2023-03-14
Restructure Documentation for Readability

aulihant:
I think there might be several phases to documentation updates. For this first phase, I'm hoping to just settle on a reordering of the pages for overall readability and to make clear what user vs developer is and provide information on how they should approach the documentation, without being too focused on content of specific pages. 

<html>
<body>

Change | Reason
-- | --
(1) Move intro into index.html | Helps make the default entrance page more meaningful
(2) Shorten User Guide | - Pulling out all pages which require developer mindset<br>- Want to keep clean information for users (who don’t plan to contribute)
(3) Add new FLOW BASICS section for flow | Group together "data model," "programming model" and "execution model" since these three main sections are the next step to understanding the flow that the user ran
<strike>(4) Reverse order of FLOW BASICS files</strike> | <strike>Reverse order Execution → Programming → Data because…<br>&nbsp;- While referenced in others, execution describes for the first time... <br>&nbsp;&nbsp;&nbsp;- flowgraph<br>&nbsp;&nbsp;&nbsp;- step/index<br>&nbsp;&nbsp;&nbsp;- node/edge<br>&nbsp;&nbsp;&nbsp;- get/set<br>&nbsp;- Pictures are always easier to absorb and the flowgraphs will help<br>&nbsp;- Data model is the most detailed one and concerned this will lose people early<br>&nbsp;It helps to see big picture then get into details</strike>
(5) Add a Developer Manual section | - Move everything else in User Guide here<br>- Move contributing modules here since it’s a bit different than other tutorials
(6) Group frontend tutorials together | Cleans up tutorial section for easier reading
(7) Adjusted some links in tutorials  | Now that merged together in one page these were added to address conflicting label warnings

</body>
</html>

Look through documentation created for this PR here: https://siliconcompiler--1389.org.readthedocs.build/en/1389/

**Note**: If these changes are approved, it would make sense for some files to move, but I didn't want to do that right away so you could see the file diffs.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -25,9 +25,9 @@ Welcome to SiliconCompiler's Documentation!
    :caption: Flow Basics
    :numbered:
 
-   user_guide/execution_model
-   user_guide/programming_model
    user_guide/data_model
+   user_guide/programming_model
+   user_guide/execution_model
 
 .. toctree::
    :maxdepth: 3
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -15,10 +15,24 @@ Welcome to SiliconCompiler's documentation!
    user_guide/installation
    user_guide/quickstart
    user_guide/glossary
+   user_guide/remote_processing
+   user_guide/faq
+   user_guide/licenses
+
+.. toctree::
+   :maxdepth: 3
+   :caption: Flow Basics
+   :numbered:
+
    user_guide/data_model
    user_guide/programming_model
    user_guide/execution_model
-   user_guide/remote_processing
+
+.. toctree::
+   :maxdepth: 3
+   :caption: Reference Manual
+   :numbered:
+      
    user_guide/tools
    user_guide/pdks
    user_guide/flows
@@ -27,8 +41,6 @@ Welcome to SiliconCompiler's documentation!
    user_guide/records
    user_guide/libraries
    user_guide/packages
-   user_guide/faq
-   user_guide/licenses
 
 .. toctree::
    :maxdepth: 3
@@ -36,17 +48,17 @@ Welcome to SiliconCompiler's documentation!
    :numbered:
 
    tutorials/contribution
+   tutorials/picorv32_ram
    tutorials/parallel
    tutorials/migen
    tutorials/chisel
    tutorials/bambu
    tutorials/bluespec
-   tutorials/picorv32_ram
    tutorials/slurmsetup
 
 .. toctree::
    :maxdepth: 3
-   :caption: Reference Manual
+   :caption: Developer Manual
    :numbered:
 
    reference_manual/schema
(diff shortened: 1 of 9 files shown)


nmoroze (commented):
Looks great! I really like moving the intro contents to the first page, recombining the frontend tutorials, and the developer manual split. Some feedback:

1: What do you think about lifting the "Getting Started" section of the front page above "Motivation"? (and perhaps adding "Read on for more context on the philosophy of the project" or something along those lines). I like that Getting Started has a concrete call-to-action, whereas Motivation feels a bit philosophical and doesn't teach me much about the project.

2: I know we discussed this previously, but reading through the sections in the Flow Basics section again, I'm less sure about this change. A slightly more radical proposal: I think we should add back an additional page that looks like the old Quickstart ([reference](https://docs.siliconcompiler.com/en/noah-examples-cleanup/user_guide/quickstart.html)), and remove Programming Model entirely. I think using `asic_demo` is nice for a smoke test, but teaches you a lot less about how SC works than showing an actual script to build a design. Also, almost all of the API calls that are taught in Programming Model were shown in the old Quickstart, so I think the reincarnation of old quickstart could just be written to incorporate the line-by-line context provided by Programming Model, now that it doesn't have to be quite as snappy (since new Quickstart handles the smoke-test aspect). I think this would also be an improvement since it would be much more concrete than Programing Model. 

I also think this would be important to include before Execution Model, since that page makes reference to the heartbeat script introduced in the old Quickstart, with some variations.

3: I'm less sure where Data model should go, since it's pretty in-the-weeds. I think it could also use some rework, since the code snippets, list of fields, and summary of schema categories seems useful to show the user early-ish, but I think the actual discussion of what the underlying JSON looks like is more of an implementation detail that should probably be buried further down in the docs. I wonder if some of it should be combined with the Schema section of the Reference Manual, since they seem to duplicate the explanation of the fields.

aulihant:
Thanks @nmoroze for the review. I think your suggestion on moving "Getting Started" up makes sense, so I incorporated it.

As for Flow Basics, I moved the order back to the original order, but I do agree it needs some rework. I'll think about how to adjust it in a way that makes sense. But does the motivation of Flow Basics still make sense as a separate section? There are links at the end of Quickstart which point you to the different flow basic sections so the idea is that a user has already run through Quickstart, and they're ready to get into how to set up a run with their design in more detail.

gadfort (commented):
LGTM, I assume we are addressing the content in a series of future PRs?

aulihant:
> LGTM, I assume we are addressing the content in a series of future PRs?

Yes, future PRs, since I think the current changes is a standalone step to improving readability on the docs.

gadfort (approved):

jldec (approved):
Thanks @aulihant 
The index/welcome page is a very nice improvement, and I like the more approachable sections in the table of contents / navigation. 
👍

PR #1390, opened 2023-03-15
Implement pre/post-processing for tasks run through a scheduler

WRansohoff:
This change implements a method of running pre/post-processing methods in the Slurm job scripts. Depending on how a cluster is set up, it may be necessary to move data around, modify the runtime environment, etc. before or after a task is launched. We probably don't want those shell commands to live in the Schema for a whole bunch of reasons, but the ephemeral `chip.status` dictionary is a good fit. Cluster administrators can put relevant commands into a `Flow` object or a target, but those commands won't persist on other machines where the jobs might get copied and re-run.

I also tried to move most of the other scheduler configurations out of `chip.status` and into the Schema, and added Slurm support for the `'defer'` parameter. I think we might want to add one more `'option', 'scheduler'` parameter for constraints/tags, but that can come in a future PR.


siliconcompiler/scheduler.py
```diff
@@ -19,10 +19,7 @@ def _deferstep(chip, step, index, status):
     scheduler_type = chip.get('option', 'scheduler', 'name', step=step, index=index)
 
     # Determine which cluster parititon to use. (Default value can be overridden on per-step basis)
-    if f'{step}_slurm_partition' in chip.status:
-        partition = chip.status[f'{step}_slurm_partition']
-    else:
-        partition = chip.status['slurm_partition']
+    partition = chip.get('option', 'scheduler', 'queue', step=step, index=index)
```
gadfort, line 22: documentation needs to be updated to reflect this is modified

siliconcompiler/scheduler.py
```diff
@@ -62,17 +60,34 @@ def _deferstep(chip, step, index, status):
         if 'slurm_account' in chip.status:
             username = chip.status['slurm_account']
             schedule_cmd.extend(['--acount', username])
+        # Only delay the starting time if the 'defer' Schema option is specified.
+        defer_time = self.get('option', 'scheduler', 'defer')
```
gadfort, line 64: step/index

siliconcompiler/scheduler.py
```diff
@@ -62,17 +60,34 @@ def _deferstep(chip, step, index, status):
...
+            schedule_cmd.extend(['--begin', defer_time])
     elif scheduler_type == 'lsf':
         # TODO: LSF support is untested and currently unsupported.
         schedule_cmd = ['lsrun']
 
+    # User-defined pre/post-processing steps for the Slurm script.
+    # Unlike most other scheduler parameters, we should keep these pre/post-process blocks in the
+    # ephemeral status dictionary, so that they cannot persist across remote connections.
+    # The best place to set these values would be a custom target created by the cluster administrator.
+    script_prefix, script_suffix = '\n', '\n'
```
WRansohoff, line 75: Per offline discussion, replace this with "if scheduler script does not already exist" logic.

tests/flows/test_slurm_local.py
```diff
@@ -2,20 +2,68 @@
...
+    gcd_chip.set('option', 'scheduler', 'slurm')
 
     # Run the chip's build process synchronously.
     gcd_chip.run()
 
     # Verify that GDS file was generated.
     assert os.path.isfile('build/gcd/job0/export/0/outputs/gcd.gds')
 
+@pytest.mark.eda
+def test_slurm_local_py_prefix(gcd_chip):
```
WRansohoff, line 21: Make these new tests unit tests on `Chip._deferstep`, not functional end-to-end tests.

WRansohoff:
In addition to the requested changes, I'm marking this as a draft. We're going to hold off on merging further Schema-changing PRs for a week or so, to let the new release settle in.

siliconcompiler/scheduler.py
WRansohoff: The Slurm documentation is still pretty barebones, and it currently only references the `('option', 'scheduler', 'name', 'slurm')` keypath.

gadfort (approved):

PR #1391, opened 2023-03-15
Update api schemas

WRansohoff:
I finally had time to review the API schemas that are reflected in the documentation, and the good news is that they appear to be up to date.

This change just updates the `"key"` field descriptions, because we moved towards using traditional username/password authentication, rather than forcing end-to-end encryption where users can manage and rotate the public keys which are associated with their accounts.


aulihant (approved):
LGTM!

PR #1394, opened 2023-03-16
update install scripts for ubuntu20.04

gadfort:


setup/install-bluespec.sh
```diff
@@ -14,24 +14,27 @@ sudo apt-get install -y ghc libghc-regex-compat-dev libghc-syb-dev \
...
-git clone $(python3 ${src_path}/_tools.py --tool bsc --field git-url) bsc
-cd bsc
-git checkout $(python3 ${src_path}/_tools.py --tool bsc --field git-commit)
+git clone $(python3 ${src_path}/_tools.py --tool bluespec --field git-url) bluespec
+cd bluespec
+git checkout $(python3 ${src_path}/_tools.py --tool bluespec --field git-commit)
 git submodule update --init --recursive
 
-make install-src
+make -j install-src
```
nmoroze, line 25: been a while since I looked at this, will this install to $PREFIX by default if it's defined?

setup/install-openroad.sh
```diff
@@ -13,8 +13,16 @@ git checkout $(python3 ${src_path}/_tools.py --tool openroad --field git-commit)
 git submodule update --init --recursive
 
 sudo ./etc/DependencyInstaller.sh
+if [ ! -z ${SC_BUILD} ]; then
+    cp ./etc/DependencyInstaller.sh ${SC_BUILD}/
+fi
```
nmoroze, line 18: what is purpose of this? I assume Docker related?

setup/install-bluespec.sh
gadfort: Yes it does, but PREFIX is generally not defined

setup/install-openroad.sh
gadfort: yeah, I need the DependencyInstaller.sh in a particular location and this ensures it's not in the /deps/ which gets removed immediately after the build.

nmoroze (approved):

setup/install-openroad.sh
gadfort: I could remove it for now, since the Docker stuff is still in development

nmoroze: I think it's fine, the main thing that stuck out to me was that you're moving it after you call it?

gadfort: trying to keep the docker image small-ish, so after calling `install-openroad.sh && rm -rf /deps` which causes the file to get deleted before I can copy it out

PR #1401, opened 2023-03-21
update install and quickstart

aulihant:


aulihant:
Move changes into docs overhaul update

PR #1403, opened 2023-03-23
add recording of child process memory usage

### LINKED ISSUE #1152: Memory usage tracking innacurate
Our memory tracking seems to undercount in some cases. My theory is that we need to add in memory used by child processes of a tool, since this issue showed up for a tool that spawns multiple children.

gadfort:
Closes #1152


gadfort:
Thanks for checking it out. I think the polling will result in some differences.

PR #1411, opened 2023-03-27
Build Wheels for Python 3.11 [SC-198]

nmoroze:
Changes:
- Update cibuildwheel action to latest verison, which builds for 3.11 by default
- Fix breaking changes:
  - Skip build of "musllinux" wheels, which got added as a default and was causing problems. This seems to be a requirement for niche distros (like Alpine), and we can evaluate support later if there's interest.
  - Manually delete "_skbuild/" directory in Windows: cibuildwheel changed to a new version of Pip which performs in-tree builds, and for some reason the built-in machinery for clearing out build artifacts in between each Python version was failing out on Windows. Adding the custom command seems to fix this.
- Add Python 3.11 to package_offline job

Verification run: https://github.com/siliconcompiler/siliconcompiler/actions/runs/4534605858

I originally added Python 3.11 to the quick/Daily CI tests, but removed that in favor of the Docker PR.

### DIFF
--- a/.github/workflows/daily_tests.yml
+++ b/.github/workflows/daily_tests.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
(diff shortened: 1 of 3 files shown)


gadfort (approved):
LGTM, thanks.

PR #1412, opened 2023-03-27
Add try/except to external calls

gadfort:
Changes:
- Adds a try/except block around calls to external code to help enable better error messaging.
- Check return code for --version checks to ensure failures are caught.


nmoroze (approved):
Looks good, thanks!

PR #1413, opened 2023-03-27
Build Linux + ARM64 wheels with caching for all platforms

nmoroze:
This PR adds support for Linux+ARM64 [SC-214/SC-219] wheels and addresses a few other issues in the process:

- Adds Surelog build caching for Linux builds
- Fix Windows cache blank key [SC-200]
- Upload created wheels even when job fails (although wheels that fail tests still don't get uploaded, because it seems like cibuildwheel doesn't copy them to the final wheelhouse/ location until they pass)
- Clean up extraneous Surelog build artifacts
  - The Surelog build creates a bunch of extra artifacts that aren't necessary to run. The cache was previously acting like a filter since it's only applied to a few paths, but this meant that Linux wheels used to be consistently bloated, and macOS/Windows wheels would occasionally be large for the first build of a new Surelog version. Now the wheels should be uniform regardless of if there was a cache hit or not

Parallelization + caching does make this more palatable than the initial attempt, but the emulated ARM build is still so slow that it's a huge performance hit -the build time even with a Surelog cache hit is now ~2 hours, up from ~45 minutes.

I had to bump timeouts on a few long tests since they were failing for ARM. There's also no pre-built KL ARM package, so I skipped those tests on that platform for now. 

Full build (cache miss): https://github.com/siliconcompiler/siliconcompiler/actions/runs/4514876214
Build w/ cached Surelog: https://github.com/siliconcompiler/siliconcompiler/actions/runs/4534346464. 

Re-running to verify rebase on 3.11 changes: https://github.com/siliconcompiler/siliconcompiler/actions/runs/4536130720

[SC-200]: https://zeroasic.atlassian.net/browse/SC-200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/core/test_create_cmdline.py
+++ b/tests/core/test_create_cmdline.py
@@ -1,3 +1,4 @@
+import logging
 import re
 import pytest
 
@@ -5,6 +6,7 @@ import siliconcompiler
 
 def do_cli_test(args, monkeypatch, switchlist=None, input_map=None):
     chip = siliconcompiler.Chip('test')
+    print(chip.get('constraint', 'timing', 'worst', 'libcorner', step='syn', index='0'))
     monkeypatch.setattr('sys.argv', args)
     chip.create_cmdline('sc', switchlist=switchlist, input_map=input_map)
     return chip
@@ -100,7 +102,6 @@ def _cast(val, sctype):
         # everything else (str, file, dir) is treated like a string
         return val.strip('"')
 
-@pytest.mark.timeout(500)
 def test_cli_examples(monkeypatch):
     # Need to mock this function, since our cfg CLI example will try to call it
     # on a fake manifest.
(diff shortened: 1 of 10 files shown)


nmoroze:
Will wait to merge until the re-run after rebasing on the 3.11 wheel changes is done, but otherwise this should be good to go.

tests/core/test_create_cmdline.py
```diff
@@ -100,6 +100,7 @@ def _cast(val, sctype):
         # everything else (str, file, dir) is treated like a string
         return val.strip('"')
 
+@pytest.mark.timeout(500)
```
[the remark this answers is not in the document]
nmoroze: This was failing the ARM build at the default timeout

tests/core/test_create_cmdline.py
```diff
@@ -1,10 +1,12 @@
+import logging
 import re
 import pytest
 
 import siliconcompiler
 
 def do_cli_test(args, monkeypatch, switchlist=None, input_map=None):
     chip = siliconcompiler.Chip('test')
+    print(chip.get('constraint', 'timing', 'worst', 'libcorner', step='syn', index='0'))
```
gadfort, line 9: remove this debug line

nmoroze: Nice catch, thanks

gadfort (approved):
LGTM

PR #1417, opened 2023-03-29
Remove import/export requirements

gadfort:
Changes:
- Removes the assumption that the starting and ending tasks are called import or export
- Renames previous import tasks as convert or parse
- switch from using hardcoded task names to `_get_task`

Adds:
- entry / exit node helper functions to find input and output nodes in the flowgraph

Note: cannot merge until it's been verified on the server


siliconcompiler/core.py
```diff
@@ -1253,21 +1253,15 @@ def _find_files(self, *keypath, missing_ok=False, job=None, step=None, index=Non
...
-                        # file may not have been gathered in imports yet)
-                        result.append(abspath)
-                        found = True
-                        break
-                if found:
+                abspath = os.path.join(self._getcollectdir(jobname=job), name)
+                if os.path.isfile(abspath):
+                    # if copy is True and file is found in import inputs,
+                    # continue. Otherwise, fall through to _find_sc_file (the
+                    # file may not have been gathered in imports yet)
```
nmoroze, line 1263: maybe update comment to say "collected inputs" or something along those lines instead of import

siliconcompiler/core.py
```diff
@@ -4935,32 +4918,63 @@ def _is_builtin(self, tool, task):
...
+
+        nodes = []
+        for step in self.getkeys('flowgraph', flow):
+            for index in self.getkeys('flowgraph', flow, step):
+                if not self.get('flowgraph', flow, step, index, 'input'):
+                    nodes.append((step, index))
+        return nodes
+
+    #######################################
+    def _get_flowgraph_exit_nodes(self, flow=None):
```
nmoroze, line 4937: I think this function is basically the same as `Chip._find_leaves()`, could we get rid of that one/merge them?

siliconcompiler/core.py
```diff
@@ -4935,32 +4918,63 @@ def _is_builtin(self, tool, task):
...
+                    nodes.append((step, index))
+        return nodes
+
+    #######################################
+    def _getcollectdir(self, jobname=None):
+        '''
+        Get absolute path to collected files directory
+        '''
+
+        return os.path.join(self._getworkdir(jobname=jobname), 'sc_collected_files')
```
nmoroze, line 4962: Nice clean up to use this instead of `import/0/inputs`, do you think we want a check somewhere to prevent someone from naming an actual step "sc_collected_files" (as unlikely as that might be...)

nmoroze (changes requested):
Looks great, thanks! Glad you got the file collection location cleaned up as well.

siliconcompiler/core.py
gadfort: probably should add a check just in case

siliconcompiler/core.py
gadfort: merged into one function

siliconcompiler/flows/asicflow.py
```diff
@@ -64,7 +64,6 @@ def setup(chip, flowname='asicflow', syn_np=1, floorplan_np=1, physyn_np=1, plac
 
     #step -->(tool, task)
     tools = {
-        'import' : ['surelog', 'import'],
```
WRansohoff, line 67: Does the initial task get set up automatically from having `('option', 'frontend') == 'verilog'`?

siliconcompiler/core.py
```diff
@@ -2342,21 +2330,22 @@ def _collect(self, step, index):
 
         '''
 
-        indir = os.path.join(self._getworkdir(step=step, index=index), 'inputs')
+        if not directory:
+            directory = os.path.join(self._getcollectdir())
```
WRansohoff, line 2334: Is this preferable to accepting step/index in the `_collect` method, and:
```
if not directory:
    directory = os.path.join(self._getworkdir(step=step, index=index), 'inputs')
```
It looks like that might be the most common override for the `directory` kwarg, anyways.

siliconcompiler/tools/yosys/syn_fpga.py
```diff
@@ -43,17 +43,12 @@ def create_vpr_lib(chip):
     step = chip.get('arg','step')
     index = chip.get('arg','index')
 
-    src = f"{chip.scroot}/tools/yosys/vpr_yosyslib"
-    dst = f"{chip._getworkdir()}/{step}/{index}/inputs/vpr_yosyslib"
+    src = f"{os.path.dirname(__file__)}/vpr_yosyslib"
+    dst = f"{chip._getworkdir(step=step, index=index)}/inputs/vpr_yosyslib"
```
WRansohoff, line 47: `os.path.join(a, b)` instead of `f'{a}/{b}'`?

WRansohoff (approved):
Looks good to me! Just a few minor comments/questions.

siliconcompiler/flows/asicflow.py
gadfort: it does by the call to `flowtools = setup_frontend(chip)`

siliconcompiler/core.py
gadfort: step/index doesn't really mean anything for the `_collect()` function. Before it was just copying files into each import step (the same files). So I don't think it makes sense to provide it as a `_collect`, and I don't think there are  any instances of using the directory override.

aolofsson (approved):
This is a great change! The hard coded entry point was never a good solution given the broad applicability of the system. Very happy to see this!

siliconcompiler/core.py
WRansohoff: It is used in the chisel driver, but fair enough.

PR #1423, opened 2023-03-30
initial capability to update tools automatically

gadfort:
Adds:
- workflow to update tools nightly

Flow:
- checkout main branch
- use _tools.py to update a single tool
- if update is available, create PR for that tool update

Note: currently uses my token to run the flow


WRansohoff (commented):
If I'm understanding this correctly, it will add an action to run every day which:

* Checks the latest Git commit hash for each open-source tool's repository,
* Updates the commit hash in `_tools.json` if there is a newer one,
* Creates a PR to use that newer tool version?

Sounds like a good idea. I'm just curious if we'll have a mechanism for marking skip / hold-back commit hashes in case an update introduces breaking changes, and whether the CI tests for the 'new version' PRs will run with the new versions of the tools.

nmoroze (approved):
Looks good! Have you had a chance to test out the PR-opening Action somewhere?

gadfort:
> If I'm understanding this correctly, it will add an action to run every day which:
> 
> * Checks the latest Git commit hash for each open-source tool's repository,
> * Updates the commit hash in `_tools.json` if there is a newer one,
> * Creates a PR to use that newer tool version?
> 
> Sounds like a good idea. I'm just curious if we'll have a mechanism for marking skip / hold-back commit hashes in case an update introduces breaking changes, and whether the CI tests for the 'new version' PRs will run with the new versions of the tools.

You are correct about the operation, and no there is no mechanism for skipping (except we can ignore that PR until its passing or add the fixes needed to get that tool enabled again).

@WRansohoff yes, I had `on push` enabled yesterday, which fake updated openroad and yosys. We could slow it down from nightly to every few days if we wanted to just to keep the churn down a bit

gadfort:
> Looks good! Have you had a chance to test out the PR-opening Action somewhere?

https://github.com/siliconcompiler/siliconcompiler/pull/1418 (auto opened PR)

WRansohoff (approved):
I think a manual override to peg a specific version for a specific tool could be useful, but that can come at a later time. Worst case scenario, we could always temporarily disable the action and update `_tools.json` by hand.

gadfort:
> I think a manual override to peg a specific version for a specific tool could be useful, but that can come at a later time. Worst case scenario, we could always temporarily disable the action and update `_tools.json` by hand.

Upon reflection, I agree, it would be pretty each to add that to _tools.json

PR #1424, opened 2023-03-30
change logger to say remote while processing remote run and restore to regular on return

gadfort:


WRansohoff (approved):
Good idea, it's probably better than logging under the initial task name until the job finishes.

PR #1426, opened 2023-03-31
Use python modules for tasks

gadfort:
This PR needs to be merged after #1417 and will require the removal of importlib for "import" handling.

Changes:
- `.node()` takes in `flow`, `step`, `task module`, which it uses to set tool / task / taskmodule
- Adds a `taskmodule` to the flowgraph schema to track this module for use throughout the local and remote execution
- Adds check_manifest to ensure modules are available and loadable
- removes `find_function` and `read_file` functions as they no longer serve a purpose

Note: This will need to tested against the server before merging
Note: `dvflow` is broken due to missing tasks in verilator and will need to be fixed in a later PR

Need for additional schema parameter:
In order to support the use case of 
```
chip = siliconcompiler('chip')
chip.read_manifest('file')
chip.run()
```
The manifest needs to contain the module information in the schema, otherwise we would have to guess and reconstruct the modulename which we cannot guarantee to be correct from session to session. For example, how should we tell the difference between `siliconcompiler.tools.openroad.place` and `someoneelse.openroad.place` if all we have is `openroad` and `place`.

Without the schema parameter, we would need to require something like:
```
chip = siliconcompiler('chip')
chip.read_manifest('file')
chip.load_target('reload_target')
chip.run()
```
to ensure proper tracking of the module.


siliconcompiler/core.py
```diff
@@ -166,20 +167,59 @@ def top(self):
...
+    def _get_tool_module(self, step, index, flow=None, error=True):
+        if not flow:
+            flow = self.get('option', 'flow')
+
+        tool, _ = self._get_tool_task(step, index, flow=flow)
+
+        taskmodule = self.get('flowgraph', flow, step, index, 'taskmodule')
+        module_path = taskmodule.split('.')
+
+        tool_module = '.'.join(module_path[0:-1] + [tool])
```
nmoroze, line 195: I feel like tool module path should be tracked separately - that way you can e.g. make a custom OpenROAD task without having to put a boilerplate OpenROAD tool driver next to it.

siliconcompiler/core.py
```diff
@@ -3302,30 +3254,51 @@ def node(self, flow, step, tool, task, index=0):
...
+        if (isinstance(task, str)):
+           task_module = task
+        elif inspect.ismodule(task):
+            task_module = task.__name__
+        else:
+            self.error(f"{task} is not a string or module and cannot be used to setup a task.", fatal=True)
+
+        task_parts = task_module.split('.')
+        if len(task_parts) < 2:
+            self.error(f"{task} is not a a valid task, it must be associated with a tool '<tool>.<task>'.", fatal=True)
```
nmoroze, line 3291: typo: extra "a"

nmoroze (commented):
Looks good at first read through. The one thing I wasn't expecting is that tool modules are found by looking next to the task modules. Was there a reason not to specify a separate module for both tool and task? I'd like to support custom tasks for the prepackaged tools, and although you can write a boilerplate tool module that imports the methods from the packaged one, that feels a little more clunky.

aolofsson:
Can you dive into the motivation for the taskmodule addition to the schema. I know we went around on this a couple of times and I can't recall the pros and cons. For example, I thought the module name could be handled through the client/server API instead of in the schema?

I think this is a place where the PR description needs to be a bit more descriptive.

gadfort:
> Can you dive into the motivation for the taskmodule addition to the schema. I know we went around on this a couple of times and I can't recall the pros and cons. For example, I thought the module name could be handled through the client/server API instead of in the schema?
> 
> I think this is a place where the PR description needs to be a bit more descriptive.

@aolofsson I added a little more to the PR description. But ultimately, without using the schema value, even local runs would require additional steps to run if all we have is the manifest, which we decided we wanted to be able to support running from a manifest file only. We can also schedule some time to chat about this if it helps.

siliconcompiler/flows/asicflow.py
```diff
@@ -62,22 +72,21 @@ def setup(chip, flowname='asicflow', syn_np=1, floorplan_np=1, physyn_np=1, plac
...
-        'placemin' : ['builtin','minimum'],
-        'cts' : ['openroad','cts'],
-        'ctsmin' : ['builtin','minimum'],
-        'route' : ['openroad','route'],
-        'routemin' : ['builtin','minimum'],
-        'dfm' : ['openroad','dfm']
+    #step --> task
+    tasks = {
+        'syn' : syn_asic,
+        'synmin' : 'builtin.minimum',
```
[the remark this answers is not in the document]
gadfort: I agree. With a little refactor we could make them tasks like all the other tools. I had started that a while ago, but it didn't make sense in that context, but with this change I think I would be easy to resurrect.

aolofsson (approved):
Looks good!

siliconcompiler/core.py
```diff
@@ -166,20 +167,59 @@ def top(self):
...
+
+        tool, _ = self._get_tool_task(step, index, flow=flow)
+
+        taskmodule = self.get('flowgraph', flow, step, index, 'taskmodule')
+        module_path = taskmodule.split('.')
+
+        tool_module = '.'.join(module_path[0:-1] + [tool])
+
+        try:
+            return importlib.import_module(tool_module)
```
WRansohoff, line 198: I think this will fully import the module before returning it, which may involve evaluating code and could be slow in some cases. Ditto for this line in `_get_task_module`.

siliconcompiler/core.py
```diff
@@ -598,16 +565,37 @@ def load_target(self, name, **kwargs):
...
+
+        if len(modules) == 0:
+            self.error(f'Could not find target {name}', fatal=True)
+
+        # Check for setup in modules
+        load_function = None
+        for module in modules:
+            load_function = getattr(module, 'setup', None)
+            if load_function:
+                break
```
WRansohoff, line 585: Do you think it's worthwhile to treat `modules` as a list? It seems like there could only be more than one value if both potential module names are found and loaded (e.g. `mypackage.demo` and `siliconcompiler.targets.mypackage.demo`).

That seems unlikely enough that it might be worth storing the loaded module as a single value, and using `break` in the earlier loop which iterates over the module names. Plus, this loop wouldn't let multiple `setup` methods be called even if multiple modules were imported.

siliconcompiler/core.py
```diff
@@ -3839,7 +3811,7 @@ def _runtask(self, step, index, status):
         ##################
         # Run preprocess step for tool
         if not is_builtin:
-            func = self.find_function(tool, "pre_process", 'tools', task)
+            func = getattr(self._get_task_module(step, index, flow=flow), 'pre_process', None)
```
WRansohoff, line 3814: This is why I think it'll be great to move to using Python modules.

However, I think that this sort of logic might be inefficient, because it will fully re-import the module every time that the `_get_task_module` is called.

Could we cache loaded modules somewhere to mitigate that?

siliconcompiler/flows/dvflow.py
```diff
@@ -26,54 +29,58 @@ def setup(chip, np=1):
     The dvflow can be parametrized using a single 'np' parameter.
     Setting 'np' > 1 results in multiple independent verificaiton
     pipelines to be launched.
+
+    This flow is a WIP
     '''
 
+    # Disabled testgen, refsim, sim, and compare because verilator tasks are missing
```
WRansohoff, line 36: Maybe mark this as a TODO?

WRansohoff (approved):
Overall looks good - just a couple of concerns about the efficiency of fully re-loading modules every time that we we want to interact with them in `core.py`. Could be addressed in a future PR, but we should make sure that is tracked somewhere if you agree that we should use some sort of caching for programmatically-imported modules.

siliconcompiler/core.py
gadfort: we would only want to to call the first one, there would be no need to call multiple. The reason for the list is to get around importing the wrong module which doesn't have a setup function and needing the second found module.

PR #1436, opened 2023-04-03
check if item in list before removing in klayout export task

gadfort:


WRansohoff (approved):

nmoroze (approved):
Seems fine, this isn't masking a bug/setup error right?

gadfort:
> Seems fine, this isn't masking a bug/setup error right?

No, the issue here is that the second GDS contains files needed by the first GDS, but those are subinstances in the GDS and therefore not present in the def.

PR #1441, opened 2023-04-05
Provide logger to Schema object

nmoroze:
I think this is the simplest fix to this long-standing annoyance, although the logging stuff as a whole could probably use some more love. 

I took a look at this since it's on the path to fixing the problem we noticed with the KLayout scripts, where using the Schema object directly has different behavior than going through the Chip object. I'd like to pull the `merge_manifest` logic into Schema, which will help solve that problem (and that requires a logger).


nmoroze:
Closing in favor of a more all-encompassing refactor of logger at some point.

PR #1443, opened 2023-04-06
re-skip two long example tests and add longer timeout to test with mu…

gadfort:
…ltiple steps


nmoroze (approved):
Looks good, is there any situation you'd expect to add back the super long ones (shorten/speed up or run on a different cadence or something)?

gadfort:
@nmoroze When we redo the CI, I think these can come back but in a different form most likely

PR #1444, opened 2023-04-06
Make `Schema()` behavior match `Chip()` in KL scripts

nmoroze:
Since the `Schema` object instantiated in the KL scripts was being initialized directly from a manifest, rather than having it be merged into a set of default values, some of the behavior was subtly different from using a `Chip` object instead. This is confusing and error-prone, so this PR moves `read_manifest()` and other methods used to implement it into `Schema`, so that it should behave just like a chip object. 

`Chip` still provides a `read_manifest` pass-through in the same style of get/set/etc., so its public API has not been changed.

I verified manually that this fixes the test case Aulihan reported - it runs thru successfully even if the lyp file is removed.

This PR is based on #1441, so that one should be merged first.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -674,7 +674,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
 
             elif isinstance(use_module, (Library, Chip)):
                 self._loaded_modules['libs'].append(use_module.design)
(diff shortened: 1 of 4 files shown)


nmoroze:
Closing in favor of a solution that also tackles some of merge_manifest's quirks/messiness

PR #1445, opened 2023-04-06
Split up Surelog build/wheels jobs

nmoroze:
This PR ensures we shouldn't hit the 6 hour job timeout limit again. It's also a necessary precursor to parallelizing wheels builds by Python version, since there's no need to build Surelog separately in each of those jobs. 

It also provides the added benefit of letting us remove the clean build script - since we now always restore from the cache in the wheels build job, the cached paths act as a filter.

I verified this works both with a cached Surelog: https://github.com/siliconcompiler/siliconcompiler/actions/runs/4620824121, and with uncached: https://github.com/siliconcompiler/siliconcompiler/actions/runs/4622089396 (with a temp change to the key to force a miss).


gadfort (approved):
Looks great. Thanks for working on this.

PR #1446, opened 2023-04-06
implement hashalgorithm selection from keypath instead of via function

gadfort:
Implement hash algorithm selection based on the schema


nmoroze (approved):
LGTM!

PR #1447, opened 2023-04-06
Fix calc_yield/calc_area/calc_dpw

### LINKED ISSUE #820: error in calc_dpw
I was just looking through your code and noticed that calc_dpw appears to have a bug:

            while math.hypot(0, y) < radius:
                y = y + yincr
                while math.hypot(x, y) < radius:
                    x = x + xincr
                    dies = dies + 1

you are using the value of x that represents the inside corner not the outside corner in your calculation.  A extreme example is  a die of radius x radius/2 would have 0 per wafer but this function would return 4.

In the case of Y it is just inefficient to check a row that can never succeed.  Both values should be pre-incremented before checking.

gadfort:
Closes #820


aolofsson (approved):
Thanks for fixing this and for adding the tests!

PR #1449, opened 2023-04-07
Fix incorrect Chip reference in scheduler.py

WRansohoff:
Small typo in the scheduler script - hopefully we can re-enable the Slurm CI tests in the future.


siliconcompiler/scheduler.py
```diff
@@ -63,7 +63,7 @@ def _deferstep(chip, step, index, status):
             schedule_cmd.extend(['--acount', username])
```
gadfort, line 63: is this `acount` correct?

WRansohoff: No, thank you 😓

gadfort (approved):
LGTM

PR #1450, opened 2023-04-07
Fix schema bugs

gadfort:


siliconcompiler/schema/schema_obj.py
```diff
@@ -476,7 +476,7 @@ def error_msg(t):
         if field in ('author', 'filehash', 'date', 'hashalgo', 'copy') and ('file' not in sc_type):
             raise TypeError(f'Invalid field {field} for keypath {keypath}: this field only exists for file parameters')
 
-        if Schema._is_list(field, sc_type):
+        if Schema._is_list(field, sc_type) or field in ('switch',):
```
nmoroze, line 479: any particular reason not to add this in to the list of fields in `_is_list()`? The special treatment seems off. 

Does seem like switch is a weird case in the schema, but we could potentially modify `scparam()` to normalize it to be a list for all the entries

PR #1453, opened 2023-04-08
add _has_field to schema_obj to avoid try/except hack

gadfort:


nmoroze (approved):
I think this is fine, although I have a slight preference for changing the schema to always include the field. Is there a reason you think that might not work?

gadfort:
> I think this is fine, although I have a slight preference for changing the schema to always include the field. Is there a reason you think that might not work?

Adding it to every parameter as None seems like it would grow the size of the manifest by a fair bit without adding much to the the schema

PR #1454, opened 2023-04-09
Fix remote_preprocess method

WRansohoff:
This change fixes a breakage in the remote flow introduced by the large change to modularize tool/task setup.


gadfort (approved):
Thanks for fixing this

PR #1455, opened 2023-04-10
Allow dir copy and implement in core

gadfort:
Changes:
- Adds `copy` field to directories
- Update `_collect` to look for directories to copy and check imported directories for files before copying

This is a step towards being able to generate runnable testcases


nmoroze (approved):
LGTM, thanks!

PR #1457, opened 2023-04-10
add _get_tool_tasks helper function to return associated tasks

gadfort:


nmoroze (commented):
This will return things that might not be actual tasks, e.g. with KLayout we have the scripts that are consumed directly by the tool:

```python
>>> import siliconcompiler
>>> from siliconcompiler.tools.klayout import klayout
>>> chip = siliconcompiler.Chip('test')
>>> chip._get_tool_tasks(klayout)
['export', 'klayout_export', 'klayout_show', 'screenshot', 'show']
```

Unless this happens to be fine for the use case you're envisioning, maybe a check for `setup()` or something might make it a bit more robust (although would break if you happened to have a Python file sitting around in the directory with a syntax error).

nmoroze (approved):
Thanks for fixing, one other small thing: it might be good to make the try/except around the import more broad, since if someone had a Python file sitting with a syntax error in the task directory, this function would throw an exception.

PR #1463, opened 2023-04-11
Change builtin tasks from methods to regular tasks

gadfort:
Changes:
- Removes the special functions and by pass code for the builtin tasks and replaces with regular tasks that can be imported
- Implements verify and mux tasks
- Adds unit tests for new tasks


tests/builtin/test_verify.py
```diff
@@ -0,0 +1,111 @@
...
+    task = {
+        'import': parse,
+        'syn': syn_asic,
+        'teststep': verify
+    }
+
+    flow = 'testflow'
+    chip.set('option', 'flow', flow)
+
+    # Parallel flow for syn
```
nmoroze, line 29: clean up old comment, flow is linear

siliconcompiler/tools/builtin/verify.py
```diff
@@ -0,0 +1,72 @@
+from siliconcompiler.tools.builtin import _common
+from siliconcompiler.schema import Schema
+
+import re
+
+def setup(chip):
+    '''
+    Tests an assertion on a list of input tasks.
+
+    The provided steplist is verified to ensure that all assertions
```
nmoroze, line 10: "provided steplist" -> "input step"? Original wording makes it seem like it operates on entire steplist, but it only takes a single input.

siliconcompiler/tools/builtin/verify.py
```diff
@@ -0,0 +1,72 @@
...
+    are True. If any of the assertions fail, False is returned.
+    Assertions are passed in as kwargs, with the key being a metric
```
nmoroze, line 12: "as kwargs" -> "using ['flowgraph', flow, step, index, 'args']"

siliconcompiler/tools/builtin/verify.py
```diff
@@ -0,0 +1,72 @@
...
+    and the value being a number and an optional conditional operator.
```
nmoroze, line 13: is operator actually optional?

tests/builtin/test_mux.py
```diff
@@ -0,0 +1,83 @@
...
+        for metric in chip.getkeys('flowgraph', flow, 'syn', str(index), 'weight'):
+            if metric != 'setupwns':
+                chip.set('metric', metric, 1000-index*1 + 42.0, step='syn', index=index)
+
+    return chip
+
+##################################
+def test_minimum(chip):
+    flow = chip.get('option', 'flow')
+    chip.set('flowgraph', flow, 'teststep', '0', 'args', 'setuptns==0')
```
nmoroze, line 60: I don't fully understand the arguments to `mux` -- if it's just doing a min or max, does the right hand side of these expressions mean anything? `==` also seems weird to use in that case.

nmoroze (changes requested):
Left a few comments inline, but this is a nice cleanup!

I was trying to think about ways to unify the new `_gather_outputs()` and `_select_inputs()` methods with our existing methods, but I can't think of anything great. It feels to me like `gather_outputs()` type logic is a match for `setup()`, except you have the chicken-and-egg where it relies on everything else being set up first. `_select_inputs()` feels like a good match for `run()`, but that would require some mucking around with its signature.

Maybe in the future we can at least stabilize and make these APIs public so people can implement their own tasks that manipulate how inputs and outputs are passed around, that feels like it could be powerful.

nmoroze (approved):
LGTM, thanks for fixing!

siliconcompiler/tools/builtin/mux.py
```diff
@@ -22,24 +24,19 @@ def _select_inputs(chip, step, index):
...
     arguments = chip.get('flowgraph', flow, step, index, 'args')
 
-    if len(arguments) != 1:
-        chip.error(f'{step}{index} has {len(arguments)} arguments, but only support 1.', fatal=True)
-    criteria = arguments[0]
+    operations = []
+    for criteria in arguments:
+        m = re.match(r'(minimum|maximum)\((\w+)\)', criteria)
+        if not m:
+            chip.error(f"Illegal checklist criteria: {criteria}", fatal=True)
```
nmoroze, line 31: actually noticed one more thing: shouldn't say "checklist"

PR #1464, opened 2023-04-11
add cache of modules used by SC to avoid reloading them each time the…

gadfort:
…y are used

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -103,6 +103,9 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         #                   to consume in a connected HPC cluster's storage.
         self.status = {}
 
+        # Cache of python modules
+        self.modules = {}
+
         self.builtin = ['minimum','maximum',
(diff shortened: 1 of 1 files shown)


siliconcompiler/core.py
```diff
@@ -3366,6 +3368,7 @@ def node(self, flow, step, task, index=0):
            task_module = task
         elif inspect.ismodule(task):
             task_module = task.__name__
+            self.modules[task.__name__] = task
```
WRansohoff, line 3371: `self.modules[task_module]` instead of `self.modules[task.__name__]`?

WRansohoff (approved):
Thanks for adding this!

siliconcompiler/core.py
WRansohoff: (Also, while you're here, it looks like the indentation after the prior `if (isinstance(task, str)):` is 3 spaces instead of 4.)

gadfort: thanks for catching those

PR #1465, opened 2023-04-11
moving crypt and server to their own cli apps to be consistent with r…

gadfort:
…est of SC CLI apps


WRansohoff (approved):
Seems like a good idea. We should discuss whether to keep `sc-crypt`, which was built for E2EE that is not present in the public beta.

PR #1468, opened 2023-04-11
adding additional_args options to create_cmdline to allow for user specified commandline options

gadfort:
Adds:
- `additional_args` parameter to `create_cmdline` to allow for custom arguments to be added to the commandline

This is a step towards being able to generate runnable testcases

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -334,7 +334,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         return switchstrs, metavar
 
     ###########################################################################
-    def create_cmdline(self, progname, description=None, switchlist=None, input_map=None):
+    def create_cmdline(self, progname, description=None, switchlist=None, input_map=None, additional_args=None):
         """Creates an SC command line interface.
 
         Exposes parameters in the SC schema as command line switches,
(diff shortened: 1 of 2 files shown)


siliconcompiler/core.py
```diff
@@ -373,6 +373,13 @@ def create_cmdline(self, progname, description=None, switchlist=None, input_map=
                 source arguments to ['input', 'fileset', ...] keypaths based on their file
                 extension. If None, the CLI will not accept positional source
                 arguments.
+            additional_args (dict of dict): Dictionary of extra arguments to add
+                to the command line parser, with the arguments matching the
+                argparse.add_argument() call.
+
+        Returns:
+            None if additional_args is not provided, otherwise a dictionary with the
+                command line options detected from the additional_args
```
nmoroze, line 382: Consider adding example of `additional_args` in docstring

siliconcompiler/core.py
```diff
@@ -461,9 +468,25 @@ def create_cmdline(self, progname, description=None, switchlist=None, input_map=
...
+
         #Grab argument from pre-process sysargs
         cmdargs = vars(parser.parse_args(scargs))
 
+        extra_params = None
+        if additional_args:
+            # Grab user specified arguments
+            extra_params = {}
+            for arg in additional_args.keys():
+                arg = arg.replace('-', '')
```
nmoroze, line 484: The actual conversion algorithm is something like:
- If user supplies "dest" kwarg, use that, otherwise:
- Remove `-` from front of string
- Replace `-` with `_` in middle of string

I think just replacing `-` will result in bugs in some cases. I tried to find a nice way to grab a value (or the final translated key) from argparse based on the actual switch name, but couldn't find anything clean. It might be best to just mimic the full argparse logic, since it doesn't seem too bad (source: https://docs.python.org/3/library/argparse.html#dest)

nmoroze (changes requested):

siliconcompiler/core.py
gadfort: I'll just use the action that is returned from the add_argument call to get the dest that avoids this confusion

nmoroze: Oh nice, that's the sort of solution I was looking for

PR #1469, opened 2023-04-12
Support per-task options for most scheduler parameters.

WRansohoff:
We agreed to punt on this change earlier, because Schema updates should go in a major release. Re-submitting with the 0.11 tag coming up.


gadfort (changes requested):
Update schema version. Otherwise looks good.

gadfort (approved):

PR #1471, opened 2023-04-13
Fix show with entrypoints and hard coded values

gadfort:
Changes:
- show doesn't work when an entry point is used, this corrects for that by ensuring the json manifest can be found
- removes a few hardcoded 'export' and file extension values


nmoroze (approved):
Looks good to me, thanks!

Issue #1466, opened 2023-04-11
Need auto test case packaging


maliberty:
Something equivalent to ORFS's ```make <step>_issue``` is needed in SC.

PR #1473, opened 2023-04-13
Add sc-issue to generate runnable testcases

gadfort:
Adds:
- sc-issue as a cli app to be able to generate and run testcases

Fixes:
- keypaths and file handling when files are sometimes not Found

`sc-issue -generate -cfg ...` will make a `sc_issue.tar.gz` which can be shared and run using `sc-issue -run`

Closes: #1466

This is an initial version, which can be improved upon, @maliberty FYI

### DIFF
--- a/siliconcompiler/apps/sc_issue.py
+++ b/siliconcompiler/apps/sc_issue.py
@@ -22,11 +22,11 @@ def main():
     failed flow or runs an issue generated with this program.
 
     To generate a testcase, use:
-        sc-issue -generate -cfg <cfg_file>
+        sc-issue -generate -cfg <stepdir>/outputs/<design>.pkg.json
       or include a different step/index than what the cfg_file is pointing to:
-        sc-issue -generate -cfg <cfg_file> -arg_step <step> -arg_index <index>
+        sc-issue -generate -cfg <otherdir>/outputs/<design>.pkg.json -arg_step <step> -arg_index <index>
       or include specific libraries while excluding others:
-        sc-issue -generate -cfg <cfg_file> -exclude_libraries -add_library sram -add_library gpio
+        sc-issue -generate -cfg <stepdir>/outputs/<design>.pkg.json -exclude_libraries -add_library sram -add_library gpio
 
     To run a testcase, use:
         sc-issue -run -file sc_issue_<...>.tar.gz
@@ -73,6 +73,15 @@ def main():
     if switches['generate']:
         step = chip.get('arg', 'step')
         index = chip.get('arg', 'index')
+
+        if not step:
+            chip.logger.error('Unable to determine step from manifest')
+        if not index:
+            chip.logger.error('Unable to determine index from manifest')
+        if not step or not index:
+            # Exit out
+            sys.exit(1)
+
         chip._generate_testcase(step,
                                 index,
                                 switches['file'],
--- a/siliconcompiler/apps/sc_issue.py
+++ b/siliconcompiler/apps/sc_issue.py
@@ -80,7 +80,7 @@ def main():
             chip.logger.error('Unable to determine index from manifest')
         if not step or not index:
             # Exit out
-            sys.exit(1)
+            return 1
 
         chip._generate_testcase(step,
                                 index,
@@ -90,11 +90,14 @@ def main():
                                 include_libraries=not switches['exclude_libraries'],
                                 include_specific_libraries=switches['add_library'],
                                 hash_files=switches['hash_files'])
+
+        return 0
+
     if switches['run']:
         if not switches['file']:
             raise ValueError('-file must be provided')
 
-        test_dir = os.path.splitext(os.path.basename(switches['file']))[0]
+        test_dir = os.path.basename(switches['file']).split('.')[0]
         with tarfile.open(switches['file'], 'r:gz') as f:
             f.extractall(path=test_dir)
 
@@ -157,6 +160,8 @@ def main():
         # Rerun setup task
         chip._setup_task(step, index)
         chip._runtask(step, index, {}, replay=True)
+
+        return 0
 
 #########################
 if __name__ == "__main__":
--- a/siliconcompiler/apps/sc_issue.py
+++ b/siliconcompiler/apps/sc_issue.py
@@ -1,0 +1,134 @@
+# Copyright 2023 Silicon Compiler Authors. All Rights Reserved.
+import sys
+import os
+import siliconcompiler
+import tarfile
+import json
+import importlib
+
+def main():
+    progname = "sc-issue"
+    chip = siliconcompiler.Chip(progname)
+    switchlist = ['-cfg',
+                  '-version',
+                  '-arg_step',
+                  '-arg_index',
+                  '-tool_task_variable',
+                  '-tool_task_option',
+                  '-loglevel']
+    description = """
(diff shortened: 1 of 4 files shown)


siliconcompiler/core.py
```diff
@@ -2404,7 +2409,12 @@ def archive(self, step=None, index=None, all_files=False):
         Args:
             step(str): Step to archive.
             index (str): Index to archive
+            nodelist (list): List of step/index pairs to include in the archive
             all_files (bool): If True, all files are archived.
+            archive_name (str): Path to the archive
+            include_manifest (bool): If True, the final manifest will be included
+            additional_files (dict of str): Dictionary with additional files to include
+                in the archive with their associated filename in the archive
```
nmoroze, line 2417: Maybe reword a little to make it clear whether keys or values are "real" filenames vs filenames in the tarball - took me a bit to parse the sentence

siliconcompiler/apps/sc_issue.py
```diff
@@ -0,0 +1,134 @@
...
+                  '-arg_step',
+                  '-arg_index',
+                  '-tool_task_variable',
+                  '-tool_task_option',
+                  '-loglevel']
+    description = """
+    -----------------------------------------------------------
+    Restricted SC app that generates a sharable testcase from a 
+    failed flow or runs an issue generated with this program.
+    -----------------------------------------------------------
```
nmoroze, line 23: Would be helpful to include simple usage examples, I had to read the code to figure out how to use it the first time

siliconcompiler/core.py
```diff
@@ -4987,6 +5037,130 @@ def __getstate__(self):
...
+                    self.hash_files(*key, step=key_step, index=key_index)
+
+        manifest_path = os.path.join(issue_dir.name, 'manifest.json')
+        self.write_manifest(manifest_path)
+        additional_files[manifest_path] = 'manifest.json'
+
+        flow = self.get('option', 'flow')
+        tool, task = self._get_tool_task(step, index, flow=flow)
+
+        # Set copy flags for _collect
```
nmoroze, line 5066: I'm a little worried about the hard-coded groups of files to copy in, seems like it could be brittle (although doing them as major schema categories helps). I think 2 things that may be missing are `['option', 'cmdfile']` and `['fpga', 'arch']`. 

 I wonder if doing this by exclusion would be safer, i.e. set all file types to copy, except a few hardcoded paths that are unnecessary/marked as exclude by the user (I imagine 'package' stuff could be bloat?)

siliconcompiler/core.py
```diff
@@ -4987,6 +5037,130 @@ def __getstate__(self):
...
+            git_data['count'] = count
+        except git.InvalidGitRepositoryError:
+            pass
+
+        issue_information = {}
+        issue_information['environment'] = {key: value for key, value in os.environ.items()}
+        issue_information['python'] = {"path": sys.path,
+                                       "version": sys.version}
+        issue_information['date'] = datetime.datetime.fromtimestamp(time.time()).strftime('%Y-%m-%d %H:%M:%S')
+        issue_information['machine'] = Chip._get_machine_info()
```
nmoroze, line 5143: Do we expect this is something that users might be using with the remote runs (and if so, would they be running `sc-issue` locally on results that come back, or getting an issue tarball straight from the server?) If yes to both, it might be nice to include whether the run was remote in the issue summary, so you know at a glance if the machine info captured here actually reflects the machine that ran the step.

siliconcompiler/core.py
```diff
@@ -2391,7 +2396,7 @@ def _collect(self, directory=None):
                 self.error(f'Failed to copy {path}', fatal=True)
 
     ###########################################################################
-    def archive(self, step=None, index=None, all_files=False):
+    def archive(self, step=None, index=None, nodelist=None, all_files=False, archive_name=None, include_manifest=True, additional_files=None):
```
nmoroze, line 2399: Why introduce `nodelist` as a param if the testcase generator only passes in a single step/index anyway?

siliconcompiler/core.py
```diff
@@ -4987,6 +5037,130 @@ def __getstate__(self):
...
+        issue_information['version'] = {'schema': self.schemaversion,
+                                        'sc': self.scversion,
+                                        'git': git_data}
+
+        issue_path = os.path.join(issue_dir.name, 'issue.json')
+        with open(issue_path, 'w') as fd:
+            json.dump(issue_information, fd, indent=4, sort_keys=True)
+        additional_files[issue_path] = 'issue.json'
+
+        self.archive(nodelist=[(step, index)], all_files=True, include_manifest=False, archive_name=archive_name, additional_files=additional_files)
```
nmoroze, line 5157: I wonder if it would make sense to factor out some of `archive()`'s functionality into a private method that gets used here, rather than adding parameters to `archive()` itself. I'm thinking something along the lines of an `_archive_node()` that takes in a `Tarfile`, `step`, and `index`, and implements the `archive()` inner-loop logic. The `Tarfile` could be instantiated by the caller, and they could add any additional files directly to that. 

I think it would help keep the public API a bit cleaner, and I had this reaction after realizing that the additional files passed in here don't necessarily come from the job directory (whereas when I first read the API changes I assumed `archive()` would be searching for `additional_files` relative to the job being archived).

nmoroze (changes requested):

siliconcompiler/core.py
gadfort: Once everything is working, I think it's something that the server should handle on a failure. At the moment it's meant for local runs

siliconcompiler/core.py
gadfort: reworked, lets see if that is better

siliconcompiler/apps/sc_issue.py
gadfort: added

siliconcompiler/core.py
gadfort: removed

siliconcompiler/core.py
gadfort: refactored

siliconcompiler/apps/sc_issue.py
```diff
@@ -0,0 +1,154 @@
...
+                  '-tool_task_variable',
+                  '-tool_task_option',
+                  '-loglevel']
+    description = """
+    -----------------------------------------------------------
+    Restricted SC app that generates a sharable testcase from a 
+    failed flow or runs an issue generated with this program.
+
+    To generate a testcase, use:
+        sc-issue -generate -cfg <cfg_file>
```
nmoroze, line 25: This is a minor thing, but when I first ran this I accidentally pointed at the final manifest in the job directory (since it was a successful run) and did not include -arg_step and -arg_index, which made it fail out with a slightly gnarly error.

I think making a note to point that you need to point to a step-specific manifest (maybe just embedding a path fragment in the example like below) might be helpful, and/or including error checking for step/index being defined.

```
sc-issue -generate <stepdir>/outputs/<design>.pkg.json
```

nmoroze (approved):
Sweet, thanks for implementing the feedback. This looks great! Had one more tiny note about docs but otherwise good to go.

siliconcompiler/apps/sc_issue.py
gadfort: good point

PR #1475, opened 2023-04-14
allow load_target to take in a module or a string

gadfort:
Changes:
- make load_target work like .use() by allowing a python module to be passed in as well as a string


nmoroze (approved):
Nice, thanks!

PR #1489, opened 2023-04-20
Cleanup sc-server round 1

gadfort:
Changes:
- fix sc-configure to allow for port to be entered
- removed calls to mkdir via subprocess and instead use builtin makedirs
- removed unused code from sc-server
- added local server tests to CI and updated to sc-server to function again.

This is a step towards getting all server tests with slurm enabled as well.

### DIFF
--- a/siliconcompiler/server.py
+++ b/siliconcompiler/server.py
@@ -159,8 +159,10 @@ class Server:
         # Create a dummy Chip object to make schema traversal easier.
         # TODO: if this is a dummy Chip we should be able to use Schema class,
         # but looks like it relies on chip.status.
-        chip = Chip('server')  # start with a dummy name, as this will be overwritten
-        chip.schema = Schema(cfg=cfg) # Add provided schema
+        # start with a dummy name, as this will be overwritten
+        chip = Chip('server')
+        # Add provided schema
+        chip.schema = Schema(cfg=cfg)
 
         # Fetch some common values.
         design = chip.get('design')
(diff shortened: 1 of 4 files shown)


siliconcompiler/server.py
```diff
@@ -156,8 +159,8 @@ async def handle_remote_run(self, request):
         # Create a dummy Chip object to make schema traversal easier.
         # TODO: if this is a dummy Chip we should be able to use Schema class,
         # but looks like it relies on chip.status.
-        chip = Chip(cfg['design']['value'])
-        chip.schema.cfg = cfg
+        chip = Chip('server')  # start with a dummy name, as this will be overwritten
+        chip.schema = Schema(cfg=cfg) # Add provided schema
```
WRansohoff, line 163: This is a very minor comment, but I'm not sure about putting comments at the end of lines in our public code. It's nice for brief examples and documentation, but maybe move these comments to the line above?

siliconcompiler/server.py
```diff
@@ -419,35 +417,40 @@ async def remote_sc_auth(self, chip, username, pk):
...
+            keypath = f'{build_dir}/pk'
             with open(os.open(keypath, os.O_CREAT | os.O_WRONLY, 0o400), 'w+') as keyfile:
                 keyfile.write(pk)
             chip.write_manifest(f"{build_dir}/configs/chip{chip.get('option', 'jobname')}.json")
-            # Create the command to run.
-            run_cmd  = f"cp -R {from_dir}/* {to_dir}/ ; "
+
+            utils.copytree(from_dir, build_dir, dirs_exist_ok=True)
+
+            utils.copytree(from_dir, build_dir, dirs_exist_ok=True)
```
WRansohoff, line 441: Do we need both `copytree` calls?

WRansohoff (approved):
LGTM, just one possibly-duplicate call that you might want to remove.

siliconcompiler/server.py
gadfort: looks like I was a little trigger happy with the copy and paste.

PR #1490, opened 2023-04-21
Remove subprocess from server code + minor cleanup

gadfort:
Changes:
- Removes subprocess calls to tar infavor of using tarfile
- Plus minor code formatting
- Adding nfs_mount to Server object  (real solution is probably to cleanup its schema accesses)


WRansohoff (approved):
Couple of small comments, nothing blocking

PR #1492, opened 2023-04-22
Fix Schema._is_empty() [SC-330]

nmoroze:
This PR fixes a bug in the Schema._is_empty() function that caused it to always return False. It also includes a few other fixes of issues that were masked by this bug:

- Need to add a `run()` function to the fake tools - each tool needs either a `run()` or the exe param defined, but the check wasn't working previously
- No more pruning of manifests dumped for tools - with is_empty fixed the pruning becomes a lot more aggressive, which makes a lot of keypaths no longer exist in the Schema in the KL driver


gadfort (approved):
Lgtm

PR #1496, opened 2023-04-23
fix server credentials finding

gadfort:
When a relative path is provided to the credentials, which contains no directory, the isdir returns false. This ensures the path used is an abs path do the directory component will have something and can fail out correctly.


WRansohoff:
What exactly is this fixing? The original purpose of `('option', 'credentials')` was to let users specify an alternative absolute or relative path to a credentials file. Credentials are secrets, I'm not sure they should be looked for in repositories or Python module search paths.

gadfort:
@WRansohoff updated the description a little and fixed the implementation

Issue #1071, opened 2022-06-30
Cryptography deprecating Python 3.6


nmoroze:
I've noticed when running SC under Python 3.6 that a message pops up saying support for this version is being deprecated by the cryptography package. Also noted in issue #1066:  
```
.../venv/lib64/python3.6/site-packages/siliconcompiler/crypto.py:3: CryptographyDeprecationWarning: Python 3.6 is no longer supported by the Python core team. Therefore, support for it is deprecated in cryptography and will be removed in a future release.
```

I think we'll want to keep supporting Python 3.6 for the time being. @WRansohoff do you know if cryptography is still a required dependency, now that we no longer support encrypting jobs? If not, maybe we should remove any stale crypto code and remove it as a dependency.

PR #1497, opened 2023-04-24
Remove crypto.py

gadfort:
This removes the long deprecated crypto.py module along with associated code.
There should be no function changes with this since this code is not used anywhere.
A follow-on PR will reenable the CI for authenticated server.py flows

Closes: #1071


WRansohoff (approved):

WRansohoff:
Oh, but we can probably remove `cryptography` from `requirements.txt` too.

gadfort:
> Oh, but we can probably remove `cryptography` from `requirements.txt` too.

good catch, will do

PR #1498, opened 2023-04-24
fix incorrect detection of invalid url

gadfort:
Fixes:
- when running sc-configure localhost:8080 it should detect that this is missing the scheme and therefore needs a dummy one pre-pended to parse the hostname correctly.

### DIFF
--- a/siliconcompiler/apps/sc_configure.py
+++ b/siliconcompiler/apps/sc_configure.py
@@ -70,7 +70,7 @@ def main():
 
     server = urlparse(srv_addr)
     has_scheme = True
-    if not server.scheme:
+    if not server.hostname:
         # fake add a scheme to the url
         has_scheme = False
         server = urlparse('https://' + srv_addr)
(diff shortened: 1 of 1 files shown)


siliconcompiler/apps/sc_configure.py
```diff
@@ -70,7 +70,7 @@ def main():
 
     server = urlparse(srv_addr)
     has_scheme = True
-    if not server.scheme:
+    if not server.hostname:
```
WRansohoff, line 73: Won't it just return early 5 lines down, when `if not server.hostname:` is checked again?

gadfort: The second check is the last check. Then it's first checked it's to see if the url was `localhost` or `https://localhost`, if it was the latter, the parsing worked correctly, if not, then we need to make it appear as if it is a real URL. If the parsing still fails to find the hostname then we exit out.

WRansohoff: Oh, I see - `server` is a urlparse object which has a `hostname` property that gets updated when `urlparse` is called.

WRansohoff (approved):

PR #1502, opened 2023-04-25
enable server auth flow in CI

gadfort:
Changes:
- removes the bifurcated flow for authenticated and non-authenticated runs in server.py
- enables one test for auth flow (disables another do to inconsistent messaging types).
- Make use of asyncio for CI to provide unused ports so port number are not hardcoded.


tests/server/test_gcd_server_auth.py
```diff
@@ -0,0 +1,115 @@
...
+            'password': user_pwd,
+        }]}))
+
+    # Start running an sc-server instance.
+    srv_proc = subprocess.Popen(['sc-server',
+                                 '-nfs_mount', './local_server_work',
+                                 '-cluster', 'local',
+                                 '-port', str(unused_tcp_port),
+                                 '-auth'])
+    time.sleep(10)
```
WRansohoff, line 87: Maybe `from .test_server import SERVER_STARTUP_DELAY`, and use that for these tests?

tests/server/test_server.py
```diff
@@ -11,17 +11,12 @@
 SERVER_STARTUP_DELAY = 10
 
 @pytest.fixture
-def gcd_remote_test(gcd_chip, request):
-    # Get the port number; avoid re-use to enable parallel tests.
-    if 'port' in request.keywords:
-        port = request.keywords["port"]
-    else:
-        port = '8080'
+def gcd_remote_test(gcd_chip, unused_tcp_port):
```
WRansohoff, line 14: Where do these `unused_tcp_port` values come from?

WRansohoff (approved):
Looks good, thanks! Just a couple of non-blocking questions.

tests/server/test_gcd_server_auth.py
gadfort: this is a stop gap, I don't think the time.sleep is correct anyways

PR #1503, opened 2023-04-25
add slurm to ci container and enable ci for slurm

gadfort:
This adds back the slurm CI for one test for now.
To enable that test, several changes in the scheduler.py were needed.
1. job_hash may not be set for local runs
2. queue may not be set, so just picking the first queue in those cases
3. removing hard coded values

### DIFF
--- a/setup/docker/slurm/cgroup.conf
+++ b/setup/docker/slurm/cgroup.conf
@@ -1,0 +1,1 @@
+CgroupPlugin=cgroup/v1
--- a/setup/docker/slurm/start_slurm.sh
+++ b/setup/docker/slurm/start_slurm.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+# Set the container's hostname in slurm.conf
+sed "s^{{ hostname }}^`hostname`^g" /etc/slurm/slurm.conf.in >  /etc/slurm/slurm.conf
+
+# Start munge and slurm daemons
+/etc/init.d/munge start
+slurmctld
+slurmd
--- a/siliconcompiler/scheduler.py
+++ b/siliconcompiler/scheduler.py
@@ -1,72 +1,74 @@
-import base64
 import os
 import shlex
 import subprocess
 import stat
 import time
+import uuid
+import json
 
 ###########################################################################
 def _deferstep(chip, step, index, status):
     '''
     Helper method to run an individual step on a slurm cluster.
-    If a base64-encoded 'decrypt_key' is set in the Chip's status
-    dictionary, the job's data is assumed to be encrypted,
-    and a more complex command is assembled to ensure that the data
-    is only decrypted temporarily in the compute node's local storage.
     '''
 
     # Determine which HPC job scheduler being used.
     scheduler_type = chip.get('option', 'scheduler', 'name', step=step, index=index)
 
+    if scheduler_type != 'slurm':
+        raise ValueError(f'{scheduler_type} is not a supported scheduler')
+
     # Determine which cluster parititon to use. (Default value can be overridden on per-step basis)
     partition = chip.get('option', 'scheduler', 'queue', step=step, index=index)
-    # Get the temporary UID associated with this job run.
-    job_hash = chip.status['jobhash']
+    if not partition:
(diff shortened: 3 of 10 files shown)


setup/docker/slurm/cgroup.conf
```diff
@@ -0,0 +1 @@
+CgroupPlugin=cgroup/v1
```
WRansohoff, line 1: You might be able to just `RUN echo "CgroupPlugin=cgroup/v1" >> /etc/slurm/cgroup.conf` in the Dockerfile.

setup/docker/slurm/start_slurm.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Set the container's hostname in slurm.conf
+sed "s^{{ hostname }}^`hostname`^g" /etc/slurm/slurm.conf.in >  /etc/slurm/slurm.conf
```
WRansohoff, line 4: You could also copy the file directly to `/etc/slurm/slurm.conf` and `sed -i "[pattern]" /etc/slurm/slurm.conf`

siliconcompiler/scheduler.py
```diff
@@ -3,68 +3,71 @@
...
     # Determine which HPC job scheduler being used.
     scheduler_type = chip.get('option', 'scheduler', 'name', step=step, index=index)
 
+    if scheduler_type != 'slurm':
+        raise ValueError(f'{scheduler_type} is not a supported scheduler')
+
     # Determine which cluster parititon to use. (Default value can be overridden on per-step basis)
     partition = chip.get('option', 'scheduler', 'queue', step=step, index=index)
+    if not partition:
+        partition = _get_slurm_partitions()
```
WRansohoff, line 24: Should the method name be singular instead of plural, since it only returns the first listed partition?

siliconcompiler/scheduler.py
```diff
@@ -3,68 +3,71 @@
...
 
-    # Job data is not encrypted, so it can be run in shared storage.
     # Write out the current schema for the compute node to pick up.
     job_dir = chip._getworkdir()
     cfg_dir = f'{job_dir}/configs'
     cfg_file = f'{cfg_dir}/{step}{index}.json'
     if not os.path.isdir(cfg_dir):
         os.mkdir(cfg_dir)
-    chip.unset('option', 'scheduler', 'name', step=step, index=index)
+    chip.set('option', 'scheduler', 'name', None, step=step, index=index)
```
WRansohoff, line 41: Is `unset` unsuitable here? Does it fail to unset a globally-defined value when step/index are specified?

WRansohoff (approved):
Looks good, thanks for making these updates. Just a few small questions and suggestions.

setup/docker/slurm/cgroup.conf
gadfort: I'd like to get away from scripts emitting scripts where possible, so even though this is a pretty simple file, it makes it clear what it's values are without having to hunt around in the Dockerfile

setup/docker/slurm/start_slurm.sh
gadfort: I debated that, but after needing to debug something once, I decided I wanted to have the original file available in case we need to continue debugging.

siliconcompiler/scheduler.py
gadfort: Good point

siliconcompiler/scheduler.py
gadfort: Yes, the `.unset` unfortunately does not behave correctly (that was a fun loop to hunt for).
@nmoroze is going to fix it, but where we are using it, it's likely not doing what we want.

setup/docker/slurm/start_slurm.sh
WRansohoff: Okay, makes sense.

siliconcompiler/scheduler.py
WRansohoff: Okay - maybe comment out the `unset` with a TODO, if we intend to fix it?

PR #1504, opened 2023-04-26
Continue to update server.py to mirror expected behavior in client.py

gadfort:
This changes the responses from server.py to `json_responses` to match the expected behavior in client.py.
This change allows for the enablement of the authenticated flow tests including the failure to auth tests.

There are no tests for check_progress yet, so that will happen in a followon PR


siliconcompiler/server.py
```diff
@@ -313,39 +249,16 @@ async def handle_check_progress(self, request):
...
-                else:
-                    return web.Response(text="Error: some authentication parameters are missing.", status=400)
-            else:
-                return web.Response(text="Error: authentication parameters were passed in, but this server does not support that feature.", status=500)
+        job_hash = job_params['job_hash']
+        username = job_params['username']
 
         # Determine if the job is running.
-        if "%s%s_%s"%(username, job_hash, jobid) in self.sc_jobs:
+        if "%s%s"%(username, job_hash) in self.sc_jobs:
```
WRansohoff, line 261: Some of these formatting strings are a bit old; I think the preferred way to format strings in Python 3 is `f"{username}{job_hash}"` instead of the old `"%s%s"%(username, job_hash)` notation.

siliconcompiler/server.py
```diff
@@ -407,6 +319,36 @@ def auth_password(self, username, password):
             return False
         return (password == self.user_keys[username]['password'])
 
+    def __handle_request(self, request):
```
WRansohoff, line 322: It looks like this function is for common input validation and error-checking, but "handle_request" sounds like a general-purpose method. Maybe rename it to something like `__check_request` or `__verify_request`?

WRansohoff (approved):
Looks good, just a couple of small suggestions.

siliconcompiler/server.py
gadfort: I have a followon PR (not open yet) that fixes this

PR #1506, opened 2023-04-26
Update server.py to include check_user API and add sc-ping to CI

gadfort:
Changes:
- Allow for sc-ping to take in non-default credentials file
- Add sc-ping to CI
- Fix sc-configure and sc-ping to use argparse


siliconcompiler/apps/sc_ping.py
```diff
@@ -72,16 +69,16 @@ def main():
...
             print('Error fetching user information from the remote server.')
-            sys.exit(1)
+            return 1
 
         # Print the user's account info, and return.
-        print(f'User {remote_cfg["username"]} validated successfully!')
-        print(f'  Remaining compute time: {(user_info["compute_time"]/60.0):.2f} minutes')
+        print(f'User {remote_cfg["username"]}:')
+        print(f'  Remaining compute time: {(user_info["bandwidth_kb"]/60):.2f}) minutes')
         print(f'  Remaining results bandwidth: {user_info["bandwidth_kb"]} KiB')
```
WRansohoff, line 77: Use `"compute_time"` as the key for "Remaining compute time".

siliconcompiler/server.py
```diff
@@ -61,9 +61,14 @@ def __init__(self, cmdlinecfg, loglevel="DEBUG"):
...
-                    self.user_keys[mapping['username']] = {
-                        'password': mapping['password'],
+                    username = mapping['username']
+                    self.user_keys[username] = {
+                        'password': mapping['password']
                     }
+                    if 'compute_time' in mapping:
+                        self.user_keys[username]['compute_time'] = mapping['compute_time']
+                    if 'bandwidth' in mapping:
+                        self.user_keys[username]['bandwidth'] = mapping['bandwidth']
```
WRansohoff, line 71: Default to 0?

siliconcompiler/utils.py
```diff
@@ -230,3 +231,8 @@ def format_fileset_type_table(indent=12):
         table += f"{indent}{filetype:<10}| {fileset:<11}| {ext}\n"
 
     return table
+
+def default_credentials_file():
+    cfg_file = os.path.join(Path.home(), '.sc', 'credentials')
```
WRansohoff, line 236: `Path` objects also overload the `/` operator, you could do:

`Path.home() / '.sc' / 'credentials'`

...Although personally, I think that `os.path.join` is a little more clear.

WRansohoff (approved):
Looks good, I think the only thing that needs addressing is the duplicated reporting of "bandwidth_kb".

Also, the docs might need updating to reflect the new `sc-configure` syntax if the argparse implementation changed that at all.

siliconcompiler/utils.py
gadfort: I've seen that used before, but I find it somewhat confusing at first

PR #1508, opened 2023-04-27
Add flake8 lint to CI

gadfort:
This adds flake8 as a linting CI. 
Currently there are a lot of excluded checks that we need to work towards reducing, which can be done one at a time.

The changes in the code are all from fixing issues raised by the E1** lint checks. There should be no functional changes.


nmoroze (approved):
This makes me very happy, thanks!

PR #1511, opened 2023-04-28
enable slurm server test

gadfort:
Enables the server slurm test
and removes tests are are already covered by other tests


WRansohoff (approved):
Looks good, as long as you intend to discard the test for an authenticated Slurm run.

gadfort:
> Looks good, as long as you intend to discard the test for an authenticated Slurm run.

@WRansohoff yes,  that test didn't provide any new coverage since the only account getting checked is the server.py account anyways.

PR #1514, opened 2023-04-28
Avoid using -r requirements for installs

gadfort:
@aulihant this solves the issue of the lint on 3.6 with the new docs build. It also makes it a little easier to install since before we asked for two install steps and now can just do one

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,8 +92,8 @@ To install the project from source (recommended for developers only).
 ```bash
 git clone https://github.com/siliconcompiler/siliconcompiler
 cd siliconcompiler
-python -m pip install -e .
-python -m pip install -e .[docs,test]  # Install dependencies for generating docs and running tests
+python -m pip install -e .             # Required install step
+python -m pip install -e .[docs,test]  # Optional install step for generating docs and running tests
 ```
 
 # Tool Installation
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -92,8 +92,8 @@ To install the project from source (recommended for developers only).
 ```bash
 git clone https://github.com/siliconcompiler/siliconcompiler
 cd siliconcompiler
-pip install -r requirements.txt
 python -m pip install -e .
+python -m pip install -e .[docs,test]  # Install dependencies for generating docs and running tests
```
nmoroze, line 96: Could technically skip the bare `python -m pip install -e .`, unless the idea is to give someone a path forward if the docs/test deps install fail?

nmoroze (approved):

README.md
gadfort: The idea was to give regular users the basic choice and developers the added sections. Maybe that is not 100% clear

aulihant: I think it _would_ be more clear to update to this:
```python -m pip install -e .              # Required install step```
```python -m pip install -e .[docs,test]  # Optional install step for generating docs and running tests```

aulihant (approved):
Confirmed instructions work; updates to verbiage on what is required and what is optional install step is a nice to have, but not a deal breaker for me.

PR #1518, opened 2023-05-01
dont take all of scpath and ensure task io is skipped

gadfort:
sc-issue was attempting to gobble up all of the scpaths if its set and missing the input/output/report keys causing the generation to fail when executed on a failed run.


nmoroze (approved):
That certainly does not seem ideal

PR #1519, opened 2023-05-01
Fix wheels

nmoroze:
Looks like the failing wheels build was due to a recent update in scikit-build that affects how the location of Python libraries are reported to CMake. I based the fix on notes from https://github.com/scikit-build/scikit-build/issues/963 and https://github.com/Qiskit/qiskit-aer/pull/1786

Fix verified in https://github.com/siliconcompiler/siliconcompiler/actions/runs/4854034031. I did notice that one of the KL show tests timed out for macOS, but is fine after re-running. Might be something to keep an eye on, not sure if we need to bump the timeout or if it's just a weird transient thing...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
 # Packages we rely on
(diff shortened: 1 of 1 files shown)


gadfort (approved):
Let's keep an eye on the timeout issue and if its a thing we'll take a look at the timeout and the test running

PR #1522, opened 2023-05-02
Continue with lint updates

gadfort:
This enables all but the E502 checks (line length).
There should be no functional changes as its mainly formatting.


nmoroze (approved):
Looks good, today I learned that PEP-8 calls for 2 blank lines around top-level functions.

PR #1524, opened 2023-05-03
Fix load target order

gadfort:
Ensures the target is loaded after the options are set and makes sure the provided targets are non-clobbering


gadfort:
@volkamir FYI

nmoroze (approved):
LGTM

PR #1525, opened 2023-05-03
Gracefully handle duplicate edges

nmoroze:
If a user accidentally adds duplicate edges to a flowgraph, they'll get a confusing error mid run: 

```
| INFO    | job0  | mystep    | 0  | Finished task in 60s
Process Process-4:
  File "/opt/rh/rh-python38/root/usr/lib64/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/opt/rh/rh-python38/root/usr/lib64/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/centos/venv/lib64/python3.8/site-packages/siliconcompiler/core.py", line 3608, in _runtask
    utils.copytree(f"../../../{in_job}/{in_step}/{in_index}/outputs", 'inputs/',
  File "/home/centos/venv/lib64/python3.8/site-packages/siliconcompiler/utils.py", line 30, in copytree
    os.link(srcfile, dstfile)
FileExistsError: [Errno 17] File exists: '../../../job0/mystep/0/outputs/design.out' -> 'inputs/design.out'
```

This PR fixes the problem in a couple places: it performs validation in `edge()` with a warning message, and adds a manifest check as a fallback (in case a user is manipulating the `flowgraph` schema directly).

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3429,13 +3429,12 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                 self.error(f'Illegal step name: {step} is reserved')
                 return
 
-        # Adding
         tail_node = (tail, str(tail_index))
         if tail_node in self.get('flowgraph', flow, head, str(head_index), 'input'):
-            self.logger.warning(f'Edge from {tail}{tail_index} -> {head}{head_index} already exists, skipping')
+            self.logger.warning(f'Edge from {tail}{tail_index} to {head}{head_index} already exists, skipping')
             return
 
-        self.add('flowgraph', flow, head, str(head_index), 'input', (tail, str(tail_index)))
+        self.add('flowgraph', flow, head, str(head_index), 'input', tail_node)
 
     ###########################################################################
     def graph(self, flow, subflow, name=None):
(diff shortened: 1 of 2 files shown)


siliconcompiler/core.py
```diff
@@ -3419,8 +3429,12 @@ def edge(self, flow, tail, head, tail_index=0, head_index=0):
...
                 return
 
-        # Adding
-        self.add('flowgraph', flow, head, str(head_index), 'input', (tail, str(tail_index)))
+        tail_node = (tail, str(tail_index))
+        if tail_node in self.get('flowgraph', flow, head, str(head_index), 'input'):
+            self.logger.warning(f'Edge from {tail}{tail_index} to {head}{head_index} already exists, skipping')
+            return
+
+        self.add('flowgraph', flow, head, str(head_index), 'input', tail_node)
```
gadfort, line 3437: why not cast the head and tail index to str() at the start of this method so the types are consistent throughout?

gadfort (approved):
Thanks for catching this. Consider the casting comment, but I'm okay with merging either way.

siliconcompiler/core.py
nmoroze: Just a copy-paste thing, but I'll clean it up - nice suggestion.

PR #1528, opened 2023-05-04
Configure docs build to use newer OS

nmoroze:
Need to specify a newer build environment for RTD due to a breaking change in a transitive dependency. 

Fix based on https://github.com/readthedocs/readthedocs.org/issues/10290#issuecomment-1535120995.


gadfort (approved):
Thanks for fixing.

PR #1529, opened 2023-05-04
fix last two lint violations

gadfort:


nmoroze (approved):
Thanks for fixing. Looks like you need to shave 2 more characters off line 3498 (frustrating...) but feel free to merge once green.

PR #1537, opened 2023-05-09
add extension to sc-show to allow users to pick the file type easily when multiple displayble files are generated

gadfort:
Adds: `-ext` option to sc-show to allow users to select the filetype to display. This is useful when multiple filetypes are available and you need a particular one, like def or odb.

### DIFF
--- a/siliconcompiler/apps/sc_show.py
+++ b/siliconcompiler/apps/sc_show.py
@@ -55,10 +55,19 @@ def main():
         if ext in default_input_map:
             input_map[ext] = default_input_map[ext]
 
-    chip.create_cmdline(progname,
-                        switchlist=['-design', '-input', '-loglevel', '-cfg'],
-                        description=description,
-                        input_map=input_map)
+    extension_arg = {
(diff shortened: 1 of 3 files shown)


siliconcompiler/apps/sc_show.py
```diff
@@ -55,10 +55,19 @@ def main():
         if ext in default_input_map:
             input_map[ext] = default_input_map[ext]
 
-    chip.create_cmdline(progname,
-                        switchlist=['-design', '-input', '-loglevel', '-cfg'],
-                        description=description,
-                        input_map=input_map)
+    extension_arg = {
+        'metavar': '<ext>',
+        'help': 'Specify the extention of the file to show'
```
nmoroze, line 60: typo: extention -> extension

nmoroze: Also, I think it could be helpful to specify default behavior/explicitly mention that this flag is optional.

nmoroze (changes requested):

PR #1538, opened 2023-05-09
allow library to specify cts buffer cell for openroad

gadfort:
Adds ability for library to specify openroad cts buffer
Fills in clkbuf and buf section of sky130


nmoroze (commented):
Do you plan to use the `'option', 'var', 'openroad_cts_clock_buffer'` in a future update? It looks like the actual driver logic just switched from using the first entry in the overall list to the last.

gadfort:
@nmoroze on line 220, the openroad driver copies in the values in openroad_cts_clock_buffer to cts_clock_buffer (this is a single entry not a list, the tcl driver was fixed to ensure a single entry is used instead of a list.)

nmoroze (approved):
Ah I see, didn't realize there was existing functionality for that. To check my understanding, the behavior for these variables (in terms of values they could be in priority order):

Value explicitly set by user > value provided in [option, var] by library > default from `variables` list?

gadfort:
@nmoroze yes that is the correct priority order. I need to rewrite the openroad driver to pull more defaults from the pdk and mainlib

PR #1539, opened 2023-05-09
[sc-issue] update to make testcases self-contained

gadfort:
Changes:
- `_find_files` can now return relative paths if the private variable `__relative_path` is set. This allows the testcase generator to build testcases with complete tests.
- Fixes incorrect libraries key path
- Rework the testcase generator to include a README.txt and a run.sh file to allow for standalone tests.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1311,7 +1311,9 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         elif len(keypath) >= 5 and keypath[0] == 'tool' and keypath[4] == 'script':
             tool = keypath[1]
             task = keypath[3]
-            refdirs = self._find_files('tool', tool, 'task', task, 'refdir', step=step, index=index, abs_path_only=True)
+            refdirs = self._find_files('tool', tool, 'task', task, 'refdir',
+                                       step=step, index=index,
+                                       abs_path_only=True)
             search_paths = refdirs
 
         for path in paths:
@@ -5301,12 +5303,14 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                 try:
                     # Rerun pre_process
                     func(self)
-                except Exception as e:
+                except Exception:
                     pass
         except SiliconCompilerError:
             pass
 
-        self._makecmd(tool, task, step, index, script_name=f'{self._getworkdir(step=step, index=index)}/replay.sh', include_path=False)
+        self._makecmd(tool, task, step, index,
+                      script_name=f'{self._getworkdir(step=step, index=index)}/replay.sh',
+                      include_path=False)
 
         # Restore normal path behavior
         self.__relative_path = None
@@ -5368,7 +5372,9 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         with open(issue_path, 'w') as fd:
             json.dump(issue_information, fd, indent=4, sort_keys=True)
 
-        jinja_env = Environment(loader=FileSystemLoader(os.path.join(self.scroot, 'templates', 'issue')))
+        jinja_env = Environment(loader=FileSystemLoader(os.path.join(self.scroot,
+                                                                     'templates',
+                                                                     'issue')))
         readme_path = os.path.join(issue_dir.name, 'README.txt')
         with open(readme_path, 'w') as f:
             f.write(jinja_env.get_template('README.txt').render(
@@ -5393,9 +5399,14 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                          issue_path,
                          readme_path,
                          run_path]:
-                tar.add(os.path.abspath(path), arcname=os.path.join(arch_base_dir, os.path.basename(path)))
+                tar.add(os.path.abspath(path),
+                        arcname=os.path.join(arch_base_dir,
+                                             os.path.basename(path)))
 
-            tar.add(job_dir, arcname=os.path.join(arch_base_dir, os.path.relpath(job_dir, issue_dir.name)))
+            tar.add(job_dir,
+                    arcname=os.path.join(arch_base_dir,
+                                         os.path.relpath(job_dir,
+                                                         issue_dir.name)))
 
         issue_dir.cleanup()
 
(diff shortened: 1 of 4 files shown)


siliconcompiler/core.py
```diff
@@ -1306,7 +1311,9 @@ def _find_files(self,
         elif len(keypath) >= 5 and keypath[0] == 'tool' and keypath[4] == 'script':
             tool = keypath[1]
             task = keypath[3]
-            refdirs = self._find_files('tool', tool, 'task', task, 'refdir', step=step, index=index)
+            refdirs = self._find_files('tool', tool, 'task', task, 'refdir',
+                                       step=step, index=index,
+                                       abs_path_only=True)
```
nmoroze, line 1316: Why do you need `abs_path_only` here?

siliconcompiler/core.py
```diff
@@ -3552,6 +3569,19 @@ def pipe(self, flow, plan):
                 self.edge(flow, prevstep, step)
             prevstep = step
 
+    ###########################################################################
+    def __write_task_manifest(self, tool, path=None, backup=True):
```
nmoroze, line 3573: Is a two underscore prefix a distinct meaning from one underscore?

siliconcompiler/core.py
```diff
@@ -3552,6 +3569,19 @@ def pipe(self, flow, plan):
...
+        suffix = self.get('tool', tool, 'format')
+        if suffix:
+            manifest_path = f"sc_manifest.{suffix}"
+            if path:
+                manifest_path = os.path.join(path, manifest_path)
+
+            if backup and os.path.exists(manifest_path):
+                shutil.copyfile(manifest_path, f'{manifest_path}.bak')
```
nmoroze, line 3581: Why add the backup?

siliconcompiler/core.py
```diff
@@ -5176,67 +5206,131 @@ def _generate_testcase(self,
...
             copy = True
-            if keypath[0] == 'libraries':
+            if keypath[0] == 'library':
                 # only copy libraries if selected
                 if include_specific_libraries and keypath[1] in include_specific_libraries:
                     copy = True
                 else:
                     copy = include_libraries
+
+                copy &= determine_copy(*keypath[2:])
```
nmoroze, line 5227: Looks clean but using bitwise operators on booleans in Python makes me feel uneasy... might be best to do `copy = copy and determine_copy(...)`

nmoroze: (using the boolean `and` also has a performance benefit since it'll short-circuit if copy is already false)

siliconcompiler/core.py
```diff
@@ -5176,67 +5206,131 @@ def _generate_testcase(self,
...
             elif keypath[0] == 'pdk':
                 # only copy pdks if selected
                 if include_specific_pdks and keypath[1] in include_specific_pdks:
                     copy = True
                 else:
                     copy = include_pdks
+
+                copy &= determine_copy(*keypath[2:])
```
nmoroze, line 5235: What is nested under 'pdk' that warrants the recursive call?

siliconcompiler/core.py
```diff
@@ -105,6 +105,10 @@ def __init__(self, design, loglevel=None):
         # Cache of python modules
         self.modules = {}
 
+        # Controls whether find_files returns an abspath or relative to this
+        # this is primarily used when generating standalone testcases
+        self.__relative_path = None
```
nmoroze, line 110: Having this extra state that affects `_find_files()` wouldn't be my preferred solution, although happy to let it through if there's nothing better.

As an alternate proposal, could you add optional args to pass in the parent path to `write_manifest()` and `_make_cmd()`? I think the behavior for `write_manifest()` could end up getting pushed down and implemented in `Chip._abspath()`. This seems cleaner to me but I could be missing something.

nmoroze (changes requested):

siliconcompiler/core.py
gadfort: Nothing, I guess I got too copy paste excited. I'll remove

siliconcompiler/core.py
gadfort: Will do

siliconcompiler/core.py
gadfort: In the testcase, this file gets overwritten and I wanted to keep a backup of it uncase there were problems with replay script that gets generated. There is something else to inspect.

siliconcompiler/core.py
gadfort: __ gets renamed at runtime to __chip_somethign. This ensures private variables are very difficult to access, while _ is just by convention private, __ is a little more private because the name gets modified by the file it's coming from. This prevents users from modifying the relative path this even by accident. https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name

siliconcompiler/core.py
gadfort: This is building the search path which still needs to be abspath. We only want the final result to be relative.

siliconcompiler/core.py
gadfort: I chose this because when a tool driver calls find_files in preprocess, runtime_options, or setup, I don't want to force them to update their code to handle relative paths and this allows us to control the global behavior of that function while the testcase is being put together. I'm open to suggestions, but it's the tool drivers that are a challenge here.

siliconcompiler/core.py
nmoroze: Ah yeah, this is vaguely familiar. Thanks for the explanation! I imagine it would make sense to move over to double underscore in a lot of other cases/adopt it for things going forward as you've done here.

siliconcompiler/core.py
nmoroze: That makes sense, although if my understanding of how this works is correct - doesn't the tool driver execution still happen in the original context when the test-case is being packaged, so it's safe for it to operate fully with abs paths, as long as all the stuff that gets dumped into the final test case package has paths switched to be relative (in particular, the replay script and manifest?)

I guess the wrinkle in this would be if `pre_process()` creates any files that happen to include paths.... not sure I see any good way around that, so perhaps this is the best solution.

siliconcompiler/core.py
nmoroze: Ah, I originally read this wrong. I thought it was defaulting to always dumping two copies of the same manifest, just realized it's copying out an existing one to avoid overwriting.

nmoroze (approved):
Looks good, thanks for answering all my questions/making the changes. My latest comment RE: `self.__relative_paths` was mostly just thinking aloud. My conclusion is I can't think of a better solution, so I'm happy with merging this!

PR #1544, opened 2023-05-10
Show "node" subfields for unset schema params

nmoroze:
This PR addresses the fact that our change to storing all value-specific data underneath the "node" dict makes the subfields not appear for unset parameters in a manifest. I fixed this by adding a `'default', 'default'` entry that is always empty and is used to initialize the node dict whenever it gets populated, mimicking the pattern that gets used for schema parameter keypaths themselves.

I was originally planning to have `'global', 'global'` be always filled in by default, but I realized this mucks with some of the logic as it loses the distinction between required and optional pernode parameters (required ones should not have a `'global', 'global'` entry).

### DIFF
--- a/docs/_ext/schemagen.py
+++ b/docs/_ext/schemagen.py
@@ -43,7 +43,8 @@ class SchemaGen(SphinxDirective):
             if 'unit' in schema:
                 entries.append([strong('Unit'), para(schema['unit'])])
 
-            entries.extend([[strong('Default Value'), para(schema['defvalue'])],
+            defvalue = schema['node']['default']['default']['value']
+            entries.extend([[strong('Default Value'), para(defvalue)],
                             [strong('CLI Switch'), code(schema['switch'])]])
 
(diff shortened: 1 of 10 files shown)


nmoroze:
Github is currently reporting service degradation for Actions, will re-run once it's restored: https://www.githubstatus.com/

nmoroze:
@gadfort I ended up implementing what we discussed offline where the `defvalue` field gets replaced by the `value` field under `default, default`. This change actually felt pretty natural during implementation - it simplifies the code in a few places.

Although this is not what we discussed, I decided not to provide a mechanism to override the default value (but I did add a mechanism to retrieve it). Even though this technically regresses on functionality, I still think it feels a little iffy to do it in the first place, and it might be okay to not allow it unless we come up with a compelling reason to make it possible.

gadfort:
Please update docs and resolve merge conflicts

gadfort (approved):
Lgtm

PR #1545, opened 2023-05-10
Allow integer index for unset()

nmoroze:
This PR fixes a bug where `unset()` would act as a silent no-op if an integer index was provided, which is inconsistent with how the `get/set/add` functions behave.

I also added additional unit tests for `unset()`, and moved the existing test from the core tests to the schema tests, since I think moving towards testing base `Schema` functionality within the schema tests will help with organization.


gadfort (changes requested):
Remove the comments and newline in the tests.

gadfort (approved):

PR #1548, opened 2023-05-12
trigger daily run if _tools.json changes

gadfort:
When tools change it would be nice to auto trigger the daily tests so we can see any failures.

Tested on: https://github.com/siliconcompiler/sc_private/actions/runs/4959488716 and https://github.com/siliconcompiler/sc_private/actions/runs/4959691379


nmoroze (approved):
LGTM

PR #1549, opened 2023-05-12
add flags to control antenna repair and checking in routing task

gadfort:
allow for bypassing antenna repair steps when they are not needed


volkamir (approved):
Reviewed changes and also ran with ant_check/ant_repair set to false to confirm it is progressing.

PR #1554, opened 2023-05-15
update klayout driver

gadfort:
Fixes:
- klayout driver was not accounting for logiclib lefs (replying on the technology file to have that hard coded in)
- allows the driver to build a new lyt file from the schema

Removes the old hardcoded paths from the lyt files
Changes the layout layermap from def/gds to def/klayout, so we can use the def/gds key pair for the actual layer map that klayout supports as of 0.27.


nmoroze (approved):
Looks good to me, nice cleanup/fixes - thanks!

PR #1558, opened 2023-05-16
Clean up Verilator driver/tasks (SC-345, SC-346)

nmoroze:
- Remove parse task: redundant with Surelog and is unused
- Allow both compile and lint to read input from flow or parameters
- Update docs
- Update lintflow to run Verilator directly on input files
- Add test of lintflow
- Delete redundant Verilator test


siliconcompiler/tools/verilator/compile.py
```diff
@@ -25,12 +25,22 @@ def setup(chip):
...
 
 
+def runtime_options(chip):
+    step = chip.get('arg', 'step')
+    index = chip.get('arg', 'index')
+
+    cmdlist = runtime_options_tool(chip)
+
+    for value in chip.find_files('input', 'hll', 'c', step=step, index=index):
+        cmdlist.append(value)
```
gadfort, line 39: should you add input,hll,c as a required field for this?

tests/flows/test_lintflow.py
```diff
@@ -0,0 +1,39 @@
...
+    v_src = os.path.join(scroot, 'tests', 'data', 'heartbeat.v')
+    chip.input(v_src)
+
+    flow = 'lint'
+    chip.node(flow, 'import', parse)
+    chip.node(flow, 'lint', lint)
+    chip.edge(flow, 'import', 'lint')
+    chip.set('option', 'flow', flow)
+
+    chip.run()
```
gadfort, line 39: this test is specific for verilator, so we should pull it out into the tools tests directory

gadfort (changes requested):
please move the tests into their appropriate categories.

siliconcompiler/tools/verilator/compile.py
nmoroze: good catch - it's not required to provide a C/C++ file to Verilator in this mode but I'll add the "if chip.valid(...) then require" pattern since that'll check that the filepath is valid if it's provided.

siliconcompiler/tools/verilator/lint.py
```diff
@@ -16,9 +14,6 @@ def setup(chip):
     step = chip.get('arg', 'step')
     index = chip.get('arg', 'index')
     task = chip._get_task(step, index)
-    design = chip.top()
 
-    chip.add('tool', tool, 'task', task, 'option', ['--lint-only', '--debug'],
-             step=step, index=index)
-    chip.set('tool', tool, 'task', task, 'input', f'{design}.v',
+    chip.add('tool', tool, 'task', task, 'option', ['--lint-only'],
```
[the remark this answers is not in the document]
nmoroze: seems like core.py will fall back to tool driver `runtime_options` if unspecified. If you plan to change that behavior and/or you just think this should be more consistent with compile.py I'll fix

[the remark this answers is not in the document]
gadfort: the fall back is not guaranteed to stay around

nmoroze:
Addressed feedback! I also snuck in another feature (support for Verilator config files via [tool, task, file]... these are helpful for waiving lint violations in 3rd party code)

gadfort (approved):
Lgtm

PR #1559, opened 2023-05-17
Added TL-Verilog frontend (Sandpiper) support for SC

shariethernet:
This commit adds support for TL-Verilog as a front-end, in Silicon Compiler! 

Hello Siliconcompiler team,

This commit adds support for TL-Verilog front-end via integration of Sandpiper, a TL-Verilog Compiler in Silicon Compiler.  This includes code changes as well as documentation updates. 

Thank you for taking the time to review this pull request. I look forward to your valuable input and working together to enhance Silicon Compiler.

Best regards,
Shrihari

### DIFF
--- a/docs/user_guide/tutorials/hw_frontends.rst
+++ b/docs/user_guide/tutorials/hw_frontends.rst
@@ -12,6 +12,8 @@ Here are some tutorials of a few different types of high-level hardware design f
 * :ref:`C HLS Frontend (Bambu) <c hls frontend>`
   
 * :ref:`bluespec frontend`
+
+* :ref:`TL-Verilog frontend (Sandpiper)`
   
 
 .. include:: migen.rst
@@ -21,3 +23,5 @@ Here are some tutorials of a few different types of high-level hardware design f
 .. include:: bambu.rst
 	     
 .. include:: bluespec.rst
+
+.. include:: tlverilog.rst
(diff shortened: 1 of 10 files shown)


gadfort:
@shariethernet Thank you for your contribution. At present we will not be able to accept this, due to the use of the sandpaper-saas microservice.
If the TL-Verilog Compiler were opensource (and freely licensed) and we could compile it into our workflow then we can revisit this, but from my poking around it doesn't appear to be true FOSS.
Please let me know if I'm missing something.

shariethernet:
@gadfort . Thanks for the review. TL-Verilog compiler is fully open source and is available at https://github.com/TL-X-org/tlv-comp . The reason for using sandpiper-saas is that it is the latest version, and would not require any re-compilation at the user end. However, I understand you intend to compile and install it locally in your workflow. Let me know support such that it has a config switch that enables either remote (SaaS) or local compilation would work well for silicon compiler.

gadfort:
@shariethernet Thanks for pointing me to the tlv-comp archive. It doesn't look to be under current development and mostly dormant (9 years since the copyright, 6 years since the only few commits + some from 2 years ago), so I don't see much value in incorporating into the SC. This will just cause problems later when we update OS's and it won't compile and prevent us from updating.  Tool that are under continuous development and currently supported by their maintainers as a general rule is a requirement, exceptions can be made, but not in this case.
I also have not been able to find any major projects that support TL-Verilog, so it's not clear to me this is a viable frontend for us to be supporting.

PR #1561, opened 2023-05-17
Fix use of find_files in setup()

gadfort:
This fixes the use of `find_files()` in `setup()` for the tools and moves it to the `pre_process()` where it belongs.

### DIFF
--- a/siliconcompiler/tools/surelog/surelog.py
+++ b/siliconcompiler/tools/surelog/surelog.py
@@ -56,7 +56,8 @@ def setup(chip):
     # system path, set the path to the bundled copy in the schema.
     if shutil.which('surelog') is None:
         surelog_path = os.path.join(os.path.dirname(__file__), 'bin')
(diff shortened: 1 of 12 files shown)


siliconcompiler/tools/surelog/surelog.py
```diff
@@ -56,7 +56,8 @@ def setup(chip):
     # system path, set the path to the bundled copy in the schema.
     if shutil.which('surelog') is None:
         surelog_path = os.path.join(os.path.dirname(__file__), 'bin')
-        chip.set('tool', tool, 'path', surelog_path, clobber=False)
+        if os.path.exists(os.path.join(surelog_path, 'surelog')):
```
nmoroze, line 59: The path is `surelog.exe` on Windows, so you'll want to check for that as well

nmoroze (changes requested):

nmoroze (approved):

PR #1563, opened 2023-05-17
make option,credentials a single file instead a list

gadfort:
For 'option,credentials, a list of files doesn't make sense and this makes it a single item instead


nmoroze (approved):
Looks good to me. Frankly I expected things to break when introducing a `file` scalar, but maybe this parameter doesn't really get used in code paths that make assumptions about files always being lists...

PR #1564, opened 2023-05-17
Add support for multiple voltages in constraints

gadfort:
This adds a 'pin' parameter in the constraints,timing,..,voltage to support multiple power/voltage domains


bnoorani (approved):
Thank you!

nmoroze (approved):

PR #1567, opened 2023-05-18
Support defines and params with Verilator

nmoroze:


siliconcompiler/tools/verilator/verilator.py
```diff
@@ -109,6 +113,8 @@ def runtime_options(chip):
             cmdlist.append('-I' + value)
         for value in chip.find_files('option', 'cmdfile'):
             cmdlist.append('-f ' + value)
+        for value in chip.get('option', 'define'):
+            cmdlist.append('-D' + value)
```
gadfort, line 117: Can you switch this to f'-D{value}' instead?

gadfort (commented):
Please fix the string formatting, but otherwise looks great

siliconcompiler/tools/verilator/verilator.py
nmoroze: Yeah. I was being consistent with the legacy code in that block, but I'll just fix it all

PR #1570, opened 2023-05-19
klayout stream operations task

gadfort:
This PR adds an operations task to klayout to allow for unit operations on the GDS/OAS files, which may be required for a tapeout to meet the foundries submission requirements.

### DIFF
--- a/siliconcompiler/tools/klayout/klayout_operations.py
+++ b/siliconcompiler/tools/klayout/klayout_operations.py
@@ -177,9 +177,6 @@ def make_property_text(layout, property_layer, property_name, destination_layer)
 
 
 def parse_operations(schema, base_layout, steps):
-    sc_step = schema.get('arg', 'step')
-    sc_index = schema.get('arg', 'index')
-
     for step in steps:
         step = step.split(":")
         step_name = step[0]


siliconcompiler/tools/klayout/klayout_operations.py
```diff
@@ -0,0 +1,263 @@
...
+    new_cell = base_layout.create_cell(other_layout_top.name)
+    new_cell.copy_tree(other_layout_top)
+
+    cell_inst = pya.CellInstArray(new_cell.cell_index(), pya.Trans())
+    top_cell.insert(cell_inst)
+
+    return base_layout
+
+
+def add_layout_top_top(base_layout, new_top_cell_name):
```
nmoroze, line 85: Name seems a bit funny, can't tell if it's a typo or if it's intentional :) As far as I can tell this is creating a new top-level cell above the base layout, so not sure if you meant `add_layout_to_top`.

siliconcompiler/tools/klayout/klayout_operations.py
```diff
@@ -0,0 +1,263 @@
...
+        step_name = step[0]
+        step_args = ":".join(step[1:])
+        args_key = step_args.split(',')
+
+        if (step_name == "merge" or step_name == "add"):
+            files = []
+            if len(args_key) > 1:
+                if 'file' not in schema.get(*args_key, field='type'):
+                    raise ValueError(f'{step_name} requires {args_key} be a file type')
+                files = schema.get(*args_key, step=sc_step, index=sc_index)
```
nmoroze, line 181: This might blow up for non pernode values

siliconcompiler/tools/klayout/operations.py
```diff
@@ -0,0 +1,167 @@
...
+        ['dummy_ANDX2=ANDX2', 'dummy_NANDX2=NANDX2'])
+
+    To add new top cell:
+
+    >>> chip.add('tool', 'klayout, 'task', 'operations', 'var', 'operations', \\
+        'add_top:tool,klayout,task,operations,var,new_name')
+    >>> chip.set('tool', 'klayout, 'task', 'operations', 'var', 'new_name', \\
+        'chip_top')
+
+    To add write out a new file:
```
nmoroze, line 75: Typo: omit "add"

nmoroze (changes requested):
This looks good, thanks! Main things I'm wondering about:

- For the operations that take in regular strings/lists of strings, is there value in having them be read out of keypaths vs. just embedding the thing directly in the command? e.g. just do `rename:chip_top` instead of `rename:tool,klayout,task,operations,var,new_name`. I get why it's useful to do for files though, so I suppose there's value in the consistency.
- Related, I think doing `schema.get(..., step=sc_step, index=sc_index)` for grabbing values could cause problems if someone supplies a keypath that doesn't allow per-node values. If the intent is to just do args that are klayout task vars, maybe we could add some validation for that (or even just insert the prefix implicitly, i.e. `rename:new_name` gets expanded to searching under ['tool', 'klayout', 'task', 'operations', 'var', 'new_name']`

gadfort:
> This looks good, thanks! Main things I'm wondering about:
> 
> * For the operations that take in regular strings/lists of strings, is there value in having them be read out of keypaths vs. just embedding the thing directly in the command? e.g. just do `rename:chip_top` instead of `rename:tool,klayout,task,operations,var,new_name`. I get why it's useful to do for files though, so I suppose there's value in the consistency.
> * Related, I think doing `schema.get(..., step=sc_step, index=sc_index)` for grabbing values could cause problems if someone supplies a keypath that doesn't allow per-node values. If the intent is to just do args that are klayout task vars, maybe we could add some validation for that (or even just insert the prefix implicitly, i.e. `rename:new_name` gets expanded to searching under ['tool', 'klayout', 'task', 'operations', 'var', 'new_name']`

I can add a check for pernode, I was attempting to be somewhat consistent in using the keypaths for the values (they are also easier to change later since we don't have a good way to change just part of a list.

nmoroze (approved):
LGTM, thanks!

Issue #1231, opened 2023-01-25
Refactor logger to be static/globally accessible


nmoroze:
Currently the application logger is owned by the Chip object. This makes it hard to directly log messages from code in the Schema object. It will become even more problematic once we switch over PDK/library/flow setup to happen in setup modules that don't take in a Chip object.

We should look into supporting a single static logger that can be accessed by any module.

nmoroze:
@gadfort suggestion: have chip object pass down logger into Schema object (optional parameter, where Schema objects that exist outside of chip context could fallback to their own logger). We could also have module `setup()` methods accept a logger as an argument if they wish to log things.

PR #1572, opened 2023-05-19
Cleanup logger handling and pass logger to schema object

gadfort:
This PR ensures the schema obj has access to a logger and avoids the need to manually deinit loggers when `run()` is engaged.

This closes: #1231

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -23,7 +23,6 @@ import math
 import pandas
 import pkgutil
 import graphviz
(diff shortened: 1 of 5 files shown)


nmoroze (approved):
Looks good, thanks for fixing this up!

The one thing I don't feel great about is having to push the logger down into static functions, but this is still an improvement and I'm not sure if there's a better solution. Would it be possible for static functions to grab an instance of the logger by ID?

PR #1573, opened 2023-05-19
Fix permission child and error message in load_target

gadfort:
Fixes:
When the OS permissions do not allow access to the memory / child processes this ensure we still function
Adds useful error when load_target fails to get module

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -637,7 +637,9 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                     modules.append(mod)
 
             if len(modules) == 0:
-                self.error(f'Could not find target {module}', fatal=True)
+                # attempt to load just the module again with raising the actual error,
+                # this will ensure a useful error message is generated to the user
+                self._load_module(module, raise_error=True)
         else:
             modules = [module]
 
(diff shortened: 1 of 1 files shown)


siliconcompiler/core.py
```diff
@@ -637,7 +637,9 @@ def load_target(self, module, **kwargs):
                     modules.append(mod)
 
             if len(modules) == 0:
-                self.error(f'Could not find target {module}', fatal=True)
+                # attempt to load just the module again with raising the actual error,
+                # this will ensure a useful error message is generated to the user
+                self._load_module(module, raise_error=True)
```
nmoroze, line 642: Seems more useful for the case where the module load fails due to an error in the file, but doesn't it make error reporting a little worse if the file is not found at all? This also wouldn't help if there happened to be an error in one of our targets, since it only tries "module" (which I suppose we'd ideally catch, but still) 

An alternate proposal would be to refactor `self._load_module()` to first check if the module exists using 
`importlib.util.find_spec`, return `None` if it does not, and then try to actually import the module while unconditionally raising an exception. 

This does introduce a behavioral change where `load_target()` would no longer fall-back to a built-in target if there's an error in the first one it tries, but that may be desirable... I assume if there's a match on `module` using that target would be the user's intent, and they'd be fine with it erroring out without falling back.

gadfort: lets table this for now, I agree it's not the right solution

nmoroze (approved):

PR #1576, opened 2023-05-19
Update Verilator to latest release version

nmoroze:


gadfort (approved):
If we have reasonable tests maybe we can enable the auto updating for verilator?

PR #1586, opened 2023-05-23
Fix yosys pre-process method to allow macro libs which lack ('asic', 'cells') configs

WRansohoff:
This PR addresses an issue which came up in a parallel investigation, and adds a test to catch it in the future.

Some macro libraries may not set values such as `('asic', 'cells', 'dontuse')`, `('asic', 'cells', 'clkbuf')`, etc.

If no value under the top-level `('asic', 'cells')` keypath is set, then the Yosys pre_process method will crash when it tries to call `chip.get('library', [...], 'asic', 'cells', 'dontuse')`, because the `'cells'` key will not exist in the Schema.


gadfort:
Closed in favor of: https://github.com/siliconcompiler/siliconcompiler/pull/1590

PR #1590, opened 2023-05-25
rework prune to avoid removing information

gadfort:
Reworks prune()
Adds loopback test for gcd to ensure python run and manifest based run both work the same

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1930,7 +1930,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
 
         Args:
             filename (filepath): Output filepath
-            prune (bool): If True, only essential parameters from the
+            prune (bool): If True, only essential fields from the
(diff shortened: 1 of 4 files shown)


siliconcompiler/core.py
```diff
@@ -1930,7 +1930,7 @@ def write_manifest(self, filename, prune=True, abspath=False):
 
         Args:
             filename (filepath): Output filepath
-            prune (bool): If True, essential non-empty parameters from the
+            prune (bool): If True, only essential parameters from the
```
nmoroze, line 1933: Maybe just say "only essential fields" (or just specify that "help" and "example" are removed) now that it's not actually deleting any parameters.

nmoroze (approved):
LGTM. I guess there could be some information loss in removing 'help' in that the help strings added for vars in some of the tool `setup()` methods get lost, but I think that's okay.

PR #1601, opened 2023-05-27
use require parameter to minimize size of testcases copied

gadfort:


gadfort:
> Looks good. I imagine we'll run into things where parameters aren't getting added to require by tool drivers, but that's probably a valuable thing to catch anyways.

yes, that is possible, but I think then the tool driver is not correct.

PR #1608, opened 2023-05-30
add puts messages in openroad export and limit timing model writing t…

gadfort:
…o 1 thread


nmoroze (commented):
LGTM. Is there somewhere where the write_timing_model issue is being tracked (maybe an OR issue)? Might be useful to reference back to it (either in a comment or at least in the PR body)

gadfort:
There is a ticket for it

nmoroze (approved):

PR #1620, opened 2023-06-01
change daily ci trigger to use workflow-and-wait

gadfort:
Also updates verilator as a test for the flow

### DIFF
--- a/.github/workflows/daily_ci.yml
+++ b/.github/workflows/daily_ci.yml
@@ -3,7 +3,6 @@ on:
   schedule:
     - cron: '0 8 * * *'
   repository_dispatch:
-    types: [trigger-daily-workflow]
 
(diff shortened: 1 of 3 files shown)


.github/workflows/daily_ci.yml
```diff
@@ -3,7 +3,6 @@ on:
   schedule:
     - cron: '0 8 * * *'
   repository_dispatch:
```
WRansohoff, line 5: Do you think `repository-dispatch:` is still required for the trigger and wait action, since it uses the `.yml` file path as a target instead of a trigger name?

WRansohoff (approved):
LGTM

.github/workflows/daily_ci.yml
gadfort: I think you are right

PR #1627, opened 2023-06-02
moving request code from sc-ping to client.py

gadfort:
There are no functional changes in this PR, just moving the code related to talking with the server to client.py to be consistent.

### DIFF
--- a/siliconcompiler/client.py
+++ b/siliconcompiler/client.py
@@ -377,3 +377,42 @@ def fetch_results(chip):
 
     # Print a message pointing to the results.
     chip.logger.info(f"Your job results are located in: {os.path.abspath(chip._getworkdir())}")
+
+
+###################################
+def remote_ping(chip):
+    '''
+    Helper method to call check_user on server
+    '''
+    # Create the chip object and generate the request
+    request_url = get_base_url(chip) + '/check_user/'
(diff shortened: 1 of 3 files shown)


siliconcompiler/client.py
```diff
@@ -377,3 +377,42 @@ def fetch_results(chip):
...
+
+###################################
+def remote_ping(chip):
+    '''
+    Helper method to call check_user on server
+    '''
+    # Create the chip object and generate the request
+    request_url = get_base_url(chip) + '/check_user/'
+
+    remote_cfg = chip.status['remote_cfg']
```
WRansohoff, line 390: Might be worth adding a `if 'remote_cfg' in chip.status:` check with an explanatory error message, in case of the unlikely event that someone might try to call this from somewhere besides the `sc-ping` app.

Or, maybe consider making `remote_cfg` a function argument.

WRansohoff (approved):
Looks good, just one suggestion to avoid possible `KeyError`s down the road.

siliconcompiler/client.py
gadfort: Sure, but we really don't have those checks anywhere else. I'll track it with a ticket for now

PR #1630, opened 2023-06-05
refactor client.py

gadfort:
This PR unifies the communications with the server into a single function `__post` to allow for simpler error handling and updates to how we talk between the client and sever.
There appears to be a lot of different ways the server can respond, so we need to fix that in the API to make this even simpler and easier to maintain.

### DIFF
--- a/siliconcompiler/client.py
+++ b/siliconcompiler/client.py
@@ -35,7 +35,7 @@ def get_base_url(chip):
 
 
 ###################################
-def __post(chip, url, post_action, success_action):
+def __post(chip, url, post_action, success_action, error_action=None):
     '''
     Helper function to handle the post request
     '''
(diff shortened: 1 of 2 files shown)


siliconcompiler/client.py
```diff
@@ -34,9 +34,68 @@ def get_base_url(chip):
...
+###################################
+def __build_post_params(chip, add_job_name=True, add_job_hash=True):
+    '''
+    Helper function to build the params for the post request
+    '''
+    # Use authentication if necessary.
+    post_params = {}
+
+    if 'jobhash' in chip.status and add_job_hash:
+        post_params['job_hash'] = chip.status['jobhash']
```
WRansohoff, line 80: Could `job_hash` be a kwarg with default value of `None`? Then this could just be `if job_hash:`, ditto below for the job name/ID.

siliconcompiler/client.py
```diff
@@ -34,9 +34,68 @@ def get_base_url(chip):
     return remote_protocol + remote_host
 
 
+###################################
+def __post(chip, url, post_action, success_action, error_action=None):
```
WRansohoff, line 38: Could the default `error_action` be a method which prints the response text?

WRansohoff (approved):
Overall looks good, just a few small suggestions.

siliconcompiler/client.py
gadfort: Line 68: `chip.error(f'Server responsed with {code}: {msg}', fatal=True)`

PR #1636, opened 2023-06-06
record records as we go through the flow

gadfort:
Changes:
- enables the records for time, tool, and scversion.
- Continues to use `'option', 'track'` to record the user and machine specific information

### DIFF
--- a/docs/development_guide/records.rst
+++ b/docs/development_guide/records.rst
@@ -1,9 +1,13 @@
 Records
 =======
 
-To support hardware provenance, the SiliconCompiler supports automated tracking of a number of execution and place of origin related parameters. Tracking is off by default in the SiliconCompiler, but can be turned on with the :keypath:`option,track` parameter. ::
+To support hardware provenance, the SiliconCompiler supports automated tracking of a number of execution and place of origin related parameters.
+SiliconCompiler will record the SiliconCompiler, tool version and options, and task start and end times by default.
+Additional tracking is off by default in the SiliconCompiler, but can be turned on with the :keypath:`option,track` parameter. ::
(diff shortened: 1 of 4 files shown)


nmoroze (changes requested):
Changes look good, could you also update the ['option', 'track'] documentation and this page: https://docs.siliconcompiler.com/en/stable/development_guide/records.html to reflect the fact that track only controls a subset of parameters now?

docs/development_guide/records.rst
```diff
@@ -1,10 +1,14 @@
 Records
 =======
 
-To support hardware provenance, the SiliconCompiler supports automated tracking of a number of execution and place of origin related parameters. Tracking is off by default in the SiliconCompiler, but can be turned on with the :keypath:`option,track` parameter. ::
+To support hardware provenance, the SiliconCompiler supports automated tracking of a number of execution and place of origin related parameters.
```
nmoroze, line 4: "the SiliconCompiler" -> "SiliconCompiler" is more consistent with our other usage (also check the other instance in this paragraph, "...off by default in the SiliconCompiler...")

bit of a nitpick, and I know it was already like this, but might be good to fix while you're updating this page anyways

docs/development_guide/records.rst
```diff
@@ -1,10 +1,14 @@
...
+SiliconCompiler will record the SiliconCompiler, tool version and options, and task start and end times by default.
```
nmoroze, line 5: "the SiliconCompiler version"

nmoroze (approved):
Looks good, just a few minor comments on wording in docs.

PR #1646, opened 2023-06-08
reorganize server files into module

gadfort:
No functional changes, just moving files around


docs/_ext/clientservergen.py
```diff
@@ -60,7 +62,11 @@ def run(self):
         new_doc = []
 
         # Get the JSONSchema API files to include.
-        api_schemas = glob.iglob('server_schema/*.json')
+        api_location = os.path.join(
+            os.path.dirname(siliconcompiler.__file__),
+            'remote',
+            'server_schema')
+        api_schemas = glob.iglob(f'{api_location}/*.json')
```
WRansohoff, line 69: These JSONSchemas aren't actually consumed by the dev server right now, but it probably is a good idea to move them out of the docs directory as we look to improve the dev server's feature parity.

WRansohoff (approved):

docs/_ext/clientservergen.py
gadfort: I was hoping to use the same schema checker thing on the server to ensure parity (future PR).

PR #1660, opened 2023-06-13
add hierarchical synthesis in yosys 

gadfort:
Adds:
- this is an initial implementation of hierarchical synthesis for yosys.

Idea:
- break the synthesis step and look for modules below a specific size, those can be flattened and the rest kept and then try again.


gadfort:
@maliberty FYI, adding this so it's available as a starting point

nmoroze (approved):
LGTM

PR #1661, opened 2023-06-13
add enablement for rtlmp in openroad

gadfort:
Parameter selection is based on survey of ORFS designs and what is enabled there.


gadfort:
@maliberty, I will need help to fill in the schelp strings

PR #1667, opened 2023-06-14
Remove '-design' parameter from scheduler.py

WRansohoff:
The design value is already in the Schema, and `chip.top()` points to the entrypoint parameter.


gadfort (approved):
Lgtm

PR #1686, opened 2023-06-19
make map_adders a pure true/false parameter in yosys

gadfort:


siliconcompiler/tools/yosys/syn_asic.py
```diff
@@ -90,6 +84,16 @@ def setup_asic(chip):
...
                  step=step, index=index)
 
+    chip.set('tool', tool, 'task', task, 'var', 'map_adders',
+             "true" if chip.valid('library', mainlib, 'option', 'file', 'yosys_addermap') else
+             "false",
+             step=step, index=index, clobber=False)
+    if chip.get('tool', tool, 'task', task, 'var', 'map_adders', step=step, index=index)[0] == \
+       "true":
+        chip.add('tool', tool, 'task', task, 'require',
+                 ",".join(['tool', tool, 'task', task, 'var', option]),
```
nmoroze, line 94: Is `option` the correct thing to use here? Seems suspect to me, although not 100% sure what the intent is.

siliconcompiler/tools/yosys/syn_asic.py
```diff
@@ -90,6 +84,16 @@ def setup_asic(chip):
                  ",".join(['library', mainlib, 'option', 'var', 'yosys_buffer_output']),
                  step=step, index=index)
 
+    chip.set('tool', tool, 'task', task, 'var', 'map_adders',
+             "true" if chip.valid('library', mainlib, 'option', 'file', 'yosys_addermap') else
```
nmoroze, line 88: You might also want to check that the addermap got set for the particular step/index beyond just the keypath being valid, otherwise I could see this behaving funny if it the adder map is provided for one Yosys node but not another.

siliconcompiler/tools/yosys/syn_asic.py
```diff
@@ -84,14 +84,19 @@ def setup_asic(chip):
                  ",".join(['library', mainlib, 'option', 'var', 'yosys_buffer_output']),
                  step=step, index=index)
 
+    library_has_addermap = \
+        chip.valid('library', mainlib, 'option', 'file', 'yosys_addermap') and \
+        chip.get('library', mainlib, 'option', 'file', 'yosys_addermap')
```
nmoroze, line 89: Should have `step=step, index=index)` I think

nmoroze (approved):
Looks good once that last comment is resolved. Thanks!

siliconcompiler/tools/yosys/syn_asic.py
gadfort: pernode is never on this parameter (I initially thought so too)

nmoroze: Ah, got it. Good to go then!

PR #1694, opened 2023-06-20
Add pathlib support to core API

nmoroze:
This PR makes a few small updates to ensure that users can provide file paths as `pathlib.Path` objects for all our core API methods.

`set()`/`add()` are taken care of in the schema value normalization method, and then I added `str()` casts in a few places as needed. The rest of the methods that I audited expand file paths received by the user using `os.path.abspath()`, which already accepts a `pathlib.Path`.

Note that paths are always normalized to strings, so file paths read out of the schema or returned from other SC methods will remain as strings. This just makes user input more flexible.

### DIFF
--- a/siliconcompiler/schema/schema_obj.py
+++ b/siliconcompiler/schema/schema_obj.py
@@ -538,7 +538,7 @@ class Schema:
                 raise TypeError(error_msg)
 
         if sc_type in ('file', 'dir'):
-            if isinstance(value, str) or isinstance(value, pathlib.Path):
+            if isinstance(value, (str, pathlib.Path)):
                 return str(value)
             else:
                 raise TypeError(error_msg)
--- a/siliconcompiler/schema/schema_obj.py
+++ b/siliconcompiler/schema/schema_obj.py
@@ -11,6 +11,7 @@ import json
 import logging
 import os
 import re
+import pathlib
 
 try:
     import yaml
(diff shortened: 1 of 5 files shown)


tests/schema/test_schema.py
```diff
@@ -40,3 +42,15 @@ def test_add_keypath_error():
...
+def test_pathlib():
+    schema = Schema()
+
+    file_path = pathlib.Path('path/to/file.txt')
+    schema.set('option', 'file', 'test', file_path)
+    assert schema.get('option', 'file', 'test') == [str(file_path)]
+
+    dir_path = pathlib.Path('a/directory/')
+    schema.set('option', 'dir', 'test', dir_path)
+    assert schema.get('option', 'dir', 'test') == [str(dir_path)]
```
gadfort, line 56: add test for reading the schema with a pathlib as well so we lock that in.

siliconcompiler/schema/schema_obj.py
```diff
@@ -528,12 +531,18 @@ def _normalize_value(value, sc_type, error_msg, allowed_values):
...
 
-        if sc_type in ('str', 'file', 'dir'):
+        if sc_type == 'str':
             if isinstance(value, str):
                 return value
             else:
                 raise TypeError(error_msg)
 
+        if sc_type in ('file', 'dir'):
+            if isinstance(value, str) or isinstance(value, pathlib.Path):
```
gadfort, line 541: you can do `isinstance(value, (str, pathlib.Path))`

gadfort (approved):
I think it would be cleaner with the `isinstance` tuple, but I'll approve

siliconcompiler/schema/schema_obj.py
nmoroze: Thanks, didn't know this was a thing but I like it! Changed over.

PR #1711, opened 2023-06-27
Add Verilog lint CI job

nmoroze:
This PR adds linting to our Verilog examples, using [Verible](https://github.com/chipsalliance/verible). The workflow is broken up into two steps:

- Since the Verible linter doesn't check formatting issues, the first step runs the Verible auto-formatter in-place, and reports an error if any file is modified (meaning that it doesn't meet the required format)
- The Verible linter is then run on each of the files

In order to ease fixing formatting issues, the formatting command sits in a stand-alone script that can be run locally (`./github/workflows/bin/format_verilog.sh`). This script is also responsible for configuring which files get checked.

As a general policy, linting is run on all Verilog that we wrote, but not on third-party examples. The one exception is GCD, since it's a pretty prolific example. Note that I inserted a few in-line lint waivers in GCD to avoid actually modifying any sort of names or logic - the changes to that file in this PR otherwise solely come from the formatting.

I waived a few of the Verible defaults in `verible.rules`. The exclusions are documented in that file, I'm open to making some of these things stricter. 

~Finally, note that this PR changes the event triggers for the lint workflow. Reviewing our other workflows, it seems like everything has a `push` trigger only, so I updated the lint workflow to make it consistent. If we wanted, we could switch everything over to `pull_request`, but I think that might have other consequences (e.g. that will actually run workflows on an imaginary merge commit of the PR branch into the target branch).~


nmoroze:
@gadfort I've finished updating the PR based on offline feedback, should be ready for another look!

setup/_tools.json
```diff
@@ -94,5 +94,11 @@
     "version": "6.9.11",
     "auto-update": false,
     "docker-skip": true
+  },
+  "verible": {
+    "version": "v0.0-3303-gd87f2420",
+    "git-url": "https://github.com/chipsalliance/verible.git",
+    "auto-update": false,
+    "docker-skip": true
```
gadfort, line 102: remove, docker-skip

PR #1715, opened 2023-06-29
add client timeout and fix some messaging when communicating with the server

gadfort:


gadfort:
@WRansohoff just checking in on this

WRansohoff (approved):
Looks good to me - sorry for the delay, it was a busy week getting back from break.

My only note is that this will prevent the server from gracefully recovering after a long-term network failure, e.g. going through a train tunnel or shutting your laptop to drive home. Currently, that use case is supported and the client picks back up where it left off. With a too-short timeout, users will no longer be able to return to jobs after long times away, at least until we implement async cancel/resume functionality.

gadfort:
We'll get the async in soon, but without the timeout, if the server messes up, SC gets stuck forever.

PR #1724, opened 2023-07-05
fix OpenROAD WNS units to honor proper units

gadfort:


troibe (approved):
Ran through this on the simple heartbeat example with a debugger.
Verified that the value is the same before and after as `setupslack` is currently set according to the value of `or_units['time']`.
Just wondering whether there would be any reasonable scenario where `setupslack` shoudl be set to something different than `or_units['time']`?

gadfort:
when we call chip.get('metric', 'setupslack') it will be returned with the SC units (ns usually), but the design units could be ps which was causing the wns to be 1/1000 the correct value.

PR #1728, opened 2023-07-06
fix streamlit import to account for older versions

gadfort:


troibe:
Looks good!
Maybe a small comment as to why/which streamlit version this is necessary for would be helpful.

troibe (approved):

PR #1731, opened 2023-07-06
revert OpenROAD to last known functional state

gadfort:
OpenROAD has slowed down a good deal on placement and estimation of parasitics, this reverts us to a commit that predates this issue until they are able to fix it.


troibe (approved):
looks good

PR #1736, opened 2023-07-07
Add support for libext, Verilog standard to Icarus

nmoroze:
This PR adds some additional functionality to the Icarus tool driver:

- Support for ['option', 'libext']
- The ability to select which Verilog language standard to use
- Explicitly provide the top-level module, which is necessary to ensure that ['option', 'entrypoint'] works correctly


gadfort (approved):
Lgtm

PR #1725, opened 2023-07-05
Adding sc-dashboard

GabrielAguirre1:
Adding Home.py and test_reports.py

PR #1748, opened 2023-07-11
remove streamlit from Python 3.6

gadfort:
Also disable currently breaking test, which needs to be fixed later

### DIFF
--- a/siliconcompiler/report/streamlit_report.py
+++ b/siliconcompiler/report/streamlit_report.py
@@ -5,12 +5,10 @@ import json
 
 try:
     from streamlit.web import bootstrap
+    from streamlit import config as _config
+    _has_streamlit = True
 except ModuleNotFoundError:
-    # In streamlit 1.10.1 bootstrap was moved to streamlit.web
-    # This is the only version available for Python 3.6
-    from streamlit import bootstrap
-
-from streamlit import config as _config
+    _has_streamlit = False
 
 import multiprocessing
 import subprocess
@@ -22,6 +20,9 @@ class Dashboard():
     __port = 8501
 
     def __init__(self, chip, port=None):
+        if not _has_streamlit:
+            raise RuntimeError("Streamlit is not available")
+
         if not port:
             port = Dashboard.__port
 
--- a/siliconcompiler/report/streamlit_report.py
+++ b/siliconcompiler/report/streamlit_report.py
@@ -2,13 +2,11 @@ import os
 import time
 import tempfile
 import json
+import sys
 
-try:
+if sys.version_info >= (3, 7, 0):
     from streamlit.web import bootstrap
     from streamlit import config as _config
-    _has_streamlit = True
-except ModuleNotFoundError:
-    _has_streamlit = False
 
 import multiprocessing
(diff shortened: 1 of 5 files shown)


troibe:
The two changes are unrelated or am I mising something?
The breaking up part looks good to me.

gadfort:
@troibe no, both are for getting the daily CI back to passing

gadfort:
3.6 was breaking, 3.10 was sometimes breaking. I opened a ticket to track the failing test to get it back to functional again.

siliconcompiler/report/streamlit_report.py
```diff
@@ -22,6 +20,9 @@ class Dashboard():
     __port = 8501
 
     def __init__(self, chip, port=None):
+        if not _has_streamlit:
+            raise RuntimeError("Streamlit is not available")
+
```
troibe, line 25: If we the user has the wrong version we could do something like this.
```suggestion
    if sys.version_info[0] < (3,7,0):
      raise RuntimeError("Dashboard is not available in 3.6")
```

siliconcompiler/report/streamlit_report.py
```diff
@@ -5,12 +5,10 @@
...
     from streamlit.web import bootstrap
+    from streamlit import config as _config
+    _has_streamlit = True
 except ModuleNotFoundError:
-    # In streamlit 1.10.1 bootstrap was moved to streamlit.web
-    # This is the only version available for Python 3.6
-    from streamlit import bootstrap
-
-from streamlit import config as _config
+    _has_streamlit = False
```
troibe, line 11:
```suggestion
```

siliconcompiler/report/streamlit_report.py
```diff
@@ -5,12 +5,10 @@
 
 try:
     from streamlit.web import bootstrap
+    from streamlit import config as _config
+    _has_streamlit = True
 except ModuleNotFoundError:
```
troibe, line 10: Let's just stick with python's default package management then.
```suggestion
from streamlit.web import bootstrap
from streamlit import config as _config
```

troibe (changes requested):
The error would pop up in both cases: streamlit is actually missing + dashboard is not available
See here:
https://github.com/siliconcompiler/siliconcompiler/pull/1725#discussion_r1254490184

See suggestions

troibe:
> 3.6 was breaking, 3.10 was sometimes breaking. I opened a ticket to track the failing test to get it back to functional again.

The 3.10 is also related to streamlit?

siliconcompiler/report/streamlit_report.py
gadfort: this will not work, as a 3.6 package will attempt to import this and fail.

troibe: Sure then an if statement for the version around it?
I would prefer not to overwrite the "ModuleNotFoundError" because that is very helpful.

gadfort: Okay

troibe (approved):
Replaced the version check now

PR #1749, opened 2023-07-11
Code refactoring

GabrielAguirre1:
I refactored the code based on Noah's suggestions in the last PR which include:

- creating _get_flowgraph_paths in utils.py to fit in with the summary table and get_flowgraph_paths in reports.py
- making show_dataframe_and_parameter_selection more readable
- Also let's compressed files readable when selected in the node information section

### DIFF
--- a/siliconcompiler/report/streamlit_viewer.py
+++ b/siliconcompiler/report/streamlit_viewer.py
@@ -586,6 +586,7 @@ def show_title_and_runs(title_col_width=0.7):
 
     return new_chip
 
+
 new_chip = show_title_and_runs()
 
 # gathering data
--- a/siliconcompiler/report/streamlit_viewer.py
+++ b/siliconcompiler/report/streamlit_viewer.py
@@ -2,14 +2,16 @@ import streamlit
 from streamlit_agraph import agraph, Node, Edge, Config
 from streamlit_tree_select import tree_select
 from streamlit_toggle import st_toggle_switch
+import streamlit_javascript
 from PIL import Image
 from pathlib import Path
 import os
 import argparse
 import json
 import pandas
+import gzip
 from siliconcompiler.report import report
-from siliconcompiler import Chip, TaskStatus
+from siliconcompiler import Chip, TaskStatus, utils
 from siliconcompiler import __version__ as sc_version
 
 # for flowgraph
@@ -21,6 +23,7 @@ FAILURE_COLOR = '#FF4E00'  # red
 INACTIVE_TOGGLE_COLOR = "#D3D3D3"
 ACTIVE_TOGGLE_COLOR = "#11567f"
 TRACK_TOGGLE_COLOR = "#29B5E8"
+
 
 sc_logo_path = \
     os.path.join(os.path.dirname(__file__), '..', 'data', 'logo.png')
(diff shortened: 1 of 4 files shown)


siliconcompiler/report/streamlit_viewer.py
```diff
@@ -211,15 +218,18 @@ def show_file_preview(header_col_width=0.89):
...
     else:
         try:
-            with open(path, 'r') as file:
-                content = file.read()
+            if compressed_file_extension == '.gz':
+                file = gzip.open(path, 'rt')
+            else:
+                file = open(path, 'r')
+            content = file.read()
+            file.close()
```
gadfort, line 226: `file` is an predefined object in Python, could you rename it to something else? maybe `fid` (not great but avoids the naming collision)?

siliconcompiler/report/utils.py
```diff
@@ -76,3 +79,28 @@ def _collect_data(chip, flow, steplist):
             metrics_unit[metric] = metric_unit if metric_unit else ''
 
     return nodes, errors, metrics, metrics_unit, metrics_to_show, reports
+
+
+def _get_flowgraph_path(chip, flow, steplist, summary_table=False):
```
gadfort, line 84: rename `summary_table` to something like `only_include_successful` since that is its real usecase

gadfort (changes requested):
a couple of small changes needed.

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -195,7 +198,11 @@ def show_file_preview(header_col_width=0.89):
             is given to the download button.
     """
     path = streamlit.session_state['selected'][0]
-    file_name, file_extension = os.path.splitext(path)
+    # This file extension may be '.gz', if it is, it is compressed.
```
troibe, line 201: Could you either add a test case for the compressed file viewer or a screenshot of it in action to this pull request?

troibe:
Looks good to me.
In general I would probably split up the refactoring and the new compressed file viewer into two pull requests.
Since one is a feature and the other just cleanup.
That's just a matter of style and Peter/Noah might have different takes on this.

gadfort:
@troibe I agree, but said it was okay since he already had this branch available.

siliconcompiler/report/utils.py
```diff
@@ -28,7 +28,8 @@ def _collect_data(chip, flow, steplist):
     # Gather data and determine which metrics to show
     # We show a metric if:
     # - it is not in ['option', 'metricoff'] -AND-
-    # - at least one step in the steplist has a non-zero weight for the metric -OR -
+    # - at least one step in the steplist has a non-zero weight for the metric
+    #  -OR -
```
nmoroze, line 32: Small thing, but I think the line length changes in this file (here and the two below) shouldn't be necessary, these lines were linting clean previously. This might have happened if you used a formatting tool that has a slightly stricter line length requirement than we lint for. It would be nice to revert these couple changes, since I think wrapping lines more than necessary can impact readability a bit.

nmoroze (commented):
Looks good to me, really nice job with the refactor! Always nice to see the code get simplified.

gadfort (approved):
LGTM

PR #1751, opened 2023-07-11
Remove install from linter job

troibe:
Removing install here speeds up linting a lot.
Also the feedback will be more useful feedback to the user.
The install gets executed in the `python_ci.yml` anyways for every push.

Tested that it still lints by adding violation:
https://github.com/siliconcompiler/siliconcompiler/actions/runs/5521649690/jobs/10070070190
```
./setup.py:7:1: F401 'streamlit' imported but unused
1     F401 'streamlit' imported but unused
Error: Process completed with exit code 1.
```

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,33 +10,14 @@ on:
 jobs:
   lint_python:
     name: Lint Python Code
-
-    strategy:
-      fail-fast: false
-      matrix:
-        version:
-          - {python: "3.6", os: "ubuntu-20.04"}
-          - {python: "3.7", os: "ubuntu-latest"}
-          - {python: "3.8", os: "ubuntu-latest"}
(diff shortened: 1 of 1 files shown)


gadfort (commented):
@nmoroze this looks good to me. It's rare that flake8 finds a violation in the different python versions. Do you think it matters if we lint in all versions vs. just the default?

nmoroze (approved):
This looks good, I don't think I've seen any examples of flake8 results differing between versions. Another idea could be to lint with both our minimum and maximum supported versions. This could catch use of language features that are unsupported in the minimum version, while the maximum would be able to run a newer flake8 version and possibly catch more things. I'll leave final decision on that up to you both.

troibe:
I'd argue if it's an actually unsupported language feature then the install itself in `python_ci.yml` should catch it.
For hints and styling we probably want to stick with the most recent version.

PR #1752, opened 2023-07-11
Enable pip caching

troibe:
Enabling the cache should save about 2min of time during test runs.

### DIFF
--- a/.github/workflows/python_ci.yml
+++ b/.github/workflows/python_ci.yml
@@ -31,6 +31,7 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.version.python }}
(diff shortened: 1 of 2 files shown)


nmoroze (approved):
LGTM. I re-ran the checks on this branch and confirmed the caching is happening, although the runtime delta seemed to vary quite a bit across versions. Perhaps the runtime of the tests is just a bit noisy, and this will help on average?

troibe:
Yea the issue here is that it's only caching the package download cache but not the installs.
Caching installs should make a bigger difference but currently requires more work.
Might be a worthwile optimization to look into in the future:
https://github.com/actions/setup-python/issues/330

PR #1753, opened 2023-07-11
Gabriel/sc logo

GabrielAguirre1:
Functionally, this works the same except the logo does not change size.

Here are two examples of the same page, not refreshed, but compressed.

Initial screen:
dashboard

After the screen's width shrinks, nothing changes:
dashboard

Limitations: 
- Since this is in HTML, some of the streamlit safety guards are taken off. If someone chooses to shrink the width to a very small window the logo text is not aligned properly


siliconcompiler/report/streamlit_viewer.py
```diff
@@ -568,7 +570,20 @@ def show_title_and_runs(title_col_width=0.7):
...
     with title_col:
-        streamlit.title(f'{new_chip.design} dashboard', anchor=False)
+        # in case ui_width is 0
+        if ui_width <= 0:
+            icon_width = 0.01
+            streamlit.title(f'{new_chip.design} dashboard', anchor=False)
+        else:
+            # 77 because you should resize by a factor of 0.5 - original pixel
+            # width is 308
+            icon_width = 77 / (ui_width * title_col_width)
```
gadfort, line 580: would it be better to open the image and get the number of pixels? incase it changes? we already open it to pass it in as an icon, so we can just use that again.

GabrielAguirre1:
Just realized there are merge conflicts. I will fix them now.

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -575,12 +576,62 @@ def show_title_and_runs(title_col_width=0.7):
...
+                    }
+                    .text2 {
+                        color: #1D4482; /* Blue color */
+                        font: IBM Plex Sans !important;
+                        font-weight:700 !important;
+                        font-size:30px !important;
+                    }
+                </style>
+            </head>''',
+            unsafe_allow_html=True
```
troibe, line 611: Why is it considered unsafe by streamlit?
Just style wise or something on top?

troibe:
@gadfort What is currently the blocker for merging this?

siliconcompiler/report/streamlit_viewer.py
GabrielAguirre1: Just looked it up. Streamlit cannot ensure that the HTML is not vulnerable to hacks.

troibe: Oh I see. I think if you are passing in a static string that should be fine.

GabrielAguirre1:
> Functionally, this works the same except the logo does not change size.
> 
> Here are two examples of the same page, not refreshed, but compressed.
> 
> Initial screen: dashboard
> 
> After the screen's width shrinks, nothing changes: dashboard
> 
> Limitations:
> 
> * Since this is in HTML, some of the streamlit safety guards are taken off. If someone chooses to shrink the width to a very small window the logo text is not aligned properly

Everything is the same in this new commit but the logo is resized. 

New logo:
Screen Shot 2023-07-18 at 11 31 22 AM

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -26,7 +27,7 @@
 
 
 sc_logo_path = \
-    os.path.join(os.path.dirname(__file__), '..', 'data', 'logo.png')
+    os.path.join(os.path.dirname(__file__), '..', 'data', 'resized_logo.png')
```
gadfort, line 30: I don't think we need the resized logo anymore, we can keep the original one

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -575,12 +576,63 @@ def show_title_and_runs(title_col_width=0.7):
...
+            unsafe_allow_html=True
+        )
+
+        streamlit.markdown(
+            f'''
+            <body>
+                <div class="logo-container">
+                    <img src="data:image/png;base64,{base64.b64encode(open(sc_logo_path,
+                    "rb").read()).decode()}" alt="Logo Image" class="logo-image"
+                    width="62" height="61">
```
gadfort, line 620: if you just specify the height, the width will be scaled correctly, no need to show both.

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -575,12 +576,63 @@ def show_title_and_runs(title_col_width=0.7):
...
+                        margin-top: -10px;
+                    }
+                    .logo-text {
+                        display: flex;
+                        flex-direction: column;
+                        margin-top: -20px;
+                    }
+                    .text1 {
+                        color: #F1C437; /* Yellow color */
+                        font: IBM Plex Sans !important;
```
gadfort, line 598: how common is this font? We already bundle a font with SC: https://github.com/siliconcompiler/siliconcompiler/blob/main/siliconcompiler/data/RobotoMono/RobotoMono-Regular.ttf  would it be possible to embed it into the HTML and use it instead to avoid being dependent on another font?

gadfort (changes requested):
Remove the extra new logo and width specification. Please look into the font embedding.

siliconcompiler/report/streamlit_viewer.py
GabrielAguirre1: Just did it!

GabrielAguirre1:
> Functionally, this works the same except the logo does not change size.
> 
> Here are two examples of the same page, not refreshed, but compressed.
> 
> Initial screen: dashboard
> 
> After the screen's width shrinks, nothing changes: dashboard
> 
> Limitations:
> 
> * Since this is in HTML, some of the streamlit safety guards are taken off. If someone chooses to shrink the width to a very small window the logo text is not aligned properly

Here is the logo and text with the proper font:
dashboard

gadfort (approved):
LGTM

PR #1754, opened 2023-07-11
resizes the flowgraph

GabrielAguirre1:
this resizes the flowgraph to take up a maximum amount of space, or 40% if the screen is smaller than 1300 pixels
Here is a before and after:

Before:
Screen Shot 2023-07-11 at 12 43 41 PM

After:
Screen Shot 2023-07-11 at 12 18 59 PM

Limitations:
- If the user changes the screen width without refreshing, the flowgraph will not change it’s size

### DIFF
--- a/siliconcompiler/report/streamlit_viewer.py
+++ b/siliconcompiler/report/streamlit_viewer.py
@@ -2,6 +2,7 @@ import streamlit
 from streamlit_agraph import agraph, Node, Edge, Config
 from streamlit_tree_select import tree_select
 from streamlit_toggle import st_toggle_switch
+import streamlit_javascript
 from PIL import Image
 from pathlib import Path
 import os
@@ -583,6 +584,8 @@ def show_title_and_runs(title_col_width=0.7):
     return new_chip
 
 
+ui_width = streamlit_javascript.st_javascript("window.innerWidth")
+
 new_chip = show_title_and_runs()
 
 # gathering data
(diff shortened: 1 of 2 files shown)


siliconcompiler/report/streamlit_viewer.py
```diff
@@ -632,10 +635,15 @@ def show_title_and_runs(title_col_width=0.7):
...
+        # in case streamlit_javascript errors
+        ui_width = 1
     if streamlit.session_state['flowgraph']:
-        node_from_flowgraph, datafram_and_node_info_col = show_flowgraph()
+        node_from_flowgraph, datafram_and_node_info_col = \
+            show_flowgraph(flowgraph_col_width=min(520 / ui_width, 0.4))
     else:
-        node_from_flowgraph, datafram_and_node_info_col = dont_show_flowgraph()
+        node_from_flowgraph, datafram_and_node_info_col = \
+            dont_show_flowgraph(flowgraph_col_width=min(120 / ui_width, 0.1))
```
gadfort, line 646: what do 520 and 120 mean? maybe you can make those variables so they get human readable names?

gadfort (changes requested):
Much better, just need the variables to named so it's easier to read.

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -583,6 +584,8 @@ def show_title_and_runs(title_col_width=0.7):
     return new_chip
 
 
+ui_width = streamlit_javascript.st_javascript("window.innerWidth")
```
troibe, line 587: Why is this in 100 lines away of where it is actually being used?

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -632,10 +635,15 @@ def show_title_and_runs(title_col_width=0.7):
     metrics_tab, manifest_tab, file_preview_tab = tabs
 
 with metrics_tab:
+    if ui_width <= 0:
+        # in case streamlit_javascript errors
+        ui_width = 1
```
troibe, line 640: What does this default translate to in terms of looks?

siliconcompiler/report/streamlit_viewer.py
GabrielAguirre1: `ui_width` is needed for this and the inclusion of the SC logo in the title. This seemed to be the most logical place to put it for both.

siliconcompiler/report/streamlit_viewer.py
gadfort: it might be better to not adjust this number and simply break up the calculations as:
```
width1 = 0.4
width2 = 0.1
if ui_width > 0:
  width1 = min(width1, .../ui_width)
  ...
```

siliconcompiler/report/streamlit_viewer.py
gadfort: lets move it closer to where it is needed here and if we decide on the SClogo then we can move it back in that PR.

troibe:
@GabrielAguirre1 I think you accidentally reverted some stuff in the merge.

gadfort (approved):

PR #1755, opened 2023-07-12
ensure calls to _runtask are spawned

gadfort:
When using threading and the remote, that was causing the os.chdir to interfere with eachother, this avoids this by ensure all calls to _runtask are done through a multiprocessing call.


WRansohoff (approved):
Looks great, thanks!

PR #1757, opened 2023-07-12
Fix server CI failures caused by a failure to detect failures in the local import task.

WRansohoff:
Moving the local import nodes into spawned processes makes them parallelizable, but it also means that failures are silently ignored.

This change detects failed local tasks, and prevents a remote run from being submitted when they don't succeed, which was the behavior before #1755 got merged.

### DIFF
--- a/siliconcompiler/remote/client.py
+++ b/siliconcompiler/remote/client.py
@@ -13,7 +13,7 @@ import tempfile
 import multiprocessing
 
 from siliconcompiler._metadata import default_server
-from siliconcompiler import utils, SiliconCompilerError
+from siliconcompiler import utils
 
 
 # Client / server timeout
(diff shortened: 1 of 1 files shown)


siliconcompiler/remote/client.py
```diff
@@ -152,6 +152,9 @@ def remote_preprocess(chip, steplist):
                                           args=(local_step, index, {}))
         run_task.start()
         run_task.join()
+        if run_task.exitcode != 0:
+            # A 'None' or nonzero value indicates that the Process target failed.
+            raise SiliconCompilerError("Could not start remote job: local setup task failed.")
```
gadfort, line 157: could you call `chip.error( fatal=True)` instead of directly raising.
Also the message might be a little more informative with:
"Could not start remote job: local setup task (<step><index>) failed."

gadfort (changes requested):
looks good, just needs a few small changes.

siliconcompiler/remote/client.py
WRansohoff: Sure, good idea.

gadfort (approved):
LGTM

PR #1758, opened 2023-07-12
add abstract_lef controls to OpenROAD

gadfort:
adds the flags needed to be able to control the abstract lef writing to the OpenROAD driver

### DIFF
--- a/siliconcompiler/tools/openroad/openroad.py
+++ b/siliconcompiler/tools/openroad/openroad.py
@@ -912,6 +912,16 @@ def _define_ord_params(chip):
              'list of files to use for specifying global connections',
              field='help')
 
+    _set_parameter(chip, param_key='ord_abstract_lef_bloat_factor',
+                   default_value='10',
+                   require=['key'],
+                   schelp='Factor to apply when writing the abstract lef')
+
+    _set_parameter(chip, param_key='ord_abstract_lef_bloat_layers',
+                   default_value='false',
+                   require=['key'],
+                   schelp='Fill all layers when writing the abstract lef')
+
 
 def _define_pex_params(chip):
     step = chip.get('arg', 'step')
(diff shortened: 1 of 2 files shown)


siliconcompiler/tools/openroad/scripts/sc_export.tcl
```diff
@@ -1,5 +1,11 @@
 # Write LEF
-write_abstract_lef "outputs/${sc_design}.lef"
+set lef_args []
+if { [lindex [dict get $sc_cfg tool $sc_tool task $sc_task {var} ord_abstract_lef_bloat_layers] 0] == "true" } {
+  lappend lef_args "-bloat_occupied_layers"
+} else {
```
troibe, line 5: Not familiar with the semantics...
Are `-bloat_occupied_layers` and `-bloat_factor` mutually exclusive?

siliconcompiler/tools/openroad/openroad.py
```diff
@@ -912,6 +912,16 @@ def _define_ord_params(chip):
              'list of files to use for specifying global connections',
              field='help')
 
+    _set_parameter(chip, param_key='ord_abstract_lef_bloat_factor',
+                   default_value='10',
```
troibe, line 916: What does this default value translate to?
Is it the same default that openroad had before?

troibe (approved):
Looks like it matches openroads parameters.

siliconcompiler/tools/openroad/openroad.py
gadfort: it's the same ad the OpenROAD default. It's the number routing pitches the abstract generator will bloat the shapes by to attempt to simplify the layout. Unfortunately that is not always perfect and leaves some complex shapes behind.

PR #1759, opened 2023-07-12
Multithread Verilator builds

nmoroze:
This PR updates our Verilator driver to multithread builds according to the SC threads parameter. It uses Verilator's `-j` flag, which should enable multi-threading both in the "Verilation" (RTL->C++) phase and in the C++ compile phase. 

In addition, I decided to bump the minimum required Verilator in order to make use of Verilator's `--build` flag and simplify the driver a bit. 4.028 was the version available in apt for Ubuntu 20.04, but I think newer versions of Verilator are more widespread now (22.04 includes 4.038).


gadfort (approved):
LGTM

PR #1766, opened 2023-07-14
Dependency caching

troibe:
Utilize pip to organize dependencies e.g. for PDKs.
The installed package needs to provide a `dependency.py` file which contains a `dependency` dict with all the source information.
Siliconcompiler reads this information and if possible downloads+caches the information to the location of the freepdk45 package.
Alternatively a path in the network drive can be specified.

Currently only the final path path of the pdk will be stored in the pdk information of the schema as before.
This path contains the unique id (project_name+commit_id or a custom unique_id).
We could also store this information in a more prominent place in the schema.

Depends on the [dependency-caching branch of freepdk45](https://github.com/siliconcompiler/freepdk45/tree/dependency-caching).

### DIFF
--- a/siliconcompiler/__init__.py
+++ b/siliconcompiler/__init__.py
@@ -7,6 +7,8 @@ from siliconcompiler._metadata import version as __version__
 
 from siliconcompiler.use import PDK, Library, Flow, Checklist
 
+import siliconcompiler.dependency as dep
+
 __all__ = [
     "__version__",
     "Chip",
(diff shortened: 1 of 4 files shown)


troibe:
`freepdk45-data` is not installed on the CI right now which leads to tests failing.

Just wanted to get some initial feedback on the approach.
After that I'll publish the `freepdk45-data` package on pypi and add it as a requirement.

troibe:
Abandoning this in favor of newer PRs #2071, #2074

PR #1767, opened 2023-07-14
cleanup openroad and yosys setup

gadfort:
Changes:
- fixes incorrect yosys scripts from ORFS
- adds steps to mirror ORFS for global placement and pin placement

Fixes:
- openroad driver using clock estimation layer for signal estimations

### DIFF
--- a/siliconcompiler/libs/sky130hd.py
+++ b/siliconcompiler/libs/sky130hd.py
@@ -71,7 +71,8 @@ def setup(chip):
     lib.add('asic', 'cells', 'endcap', 'sky130_fd_sc_hd__decap_4')
 
     lib.add('asic', 'cells', 'dontuse', [
-        'sky130_fd_sc_hd__probe_p_8 sky130_fd_sc_hd__probec_p_8',
+        'sky130_fd_sc_hd__probe_p_8',
+        'sky130_fd_sc_hd__probec_p_8',
         'sky130_fd_sc_hd__lpflow_bleeder_1',
         'sky130_fd_sc_hd__lpflow_clkbufkapwr_1',
         'sky130_fd_sc_hd__lpflow_clkbufkapwr_16',
--- a/siliconcompiler/libs/sky130hd.py
+++ b/siliconcompiler/libs/sky130hd.py
@@ -118,9 +118,9 @@ def setup(chip):
 
     # Defaults for OpenROAD tool variables
     lib.set('option', 'var', 'openroad_place_density', '0.6')
-    lib.set('option', 'var', 'openroad_pad_global_place', '4')
-    lib.set('option', 'var', 'openroad_pad_detail_place', '2')
-    lib.set('option', 'var', 'openroad_macro_place_halo', ['1', '1'])
+    lib.set('option', 'var', 'openroad_pad_global_place', '1')
+    lib.set('option', 'var', 'openroad_pad_detail_place', '0')
+    lib.set('option', 'var', 'openroad_macro_place_halo', ['40', '40'])
     lib.set('option', 'var', 'openroad_macro_place_channel', ['80', '80'])
 
     # Yosys techmap
--- a/siliconcompiler/libs/sky130hd.py
+++ b/siliconcompiler/libs/sky130hd.py
@@ -71,8 +71,7 @@ def setup(chip):
     lib.add('asic', 'cells', 'endcap', 'sky130_fd_sc_hd__decap_4')
 
     lib.add('asic', 'cells', 'dontuse', [
-        'sky130_fd_sc_hd__probe_p_8',
-        'sky130_fd_sc_hd__probec_p_8',
+        'sky130_fd_sc_hd__probe_p_8 sky130_fd_sc_hd__probec_p_8',
         'sky130_fd_sc_hd__lpflow_bleeder_1',
         'sky130_fd_sc_hd__lpflow_clkbufkapwr_1',
         'sky130_fd_sc_hd__lpflow_clkbufkapwr_16',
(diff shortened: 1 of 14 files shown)


siliconcompiler/libs/sky130hd.py
```diff
@@ -71,8 +71,7 @@ def setup(chip):
     lib.add('asic', 'cells', 'endcap', 'sky130_fd_sc_hd__decap_4')
 
     lib.add('asic', 'cells', 'dontuse', [
-        'sky130_fd_sc_hd__probe_p_8',
-        'sky130_fd_sc_hd__probec_p_8',
+        'sky130_fd_sc_hd__probe_p_8 sky130_fd_sc_hd__probec_p_8',
```
troibe: This one looks odd to me?

troibe: Nevermind fixed in next commit...

troibe: Nevermind fixed in next commit...

siliconcompiler/tools/openroad/scripts/sc_place.tcl
```diff
@@ -2,7 +2,7 @@
 # Global Placement (without considering IO placements)
 #######################
 
-if { $openroad_gpl_enable_skip_io } {
+if { $openroad_gpl_enable_skip_io && [sc_design_has_placeable_ios] } {
```
troibe, line 5: Shouldn't this be the negation?

siliconcompiler/tools/openroad/scripts/sc_procs.tcl
```diff
@@ -110,3 +110,128 @@ proc sc_pin_placement { args } {
...
+# Get supply nets in design
+###########################
+
+proc sc_supply_nets {} {
+  set nets []
+
+  foreach net [[ord::get_db_block] getNets] {
+    set type [$net getSigType]
+    if {$type == "POWER" || $type == "GROUND"} {
+      lappend nets [$net getName]
```
troibe, line 218: Why did you choose to only return the net name?

siliconcompiler/tools/openroad/scripts/sc_floorplan.tcl
```diff
@@ -438,20 +418,17 @@ if {$openroad_pdn_enable == "true" && \
...
-    if { ![$net isSpecial] } {
-      utl::warn FLW 1 "$net_name is marked as a supply net, but is not marked as a special net"
-    }
-    if { $openroad_psm_enable == "true" } {
-      puts "Check supply net: $net_name"
-      check_power_grid -net $net_name
-    }
+foreach net [sc_supply_nets] {
+  if { ![[[ord::get_db_block] findNet $net] isSpecial] } {
+    utl::warn FLW 1 "$net_name is marked as a supply net, but is not marked as a special net"
```
troibe, line 423: Just asking because you search for the net after you use `sc_supply_nets` here?

siliconcompiler/tools/openroad/scripts/sc_procs.tcl
```diff
@@ -110,3 +110,128 @@ proc sc_pin_placement { args } {
     {*}$openroad_ppl_arguments \
     {*}$ppl_args
 }
+
+###########################
+# Check if OR has a GUI
+###########################
+
+proc sc_has_gui {} {
+  if {[expr [llength [info procs save_image]] > 0]} {
```
troibe, line 119: I know right now `save_image` and the SC GUI are pretty related.
Do you think it makes sense to have a more independent GUI flag in tcl?
Maybe in the future the SC user wants to create that image but is not using the GUI feature.

siliconcompiler/tools/openroad/scripts/sc_place.tcl
gadfort: This call to global_placement only makes sense if the IOs are still placeable, otherwise the next call to pin placement doesn't do much, but I can reorganize this to make it easier to see.

siliconcompiler/tools/openroad/scripts/sc_floorplan.tcl
gadfort: OpenROAD uses a hash in findNet so it's relatively fast

siliconcompiler/tools/openroad/scripts/sc_procs.tcl
gadfort: `sc_has_gui` is just a helper function to check if OpenROAD was compiled with GUI support. I've just been prepending `sc_` to the helper functions to try and avoid name collisions

siliconcompiler/tools/openroad/scripts/sc_place.tcl
troibe: Oh now I see. I didn't realize this step would **only** get executed when the IOs are present.

siliconcompiler/tools/openroad/scripts/sc_procs.tcl
troibe: Oh I thought this is referring to whether siliconcompiler's gui is turned on.
If it's about OpenROAD then it looks good to me.

siliconcompiler/tools/openroad/scripts/sc_place.tcl
```diff
@@ -1,17 +1,21 @@
 #######################
-# Global Placement (without considering IO placements)
+# Global Placement and Refinement of Pin Placement
 #######################
 
-if { $openroad_gpl_enable_skip_io && [sc_design_has_placeable_ios] } {
-  utl::info FLW 1 "Performing global placement without considering IO"
-  sc_global_placement -skip_io
-}
+if { [sc_design_has_placeable_ios] } {
```
troibe, line 5: Great that's how I would have expected it!

troibe (approved):
Looks good to me so far!

siliconcompiler/tools/openroad/scripts/sc_apr.tcl
```diff
@@ -405,7 +405,7 @@ if { $sc_task == "show" || $sc_task == "screenshot" } {
 
   report_units_metric
 
-  utl::set_metrics_stage "sc__prestep__{}"
+  utl::push_metrics_stage "sc__prestep__{}"
```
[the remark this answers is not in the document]
gadfort: push_metrics_stage  preserves the previous stack of metrics, set_metrics_stage just clears it.
This will be useful once we can start collecting finer-grain metrics in the actual scripts as well. Right now it's a NOP.

PR #1769, opened 2023-07-14
fix path to json api for server

gadfort:
Server API is correcting building again: https://siliconcompiler--1769.org.readthedocs.build/en/1769/reference_manual/server_api.html


WRansohoff (approved):
Ah, that explains why the docs page was blank. Thanks for fixing that!

PR #1774, opened 2023-07-17
Improve feedback in remote progress checks

WRansohoff:
This change updates the remote client to make use of extra information that the server will start to return with the next update. It will print more information with the progress checks, and fetch partial results for each node in the flowgraph as they complete.

The printout for each progress check will look something like this:

```
| INFO    | job0  | remote     | 0  | Job is still running. Status:
| INFO    | job0  | remote     | 0  |   Completed (2): syn0, floorplan0
| INFO    | job0  | remote     | 0  |   Running (1):
| INFO    | job0  | remote     | 0  |     physyn0 (00:01:40)
| INFO    | job0  | remote     | 0  |   Pending (6): place0, cts0, route0, ...
| INFO    | job0  | remote     | 0  |   Fetching newly-completed results:
| INFO    | job0  | remote     | 0  |     syn0
| INFO    | job0  | remote     | 0  |     floorplan0
```


troibe (commented):
Looks good to me. Since server and client are not automatically synced could you add the corresponding scserver commit that this depends on?

siliconcompiler/remote/client.py
```diff
@@ -196,30 +196,116 @@ def remote_run(chip):
...
-                    cur_step = msg_lines[0][msg_lines[0].find(': ') + 2:]
-                    cur_log = '\n'.join(msg_lines[1:])
-                    chip.logger.info("Job is still running (%d seconds, step: %s)." % (
-                                     int(time.monotonic() - step_start), cur_step))
-                    if cur_log:
-                        chip.logger.info(f"Tail of current logfile:\n{cur_log}\n")
-                else:
-                    chip.logger.info("Job is still running (%d seconds, step: unknown)" % (
-                                     int(time.monotonic() - step_start)))
+                try:
```
gadfort, line 212: this block is quite long, could you please refactor.

siliconcompiler/remote/client.py
```diff
@@ -196,30 +196,116 @@ def remote_run(chip):
...
+                    for node, node_info in job_info.items():
+                        status = node_info['status']
+                        nodes_to_log[status].append((node, node_info))
+                        if (status == 'completed') and \
+                           (node not in completed):
+                            completed.append(node)
+                            nodes_to_fetch.append(node)
+
+                    # Log information about the job's progress.
+                    # To avoid clutter, only log up to 3 completed/pending nodes, on a single line.
```
gadfort, line 233: make 3 a configurable number

siliconcompiler/remote/client.py
```diff
@@ -341,9 +427,11 @@ def success_action(resp):
 
 
 ###################################
-def fetch_results_request(chip):
+def fetch_results_request(chip, node=''):
```
gadfort, line 430: node=None (to be consistent with the rest of SC

siliconcompiler/remote/client.py
```diff
@@ -371,13 +462,15 @@ def success_action(resp):
 
 
 ###################################
-def fetch_results(chip):
+def fetch_results(chip, node=''):
```
gadfort, line 465: node=None

siliconcompiler/remote/client.py
```diff
@@ -395,25 +489,27 @@ def fetch_results(chip):
 
     # Unauthenticated jobs get a gzip archive, authenticated jobs get nested archives.
     # So we need to extract and delete those.
-    with tarfile.open(f'{job_hash}.tar.gz', 'r:gz') as tar:
-        tar.extractall()
+    with tarfile.open(f'{job_hash}{node}.tar.gz', 'r:gz') as tar:
+        tar.extractall(path=(node if node else ''))
```
gadfort, line 493: what does the tarball look like? does it have build/design/jobname/step/index or is it different?

siliconcompiler/remote/client.py
gadfort, line 430: add a file parameters so we can specify the output path for the file.

siliconcompiler/remote/client.py
```diff
@@ -395,25 +489,27 @@ def fetch_results(chip):
...
     for import_link in glob.iglob(job_hash + '/' + top_design + '/**/*',
                                   recursive=True):
         if os.path.islink(import_link):
             os.remove(import_link)
     # Copy the results into the local build directory, and remove the
-    # unzipped directory (including encrypted archives).
-    utils.copytree(job_hash,
+    # unzipped directory.
+    basedir = os.path.join(node, job_hash) if node else job_hash
+    utils.copytree(basedir,
```
gadfort, line 505: is any of this needed if we are archiving the node, shouldn't we just untar in place?

gadfort (changes requested):
The main body of the remote run method should be broken into smaller more manageable chunks. Please clarify how the tarball looks on return and my preference would be to untar it in place instead of copying the files around afterwards.

siliconcompiler/remote/client.py
WRansohoff: Once the upcoming `get_results/` changes are merged, it will have the results of the SC archive calls. So, it will contain the `('option', 'builddir')` top level with the subset of result files which get returned.

siliconcompiler/remote/client.py
WRansohoff: The build directory gets renamed on the server-side to include the job ID, so it won't be the same as the client's `('option', 'builddir')`. Hence the renaming.

siliconcompiler/remote/client.py
WRansohoff: Yes - actually that's part of why I want to get this merged soon, to pull out the logic which prints `check_progress` results for the sake of the `sc-remote` CLI app. I guess I can just do it here.

siliconcompiler/remote/client.py
WRansohoff: The `node` value is used to assemble file paths, so if it is set to `None`, the file paths will include a `None` string where they are expected to be empty.

WRansohoff: Per offline discussion, I'm making these a required argument.

siliconcompiler/remote/server.py
```diff
@@ -323,10 +324,25 @@ async def remote_sc(self, chip, username):
...
+                                               f'{job_hash}_{step}{index}.tar.gz'),
+                                  mode='w:gz')
+                chip._archive_node(tf, step=step, index=index)
+                tf.close()
+        chip.set('option', 'steplist', steplist)
 
         # Create a single-file archive to return if results are requested.
-        with tarfile.open(os.path.join(self.nfs_mount, f'{job_hash}.tar.gz'), "w:gz") as tar:
+        # TODO: Remove after API updates go out.
+        with tarfile.open(os.path.join(self.nfs_mount, job_hash, f'{job_hash}_.tar.gz'), "w:gz") as tar:
```
gadfort, line 345: why is this behavior different? it seems like we would want server.py to have the proper API implemented.

gadfort (changes requested):
I think we need to ensure server.py is honoring the proper API for the CI to work.

siliconcompiler/remote/server.py
```diff
@@ -323,10 +324,25 @@ async def remote_sc(self, chip, username):
...
+            indexlist = chip.getkeys('flowgraph', chip.get('option', 'flow'), step)
+            for index in indexlist:
+                chip.cwd = os.path.join(chip.get('option', 'builddir'), '..')
+                tf = tarfile.open(os.path.join(self.nfs_mount,
+                                               job_hash,
+                                               f'{job_hash}_{step}{index}.tar.gz'),
+                                  mode='w:gz')
+                chip._archive_node(tf, step=step, index=index)
+                tf.close()
+        chip.set('option', 'steplist', steplist)
```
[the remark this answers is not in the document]
WRansohoff: Sure, we can do that - I was a little worried that this wouldn't work on complex graphs, but figured it's a minimal example implementation.

siliconcompiler/remote/client.py
```diff
@@ -386,34 +523,35 @@ def fetch_results(chip):
                    f"(Response code: {results_code})", fatal=True)
 
     # Call 'delete_job' to remove the run from the server.
-    delete_job(chip)
+    if not node:
+        delete_job(chip)
```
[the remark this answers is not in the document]
WRansohoff: Not anymore, good point.

siliconcompiler/remote/client.py
```diff
@@ -196,30 +278,70 @@ def remote_run(chip):
...
+    for node in all_nodes:
+        if node not in completed:
+            node_proc = multiprocessor.Process(target=fetch_results,
+                                               args=(chip, node))
+            node_proc.start()
+            result_procs.append(node_proc)
+    # Make sure all results are fetched before letting the client issue
+    # a deletion request.
+    for proc in result_procs:
+        proc.join()
```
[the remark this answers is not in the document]
WRansohoff: Yes, because the final `check_progress/` response doesn't necessarily include node info; it might just says that the job is completed.

siliconcompiler/remote/client.py
```diff
@@ -196,30 +278,70 @@ def remote_run(chip):
...
+                nodes_to_fetch.append(node)
+        if nodes_to_fetch:
+            chip.logger.info('  Fetching newly-completed results:')
+            for node in nodes_to_fetch:
+                node_proc = multiprocessor.Process(target=fetch_results,
+                                                   args=(chip, node))
+                node_proc.start()
+                result_procs.append(node_proc)
+                chip.logger.info(f'    {node}')
+        completed = new_completed.copy()
```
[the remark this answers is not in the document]
WRansohoff: The `new_completed` list is a superset of the previous `completed` list. Maybe I should rename the variable if that's confusing.

siliconcompiler/core.py
```diff
@@ -3854,6 +3854,66 @@ def _setup_task(self, step, index):
...
+            # completed successfully. Inspired by:
+            # https://stackoverflow.com/a/70029046.
+
+            os.listdir(os.path.dirname(lastdir))
+
+            lastcfg = f"{lastdir}/outputs/{self.get('design')}.pkg.json"
+            # Determine if the task was successful, using provided status dict
+            # or the node Schema if no status dict is available.
+            stat_success = False
+            if status:
```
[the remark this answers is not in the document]
WRansohoff: Providing an empty status dict, and the finalize method falls back to the per-node manifests for fetching a task's status.

siliconcompiler/remote/server.py
WRansohoff: Storing a whole-job archive makes the server backwards-compatible with older clients - once the API update is out on pypi, we can safely remove this.

Although, since the code is in the SC repo's development server, we can probably just remove it now.

siliconcompiler/remote/client.py
```diff
@@ -196,30 +283,71 @@ def remote_run(chip):
...
+    # Make sure all results are fetched before letting the client issue
+    # a deletion request.
+    for proc in result_procs:
+        proc.join()
+
+    # Produce final manifest, based on the returned per-node results.
+    # (Un-set the 'remote' option to avoid steplist-based summary/show errors)
+    chip.unset('option', 'remote')
+    chip._finalize_run(chip.list_steps(),
+                       copy.deepcopy(os.environ))
```
gadfort, line 329: you need to use the original environment: https://github.com/siliconcompiler/siliconcompiler/blob/04df3ec4b1c02080330311e3777a1dc13d6d890b/siliconcompiler/core.py#L4047C9-L4047C21

siliconcompiler/remote/client.py
```diff
@@ -196,30 +283,71 @@ def remote_run(chip):
...
+
+
+###################################
+def check_progress(chip, step_start, all_nodes, completed=[]):
```
gadfort, line 333: completed=None

siliconcompiler/remote/client.py
```diff
@@ -196,30 +283,71 @@ def remote_run(chip):
...
+    try:
+        is_busy_info = is_job_busy(chip)
+        is_busy = is_busy_info['busy']
+        if is_busy:
+            all_completed = _process_progress_info(chip,
+                                                   is_busy_info,
+                                                   step_start,
+                                                   completed)
```
gadfort, line 341: the return value is not needed since you can just operate directly on the input argument., that looks to avoid all the copies in this code.

siliconcompiler/remote/client.py
```diff
@@ -168,6 +171,90 @@ def remote_preprocess(chip, steplist):
...
+    try:
+        # Decode response JSON, if possible.
+        job_info = json.loads(progress_info['message'])
+        # Retrieve total elapsed time, if included in the response.
+        total_elapsed = ''
+        if 'elapsed_time' in job_info:
+            total_elapsed = f' (runtime: {job_info["elapsed_time"]})'
+
+        # Sort and store info about the job's progress.
+        all_completed = completed.copy()
```
gadfort, line 211: no need to copy, just append to the org list

siliconcompiler/remote/client.py
```diff
@@ -196,30 +283,71 @@ def remote_run(chip):
...
+                nodes_to_fetch.append(node)
+        if nodes_to_fetch:
+            chip.logger.info('  Fetching completed results:')
+            for node in nodes_to_fetch:
+                node_proc = multiprocessor.Process(target=fetch_results,
+                                                   args=(chip, node))
+                node_proc.start()
+                result_procs.append(node_proc)
+                chip.logger.info(f'    {node}')
+        completed = all_completed.copy()
```
gadfort, line 310: this copy is just copying a copy, no need for it

gadfort (changes requested):
Please correct he environment thing and default value of complete=[].
I don't think you need all these copies, so I'd prefer to remove them.

gadfort (approved):
LGTM, Thanks

PR #1778, opened 2023-07-18
rename test_dasboard -> test_dashboard

gadfort:


troibe (approved):
Should be fine. The filename isn't mentioned anywhere explicitely.

PR #1779, opened 2023-07-18
Gabriel/toggle switch to checkbox

GabrielAguirre1:
I wanted a toggle for the user to be able to show the manifest unfiltered. I have had trouble using the current toggle (a widget we import) rendering on unselected tabs. I'm not sure how to fix this problem, so while I figure that out, I think a good solution is to use a checkbox that is native to Streamlit (which avoids this error). A consequence of this decision is that all toggles should really be checkboxes - it's silly to have them mismatched.

Here is a photo of the checkboxes where the toggles were:
Flowgraph

Here is a photo of the checkbox to switch the manifest from filtered to unfiltered:
Manifest Tree

I have tested the functionality of the checkboxes and the functionality of the search bars given the unfiltered manifest.


siliconcompiler/report/streamlit_viewer.py
```diff
@@ -448,25 +446,8 @@ def show_dataframe_header(header_col_width=0.7):
...
-        # this if statement deals with a problem with st_toggle_switch. The
-        # widget does not update during the streamlit.experimental_rerun.
-        # We need to keep track of the 'true' flip of the toggle to make sure
-        # everything is synced
-        if 'right after rerun' in streamlit.session_state and \
-           streamlit.session_state['right after rerun']:
-            transpose = streamlit.session_state['transpose']
-            streamlit.session_state['right after rerun'] = False
-        streamlit.session_state['transpose'] = transpose
+        streamlit.markdown('')
```
troibe, line 449: Do you need this twice for alignment?

troibe:
So the current behaviour is that there is still some filtering going on even if the user doesn't search for any keys or values?
What is getting filtered and why?

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -309,7 +302,7 @@ def show_metrics_for_file(chip, step, index):
             streamlit.warning("This file does not include any metrics.")
 
 
-def show_manifest(manifest, max_num_of_keys_to_show=20):
+def show_manifest(chip, manifest, max_num_of_keys_to_show=20):
```
troibe, line 305: Wouldn't it make more sense to have filtering enabled/disabled in your `make_manifest` function?
If this is way easier feel free to leave it though.

siliconcompiler/report/streamlit_viewer.py
GabrielAguirre1: Unfortunately, yes.

GabrielAguirre1:
> So the current behaviour is that there is still some filtering going on even if the user doesn't search for any keys or values? What is getting filtered and why?

There are a few places where the manifest is simplified. Some of the keys are labeled as 'default', and those branches are essentially duplicate information, so I remove those branches. Some of the key follow the pattern `manifest_branch[step][index]`, and those have been concatenated to `manifest_branch[step + index]`. Finally, the leaves hold about 7 keys and only one of the values of those keys is useful to the user, so I've simplified the leaf to that value.

gadfort:
@GabrielAguirre1 I think the phrase "filtered" is a bit misleading, you are showing an unfiltered manifest (but it's been cleaned up a bit), so maybe "raw" is a better term?

GabrielAguirre1:
> @GabrielAguirre1 I think the phrase "filtered" is a bit misleading, you are showing an unfiltered manifest (but it's been cleaned up a bit), so maybe "raw" is a better term?

Sounds good. Just changed it.

siliconcompiler/report/streamlit_viewer.py
troibe: I guess it depends on whether you want the user to allow filtering the "raw" manifest as well or just the "clean" manifest.
@gadfort Would filtering on the "raw" manifest provide any value to the user or is it too overwhelming?

troibe: Nevermind filtering comes later so this works.

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -322,19 +315,24 @@ def show_manifest(manifest, max_num_of_keys_to_show=20):
 
     key_search_col, value_search_col = streamlit.columns(2, gap="large")
 
+    if streamlit.checkbox('raw'):
```
gadfort, line 318: can you move the checkbox up next to to the search bars?

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -477,16 +458,9 @@ def display_flowgraph_toggle(label_after):
...
-                                 key="flowgraph_toggle",
-                                 default_value=label_after,
-                                 label_after=True,
-                                 # the colors are optional
-                                 inactive_color=INACTIVE_TOGGLE_COLOR,
-                                 active_color=ACTIVE_TOGGLE_COLOR,
-                                 track_color=TRACK_TOGGLE_COLOR)
+    streamlit.markdown("")
+    streamlit.markdown("")
+    fg_toggle = not streamlit.checkbox('Hide')
```
gadfort, line 463: When this is hidden it still says Hide, but it's not clear what it is doing. This should probably be changed to indicate this is the flowgraph somehow.

gadfort (changes requested):
Please figure out the flowgraph hiding and moving of the Raw switch.
You should add `.., help=''` to the checkboxes to provide more context about what they are supposed to it.

GabrielAguirre1:
I made the changes suggested.

Here is what the Hide flowgraph checkbox looks like. And the '?' is added for the tooltip. I'm not sure how I feel about it yet, but I'm not sure if I can change it at the moment. 
Screen Shot 2023-07-20 at 5 04 52 PM

Here is the checkbox for getting the raw manifest. I implemented logic so that the checkbox doesn't take up too much space on a larger monitor. 
Manifest Tree

GabrielAguirre1:
Here are the two fixes @gadfort and I spoke about. 

Here is the toggle when the flowgraph is hidden.
Data Metrics

Here is the toggle when the flowgraph is not hidden.
Flowgraph

Here is "Raw Manifest" --> "Raw manifest".
Screen Shot 2023-07-31 at 10 20 35 AM

gadfort (approved):

PR #1783, opened 2023-07-19
Update remote client to use job IDs generated by the server.

WRansohoff:
Since the "job hash" identifier is now implemented as a UUID rather than a literal hash of the files/archive being uploaded, we can have the server generate it for the client instead of accepting it as an input parameter.

This change updates SC's remote client to use a job ID from the server's response, instead of generating its own.

### DIFF
--- a/siliconcompiler/remote/client.py
+++ b/siliconcompiler/remote/client.py
@@ -7,7 +7,6 @@ import requests
 import shutil
 import time
 import urllib.parse
-import uuid
 import tarfile
 import tempfile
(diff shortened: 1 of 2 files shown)


gadfort (commented):
Look good, but it looks like you aren't doing anything with the interval value, that should probably be used in the `is_busy` loop

gadfort (approved):
Please add issue to track updating the `is_busy` behavior

troibe (approved):

PR #1787, opened 2023-07-20
bump minimum version of OR to better support IR drop analysis

gadfort:
Adds:
- final IR drop check to export

### DIFF
--- a/siliconcompiler/tools/openroad/scripts/sc_apr.tcl
+++ b/siliconcompiler/tools/openroad/scripts/sc_apr.tcl
@@ -153,6 +153,7 @@ source "$sc_refdir/sc_procs.tcl"
 ###############################
 
 # Read Liberty
+utl::info FLW 1 "Defining timing corners: $sc_corners"
 define_corners {*}$sc_corners
 foreach lib "$sc_targetlibs $sc_macrolibs" {
   #Liberty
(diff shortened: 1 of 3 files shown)


siliconcompiler/tools/openroad/scripts/sc_apr.tcl
```diff
@@ -240,7 +241,7 @@ set openroad_ifp_tie_separation [lindex [dict get $sc_cfg tool $sc_tool task $sc
 set openroad_pdn_enable [lindex [dict get $sc_cfg tool $sc_tool task $sc_task {var} pdn_enable] 0]
 
 set openroad_psm_enable [lindex [dict get $sc_cfg tool $sc_tool task $sc_task {var} psm_enable] 0]
-set openroad_psm_skip_nets [lindex [dict get $sc_cfg tool $sc_tool task $sc_task {var} psm_skip_nets] 0]
```
troibe, line 243: What is this change for?

troibe (approved):
LGTM

siliconcompiler/tools/openroad/scripts/sc_apr.tcl
troibe: Yea makes sense. I was just confused as to why before that you were getting only the first element...

PR #1796, opened 2023-07-24
Gabriel/dashboard/graphs

GabrielAguirre1:
This will give users the opportunity to view data against other runs in a chip's history and/or other chip schemas the user can pass in through the CLI. I've included new test cases for the functions I added to report.py.

The tab has an area to select which jobs/chips to include in the comparison for all graph, a sliding bar to select how many graphs to initialize, and the graphs. Points can be hovered over to display the exact value.
Screen Shot 2023-07-21 at 4 51 08 PM

Users must select 1 metric and as many or as few nodes as they would like.
Screen Shot 2023-07-24 at 2 52 41 PM

This is the area to select which jobs/chips to include:
Metrics Manifest de Preview Graphs

Edge cases:

If the user decides to select no nodes (this cannot be prevented to my knowledge) this graph will appear. "Errors" just happens to be the metric it is on.
Select a Metric

If the data contains some blank values for some runs and not for others this screen will be displayed. If all nodes do not have a value for a given run, the run will not be displayed on the Y-axis. 

Screen Shot 2023-07-24 at 1 56 31 PM


gadfort:
@GabrielAguirre1 please resolve merge conflicts

GabrielAguirre1:
I left out an additional detail.

I added a flag `-comparison_chip` to let users include more chips. They are forced in the history of the master chip object.
You can also use them for comparisons on the graphs which was the idea of including this feature in this PR. In the photo below, both chip1 and chip2 are added.
Screen Shot 2023-07-24 at 3 16 07 PM

troibe:
> If the data contains some blank values for some runs and not for others this screen will be displayed.

Which values are missing in that example?

GabrielAguirre1:
> > If the data contains some blank values for some runs and not for others this screen will be displayed.
> 
> Which values are missing in that example?

The value for import0 on job0_optimize_1.

siliconcompiler/report/streamlit_report.py
```diff
@@ -41,7 +41,22 @@ def __init__(self, chip, port=None):
             ("server.port", self.__port)
         ]
 
-        self.__config = {"manifest": self.__manifest}
+        # pass in a json object called __comparison_chips
+        # the key is the chip_name and value is the filepath
+        # if another argument is passed
+        self.__comparison_chips = []
```
gadfort, line 47: add comment to indicate that the use of list is to preserve order

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -51,9 +52,13 @@
 
     chip = Chip(design='')
     chip.read_manifest(config["manifest"])
+    for file_path in config['comparison_chips']:
+        comparison_chip = Chip(design='')
+        comparison_chip.read_manifest(file_path)
+        chip.schema.cfg['history'][os.path.basename(file_path)] = comparison_chip.schema.cfg
```
gadfort, line 58: we seem to have lost the name a user might have provided

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -8,6 +8,7 @@
 import argparse
 import json
 import pandas
+import altair
```
gadfort, line 11: new dependency?

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -696,6 +794,51 @@ def make_metric_to_metric_unit_map(metric_dataframe):
...
+        new_chip.schema = chip.schema.history(job)
+        new_chip.set('design', chip.design)
+        chips.append({'chip_object': new_chip, 'chip_name': job})
+        jobs.append(job)
+    job_selector_col, graph_adder_col = streamlit.columns(2, gap='large')
+    with job_selector_col:
+        with streamlit.expander('Select Jobs'):
+            selected_jobs = select_runs(jobs)
+    with graph_adder_col:
+        graphs = streamlit.slider('pick the number of graphs you want', 0, 10,
```
gadfort, line 820: does it make sense to have 0 graphs?

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -708,14 +851,14 @@ def make_metric_to_metric_unit_map(metric_dataframe):
     # setting up tabs
     if 'flowgraph' not in streamlit.session_state:
         streamlit.session_state['flowgraph'] = True
+    tabs = ["Metrics", "Manifest", "File Viewer", "Graphs"]
```
gadfort, line 854: lets only show the graphs tab is there is something to graph (ie. more than one history item)

gadfort (changes requested):
We should think about the `comparison_chip`, I don't like it, maybe `-graph_cfg` for the CLI and `graph_chips` for the `_dashboard`?

siliconcompiler/report/report.py
```diff
@@ -294,12 +285,57 @@ def get_files(chip, step, index):
...
+                    value = int(value)
+                else:
+                    try:
+                        value = float(value)
+                    except (TypeError, ValueError):
+                        pass
+                if metric_unit:
+                    metric_units.add(metric_unit)
+                chips_included.add(chip_name)
+                metric_datapoints[(step, index)][chip_name] = value
```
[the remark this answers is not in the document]
GabrielAguirre1: I could, but that would lead to a lot of extra computation with how the `utils._collect_data`  works at the moment. If there are 20 metrics and 10 nodes on 10 chips for a query of 1 metric and 1 node and all 10 chips, the bulk of the computation runs 10 times with how this function is constructed. Your suggestion would cause this to run 2000 times. We could speed it up by getting all the metrics tables at the start and reusing those. I'm not sure which option is better though.

siliconcompiler/report/streamlit_report.py
```diff
@@ -41,7 +41,22 @@ def __init__(self, chip, port=None):
...
+        self.__comparison_chips_names = []
+        if comparison_chips:
+            for chip_object_and_name in comparison_chips:
+                chip_file_path = \
+                    os.path.join(self.__directory,
+                                 f"{chip_object_and_name['name']}.json")
+                self.__comparison_chips.append({'chip': chip_object_and_name['chip'],
+                                                'name': chip_file_path})
+                self.__comparison_chips_names.append(chip_file_path)
```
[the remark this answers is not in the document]
GabrielAguirre1: I thought it would be more efficient to do it this way. I would have to write something like `[x['name'] for x in self.__comparison_chips]` on line 59. I thought that since I was already iterating through the list earlier, I might as well do the computation then. I realize this might be such a small difference that it is pointless to include, and I'm not even sure if there is a difference at all. If you prefer this other way, I will change it to that.

siliconcompiler/report/streamlit_viewer.py
GabrielAguirre1: At the moment, we do not let users supply names. Names are just 'chip1', 'chip2', ...

siliconcompiler/report/streamlit_viewer.py
GabrielAguirre1: Yes. Do I need to add this to the requirements.txt?

siliconcompiler/report/streamlit_viewer.py
GabrielAguirre1: Probably not. I will change that.

siliconcompiler/report/report.py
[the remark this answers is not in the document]
gadfort: cant you call it 10 times for each chip and just index into the matrix that is returned?

siliconcompiler/report/streamlit_viewer.py
gadfort: lets discuss

siliconcompiler/report/streamlit_viewer.py
gadfort: yes

siliconcompiler/report/report.py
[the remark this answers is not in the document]
GabrielAguirre1: I think this is probably something we should talk about.

GabrielAguirre1:
I made the name of the chip object an optional field by doing some of the parsing of the command line in sc_dashboard.py. I think there is a way to modify `dashboard_arguments`, but the solutions I found that would let me do that seemed a lot less simple. If I should look into that some more, just let me know.

siliconcompiler/apps/sc_dashboard.py
```diff
@@ -54,12 +55,24 @@ def main():
...
     graph_chips = []
     if switches['graph_cfg']:
-        for i, file_path in enumerate(switches['graph_cfg']):
+        for name_and_file_path in switches['graph_cfg']:
+            args = len(name_and_file_path)
+            if args == 0:
+                continue
+            elif args == 1:
+                file_path = name_and_file_path[0]
+                name = file_path
```
gadfort, line 64: make the name "cfgN", where N is just an increment of the position the list

siliconcompiler/apps/sc_dashboard.py
```diff
@@ -54,12 +55,24 @@ def main():
 
     graph_chips = []
     if switches['graph_cfg']:
-        for i, file_path in enumerate(switches['graph_cfg']):
+        for name_and_file_path in switches['graph_cfg']:
+            args = len(name_and_file_path)
```
gadfort, line 59: I think len will always be >1 since its a string you are parsing, you may need to split the string and then check

GabrielAguirre1: On my end, name_and_file_path is a list. where the elements are the strings separated by spaces. Maybe you should test this as well, there seems to be some difference between our systems.

siliconcompiler/apps/sc_dashboard.py
GabrielAguirre1: I don't understand what you mean. Let's discuss.

requirements.txt
```diff
@@ -26,6 +26,7 @@ streamlit; python_version >= "3.7"
 streamlit_agraph; python_version >= "3.7"
 streamlit_tree_select; python_version >= "3.7"
 streamlit_javascript; python_version >= "3.7"
+altair
```
gadfort, line 29: add version check

gadfort (approved):

PR #1797, opened 2023-07-25
add verbose flag to repair_design for better logging and update openroad minimum version to match

gadfort:


troibe (approved):
Looks good to me.

PR #1800, opened 2023-07-25
Gabriel/restructuring metrics tab

GabrielAguirre1:
Jurgen had some suggestions for the metrics tab. There were a few suggestions the main one being that the information is now stacked. There was a very small formatting change to the reports.py file (1 line), so I just threw it in here because it doesn't seem objectionable.

Here's a picture of my screeen zoomed out, but this is probably something you should run on your computer to get the full effect.
Screen Shot 2023-07-25 at 3 50 09 PM


gadfort:
@GabrielAguirre1 I like some aspects of this, lets talk tomorrow morning about this

gadfort (commented):
The layout is somewhat better, but I don't think its right.
here are some of my thoughts:
1. You need to add a heading to the files section since that is a little confusing now.
2. the metrics table is way too big and occupies my whole screen (it feels very empty, so maybe there is some padding occurring that is making it larger than it needs to be.
3. The flowgraph is somewhat awkward to access now, I'm not sure it being horizontal is the right solution vs. just spending the time to make it thinner in the vertical layout.

jldec:
Left some thoughts in https://zeroasic.slack.com/archives/C05183355GR/p1691415074204219

gadfort:
I'm going to close this since I think we need to address the tabbing layout first before we can rework the layout fully.

PR #1803, opened 2023-07-26
code refactoring

GabrielAguirre1:
I refactored the code so that it was more module and cleaner overall. This is in preparation for giving users the ability to change layouts easily. One of the functions has a suffix of 'layout_a' because I believe it will only be used for that layout, but it deserved its own function as well.


siliconcompiler/report/streamlit_viewer.py
```diff
@@ -655,118 +564,118 @@ def show_title_and_runs(title_col_width=0.7):
...
+    # concatenate metric and unit
+    metric_dataframe.index = metric_dataframe.index.map(lambda x: f'{x[0]} ({x[1]})'
+                                                        if x[1] else x[0])
+    return metric_to_metric_unit_map
+
 
-with file_viewer_tab:
-    if display_file_content:
-        show_file_viewer(new_chip, step, index)
+layout = 'A'
```
gadfort, line 643: lets give this a better name

gadfort (changes requested):
layout 'A' is not that descriptive of a name, please find something else.
Also for the functions, please add a short docstring to indicate what the module is.

gadfort (approved):
lgtm

PR #1811, opened 2023-07-28
Replace `sc-ping` CLI app with a more featureful `sc-remote`

WRansohoff:
This PR implements an updated CLI companion for the remote workflow. It has similar functionality to the current `sc-ping` app, but it can also:

* Work with unauthenticated client/server connections
* Return basic version info about the SC / server software running on the remote side.
* Cancel running jobs.
* Delete completed jobs.
* Re-connect to running or completed jobs.

The latter functionality will let us continue to interact with remote jobs if we ctrl+C out of the run, reboot our machines, temporarily lose network connectivity, etc.


siliconcompiler/apps/sc_remote.py
```diff
@@ -0,0 +1,113 @@
...
+    # Argument Parser
+    parser = argparse.ArgumentParser(prog=progname,
+                                     formatter_class=argparse.RawDescriptionHelpFormatter,
+                                     description=description)
+    parser.add_argument('-jobid', required=False)
+    parser.add_argument('-cfg', required=False)
+    parser.add_argument('-reconnect', action='store_true', required=False)
+    parser.add_argument('-cancel', action='store_true', required=False)
+    parser.add_argument('-delete', action='store_true', required=False)
+    parser.add_argument('-credentials', nargs='?', default=default_credentials_file())
```
gadfort, line 36: I think we should use the SC command line call so it can handle the loading of the CFG and credentials. also nice to have the same banner printed for the sc-apps

siliconcompiler/apps/sc_remote.py
```diff
@@ -0,0 +1,113 @@
...
+
+    args = parser.parse_args()
+
+    # Sanity checks.
+    if (args.reconnect and (args.cancel or args.delete)):
+        print('Error: -reconnect is mutually exclusive to -cancel and -delete')
```
gadfort, line 42: use chip.logger.error

siliconcompiler/apps/sc_remote.py
```diff
@@ -0,0 +1,113 @@
...
+        return 1
+    elif (args.cancel and (args.reconnect or args.delete)):
+        print('Error: -cancel is mutually exclusive to -reconnect and -delete')
+        return 1
+    elif (args.delete and (args.reconnect or args.cancel)):
+        print('Error: -delete is mutually exclusive to -reconnect and -cancel')
+        return 1
```
gadfort, line 49: this could probably be simplified to just check of any two are True, you cannot reach the `delete and ...` since it would have been caught earlier.

siliconcompiler/apps/sc_remote.py
```diff
@@ -0,0 +1,113 @@
...
+    if os.path.isfile(args.credentials):
+        with open(args.credentials, 'r') as cfgf:
+            try:
+                remote_cfg = json.loads(cfgf.read())
+            except json.JSONDecodeError:
+                print('Error reading remote configuration file: invalid JSON')
+                return 1
+    else:
+        # TODO: I think this default is stored somewhere - client.py? _metadata.py?
+        remote_cfg = {'address': 'https://server.siliconcompiler.com',
```
gadfort, line 69: it's in _metadata, please use the client to retrieve this information or rely on it to add the `remote_cfg` and `jobhash` to the chip. This will simpliyf removing the `status` dict later

siliconcompiler/apps/sc_remote.py
```diff
@@ -0,0 +1,113 @@
...
+    # If no job-related options are specified, fetch and report basic info.
+    # Create temporary Chip object and check on the server.
+    chip = Chip('server_test')
+    chip.status['remote_cfg'] = remote_cfg
+    chip.status['jobhash'] = args.jobid
+    remote_ping(chip)
+
+    # If the -cancel flag is specified, cancel the job.
+    if (args.jobid and args.cancel):
+        cancel_job(chip)
```
gadfort, line 82: you already checked for jobid, no need to do that again here. same below

siliconcompiler/apps/sc_remote.py
```diff
@@ -0,0 +1,113 @@
...
+    # in its "check_progress/ until job is done" loop.
+    elif (args.jobid and args.reconnect):
+        # TODO: Will require optional '-cfg' argument, we can't reconnect
+        # without the job's manifest. If we update the server to return the
+        # design name in the 'check_progress/' response, we could instead accept
+        # optional '-builddir' and '-jobname' arguments, then get the design
+        # and node names from a call to 'check_progress/'.
+        # Also, total runtime value will be incorrect; maybe we can have the
+        # server return the job's "created_at" time in the check_progress/ response.
+        remote_run_loop(chip, time.monotonic())
```
gadfort, line 98: Lets not print the wrong time.
Not really sure what you mean by optionally -cfg?

siliconcompiler/remote/client.py
```diff
@@ -570,18 +597,45 @@ def post_action(url):
...
+        time_remaining = user_info["compute_time"] / 60.0
+        bandwidth_remaining = user_info["bandwidth_kb"]
+        chip.logger.info(f'  Remaining compute time: {(time_remaining):.2f} minutes')
+        chip.logger.info(f'  Remaining results bandwidth: {bandwidth_remaining} KiB\n')
+
+    # Print status value.
+    server_status = response_info['status']
+    chip.logger.info(f'Server status: {server_status}')
+    if server_status != 'ready':
+        chip.logger.info('  Status is not "ready", server may not be able to accept new jobs.')
```
gadfort, line 626: is "may not" the right phrase, if it's not ready, it's not receiving jobs?

siliconcompiler/remote/server.py
```diff
@@ -300,8 +301,16 @@ async def handle_check_user(self, request):
...
-            'bandwidth_kb': self.user_keys[username]['bandwidth']
+            'user_info': {
+                'compute_time': self.user_keys[username]['compute_time'],
+                'bandwidth_kb': self.user_keys[username]['bandwidth'],
+            },
+            'status': 'ready',
+            'versions': {
+                'SC': sc_version,
+                'SCSchema': sc_schema_version,
+                'SCServer': '0.0.0',  # (Test dev server)
```
gadfort, line 312: this version would be the same as the `sc_version`

tests/apps/test_sc_remote.py
```diff
@@ -1,12 +1,12 @@
...
+from siliconcompiler.apps import sc_remote
 import json
 import os
 import subprocess
 import time
 
 
 ###########################
-def test_sc_ping(monkeypatch, unused_tcp_port):
+def test_sc_remote(monkeypatch, unused_tcp_port):
```
gadfort, line 9: need to expand the tests here to ensure coverage of the new options

gadfort (changes requested):
lets settle on the server response API.
please add testing to ensure these options and features work as expected.

siliconcompiler/apps/sc_remote.py
WRansohoff: Can we add extra parameters with `chip.create_cmdline`? The schema doesn't have equivalents of `-reconnect`, `-cancel`, `-delete`.

siliconcompiler/apps/sc_remote.py
WRansohoff: We can't "reconnect" to a remote job without knowing its design name, build directory, flowgraph nodes, etc. So until we add a local mapping of job IDs to build directories, we need `-cfg` in addition to `-jobid` to resume a remote job.

siliconcompiler/remote/client.py
```diff
@@ -444,6 +453,24 @@ def success_action(resp):
...
+    '''
+
+    def post_action(url):
+        return requests.post(url,
+                             data=json.dumps(__build_post_params(chip,
+                                                                 job_hash=chip.status['jobhash'])),
+                             timeout=__timeout)
+
+    def success_action(resp):
+        return resp.text
```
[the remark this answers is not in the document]
WRansohoff: It is - in this case, the caller would `json.loads(cancel_job(jobid))` if it wanted to inspect the response.

siliconcompiler/apps/sc_remote.py
gadfort: yes, https://docs.siliconcompiler.com/en/latest/reference_manual/core_api.html#siliconcompiler.core.Chip.create_cmdline

siliconcompiler/remote/client.py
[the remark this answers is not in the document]
gadfort: return the dict instead of the text then.

siliconcompiler/apps/sc_remote.py
WRansohoff: This won't actually impact the output - we don't print the total time in the progress check, just the runtime for each node which is reported by the server.

siliconcompiler/remote/server.py
```diff
@@ -83,7 +84,7 @@ def __init__(self, cmdlinecfg, loglevel="DEBUG"):
         self.app.add_routes([
             web.post('/remote_run/', self.handle_remote_run),
             web.post('/check_progress/', self.handle_check_progress),
-            web.post('/check_user/', self.handle_check_user),
```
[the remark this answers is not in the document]
WRansohoff: It's not used anymore - I'll remove it from the server.

gadfort (approved):

PR #1814, opened 2023-07-29
fix typo in switch flag in schema

gadfort:


troibe (approved):
LGTM
Does the `SCHEMA_VERSION` also need to be adjusted in the places where it is used?

gadfort:
I'm not sure I follow? where is `SCHEMA_VERSION` used? it's more or less just a version record

PR #1815, opened 2023-07-31
Add docs scripts to ci

troibe:
Currently we don't test the examples in our user guide.
This adds them to our CI.

The following examples were omitted because they are not actually used anywhere in the docs:
- [challenge_nda.py](https://github.com/siliconcompiler/siliconcompiler/blob/main/docs/user_guide/examples/challenge_nda.py)
- [challenge_nsquared.py](https://github.com/siliconcompiler/siliconcompiler/blob/main/docs/user_guide/examples/challenge_nsquared.py)
- [pattern_forkjoin.py](https://github.com/siliconcompiler/siliconcompiler/blob/main/docs/user_guide/examples/pattern_forkjoin.py)
- [pattern_general.py](https://github.com/siliconcompiler/siliconcompiler/blob/main/docs/user_guide/examples/pattern_general.py)
- [pattern_pipeline.py](https://github.com/siliconcompiler/siliconcompiler/blob/main/docs/user_guide/examples/pattern_pipeline.py)


tests/docs/conftest.py
```diff
@@ -0,0 +1,31 @@
...
+        ex_dir = os.path.join(scroot, 'docs', 'user_guide', 'examples')
+
+        # pytest's monkeypatch lets us modify sys.path for this test only.
+        monkeypatch.syspath_prepend(ex_dir)
+        # Add test dir to SCPATH to ensure relative paths resolve.
+        monkeypatch.setenv('SCPATH', ex_dir, prepend=os.pathsep)
+
+        return ex_dir
+
+    return setup
```
gadfort, line 31: is this needed? couldn't you just do the monkeypatch'ing here?

tests/docs/test_flowgraph_doe.py
```diff
@@ -0,0 +1,14 @@
+import os
+
+import pytest
+
+
+@pytest.mark.eda
```
gadfort, line 6: does this require any EDA tools?

tests/docs/test_write_manifest.py
```diff
@@ -0,0 +1,13 @@
+import os
+
+import pytest
+
+
+@pytest.mark.eda
```
gadfort, line 6: eda is not needed for this one

gadfort (changes requested):
I don't think you need to return the fixture setup and can just make the changes there. Check the EDA flag, if the run needs tooling then we need to mark it that way, otherwise that mark is not needed.

gadfort:
I think the unused ones are used to build some images that are used in presentations, etc, maybe you can move them to a different directory for image generation so its clear what they are.

troibe:
> I think the unused ones are used to build some images that are used in presentations, etc, maybe you can move them to a different directory for image generation so its clear what they are.

Oh ok I just saw that. It's creates the images in the `docs/user_guide/_images` folder.

gadfort (approved):
LGTM

PR #1818, opened 2023-08-01
Write manifest on task failure

troibe:
Just wanted to get some initial feedback on how to exhaustively make sure the manifest is written.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3523,6 +3523,8 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                     version = parse_version(proc.stdout)
                 except Exception as e:
                     self.logger.error(f'{tool} failed to parse version string: {proc.stdout}')
+                    # Why is this not handled with self._haltstep()?
+                    # Where does this bubble up to?
                     raise e
 
                 self.logger.info(f"Tool '{exe_base}' found with version '{version}' "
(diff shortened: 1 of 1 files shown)


troibe:
Especially in the lines where I put a comment and potentially also in other places I'm not quite sure about the error handling.
This of course needs to be really really well understood for me to makes this comprehensive.

siliconcompiler/core.py
```diff
@@ -3523,6 +3523,8 @@ def _runtask(self, step, index, status, replay=False):
                     version = parse_version(proc.stdout)
                 except Exception as e:
                     self.logger.error(f'{tool} failed to parse version string: {proc.stdout}')
+                    # Why is this not handled with self._haltstep()?
+                    # Where does this bubble up to?
```
nmoroze, line 3527: The goal with this (and the other block you commented on) is to bubble up exceptions that happen in setup module Python code (e.g. tool driver `parse_version` or `post_process`). Since these might be user-supplied, it's useful to bubble up exceptions to facilitate them debugging. They end up bubbling up all the way to the interpreter, nothing else catches these.

nmoroze: I think it would be reasonable to dump the manifest in these cases, and trying to unify them to use `_haltstep()` seems reasonable. Perhaps a bit hacky, but if we could recover and display the stacktrace without re-raising the exception, that could be a possible solution.

PR #1824, opened 2023-08-04
Catch back up to main on SC-567

petergrossmann21:
There are conflicts; need to investigate.


gadfort (approved):
just update the schema version to your version and correct the defaults.json file

PR #1828, opened 2023-08-05
gate pin placement refinement if there are unplaced instances

gadfort:
Fixes:
- pin placement generates very dense pin placement when the design is not placed, this avoids this when there are unplaced instances in the design.


troibe (approved):
LGTM
This seems to reduce the "Missing route to pin" warnings by one (64 -> 63) which is why `test_gcd.py` fails.
Looks like that is the effect you wanted so you can merge it after adjusting the test.

PR #1831, opened 2023-08-07
add klayout_allow_missing_cell to klayout driver to allow to missing cells in stream files

gadfort:
Adds capability for libraries to specify cells which are not present in the provided GDS/OAS files


troibe (approved):
Looks good but why would one want ths option?

gadfort:
Usually you would not want this, but sometimes the IP may not have the GDS and it will be instantiated by the foundry

PR #1833, opened 2023-08-07
ensure rotation is an int

gadfort:


troibe (approved):
LGTM
What type was it before?

gadfort:
It's a float, so you would end up with something like "R0.0", when it should have been R0

PR #1835, opened 2023-08-07
Update get_results error to allow single nodes to fail without interrupting the whole job

WRansohoff:
We should treat failures to fetch remote results for individual flowgraph nodes as warnings, so that they don't prematurely fail out of the job run.

Individual nodes could fail in a way that is not fatal for the entire job, for example in a branching flowgraph. Interrupting the job when a result cannot be fetched could also interfere with ongoing downloads for other nodes which did complete. Finally, SC is already set up to report build failures after the client attempts to fetch all of its results.

### DIFF
--- a/siliconcompiler/remote/client.py
+++ b/siliconcompiler/remote/client.py
@@ -515,8 +515,10 @@ def fetch_results_request(chip, node, results_path):
             return 0
 
         def error_action(code, msg):
-            chip.logger.warning(f'Error fetching results for node: {node}')
-            return 1
+            # Results are fetched in parallel, and a failure in one node
+            # does not necessarily mean that the whole job failed.
+            chip.logger.warning(f'Could not fetch results for node: {node}')
+            return 0
 
         return __post(chip,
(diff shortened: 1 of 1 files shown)


gadfort:
What happens afterwards? Shouldn't we continue to attempt to get this or does this indicate the data is no longer on the server?

WRansohoff:
Well, we didn't used to have an `error_action` for the `get_results` call. Barring catastrophic server errors, an error on the `get_results` API call indicates that the results were not found.

So the job should keep running normally, and allow any important node failures to be reported by SC. Again, we could easily run into a situation where `floorplan12` fails, `place12` does not run, and the overall job completes successfully. We could also encounter a fatal failure partway through the run, which should be picked up by SC at the end of the run rather than crashing and preventing other parallel `get_results` calls from going through.

gadfort:
I think this is fine, but it feels like we should be logging the nodes that finished on the remote but failed to download so we can report that to the user at the end.

gadfort (approved):

PR #1836, opened 2023-08-07
add klayout units to driver to allow for control of units

gadfort:


troibe (approved):
Looks good
Why is layer 0 the right layer for the units?
Is that some sort of convention?

PR #1842, opened 2023-08-08
Log job IDs for remote jobs to a file

WRansohoff:
It probably isn't reasonable to expect that people will remember or write down the UUID which gets printed at a start of a job run.

Previously, the ID wasn't something that the user needed to worry about. Now that we're adding a way to interact with remote jobs asynchronously though, it seems like it would be a good idea to store the ID somewhere that users can find it later.

For a longer-term solution, maybe in the next release, we could also consider stashing the ID somewhere in the schema, and using `-cfg` instead of `-jobid` as an argument to the `sc-remote` CLI app.


gadfort:
Lets discuss this tomorrow. I don't know if writing it to a file is going to be that helpful if there is no indication that the information is there.

gadfort:
Closing in favor to making dedicated schema key to store this information

PR #1846, opened 2023-08-08
Gabriel/code_cleanup

GabrielAguirre1:
I fixed variable names for constants, made all the expander selecting tools say 'Apply' on them per Jurgen's suggestion for making them consistent, and found, and solved a bug in the graphs module.


siliconcompiler/report/streamlit_viewer.py
```diff
@@ -19,35 +19,31 @@
...
-           "About": "\n\n".join(sc_about)}
+           "About": "\n\n".join(SC_ABOUT)}
 
 # opened by running command in siliconcompiler/apps/sc_dashboard.py
-parser = argparse.ArgumentParser('dashboard')
-parser.add_argument('cfg', nargs='?')
-args = parser.parse_args()
+PARSER = argparse.ArgumentParser('dashboard')
+PARSER.add_argument('cfg', nargs='?')
+ARGS = PARSER.parse_args()
```
gadfort, line 40: these should not be all caps
The all caps is meant for static information

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -335,7 +326,7 @@ def manifest_module(chip, manifest, ui_width, max_num_of_keys_to_show=20,
                               help='Click here to see the JSON before it was made more readable'):
             manifest_to_show = chip.schema.cfg
         else:
-            manifest_to_show = manifest
+            manifest_to_show = MANIFEST
```
gadfort, line 329: where does this come from?

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -401,7 +391,7 @@ def metrics_dataframe_module(metric_dataframe):
     display_to_data = {}
     display_options = []
     for metric_unit in metrics_list:
-        metric = metric_to_metric_unit_map[metric_unit]
+        metric = METRIC_TO_METRIC_UNIT_MAP[metric_unit]
```
gadfort, line 394: why all caps?

siliconcompiler/report/streamlit_viewer.py
```diff
@@ -801,15 +792,15 @@ def graphs_module(metric_dataframe):
         metric_dataframe (pandas.DataFrame) : A dataframe full of all metrics and all
             nodes of the selected chip
     '''
-    metrics = metric_dataframe.index.map(lambda x: metric_to_metric_unit_map[x])
+    metrics = metric_dataframe.index.map(lambda x: METRIC_TO_METRIC_UNIT_MAP[x])
```
gadfort, line 795: why all caps?

gadfort (changes requested):
I like the change to Apply that seems like it makes better sense.
some of the all caps I think should have stayed lower case.

siliconcompiler/report/streamlit_viewer.py
GabrielAguirre1: Yeah, I was going back and forth on this. I realize the argument that the variable is not static from run to run, but it is static for the duration of any given run. I'm seeing that `PARSER` is updated in line 39, so that would not be static given either definition, but `ARGS` and some of the other variables I capitalized are. To me, the second argument makes more sense, but I can change it back if that's not standard practice.

siliconcompiler/report/streamlit_viewer.py
GabrielAguirre1: It is set in the beginning and remains unchanged, so I made it a constant that is set at the beginning of running the dashboard.

siliconcompiler/report/streamlit_viewer.py
gadfort: To me all caps implies it's a constant for everything (see the strings and color defs you have), these are not constant and should not be caps

siliconcompiler/report/streamlit_viewer.py
gadfort: lower case

PR #1847, opened 2023-08-08
Add disable write cache option to Surelog

nmoroze:
Adds an option to Surelog that fixes the crash that Wenting reported.

### DIFF
--- a/siliconcompiler/tools/surelog/surelog.py
+++ b/siliconcompiler/tools/surelog/surelog.py
@@ -52,6 +52,11 @@ def setup(chip):
     if lowmem == ['true']:
         options.append('-lowmem')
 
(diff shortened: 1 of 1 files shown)


gadfort:
@nmoroze do we need the cache? it seems like we should default to true?

nmoroze:
If my reading of the docs is correct, then doing `-nowritecache` as the default is fine since we're supplying `-nocache` anyways. However, I think my understanding is pretty shaky and I was being wary of changing default behavior. I'm open to changing it and seeing what happens if you'd prefer that, though.

gadfort (approved):
@nmoroze I suppose this okay, please work with Wenting to ensure his setup is using this flag.

PR #1855, opened 2023-08-09
update contributor guidelines

gadfort:


CONTRIBUTING.md
```diff
@@ -116,41 +116,32 @@ $ git push -u origin your-branch-name
 ## Running the tests
 
 - Run the basic test suite with pytest.
+```sh
+$ pytest -m "not eda"
+```
+- This runs the tests for the current environment, without invoking any tools.
 
+- Run all basic test suite with pytest. Note that this will require all tools to be installed for the tests to pass.
```
nmoroze, line 124: Are there more tests not encapsulated by this ("basic" as opposed to)? Would it make more sense to say "Run entire test suite"?

nmoroze (approved):
small language nit, but otherwise looks good!

PR #1864, opened 2023-08-14
fix sc-show to find gz'd files and load tools when setting up common show tools

gadfort:
allow show to find gz'd files and preload the modules for show into Chip to ensure they are present if flowgraph does not contain them

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -170,6 +170,10 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         return None
 
     ###########################################################################
+    def _get_loaded_modules(self):
+        return self.modules
+
+    ###########################################################################
     def _get_tool_task(self, step, index, flow=None):
         '''
         Helper function to get the name of the tool and task associated with a given step/index.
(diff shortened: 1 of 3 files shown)


siliconcompiler/core.py
```diff
@@ -169,6 +169,10 @@ def _load_module(self, module_name, raise_error=False):
 
         return None
 
+    ###########################################################################
+    def _get_loaded_modules(self):
```
troibe, line 173: What is the benefit of this method?
Is `chip.modules` supposed to be private?
Also are they definitely "loaded" at any point when calling it?

siliconcompiler/core.py
```diff
@@ -4350,6 +4361,8 @@ def show(self, filename=None, screenshot=False, extension=None):
         self.set('option', 'hash', False, clobber=True)
         self.set('option', 'nodisplay', False, clobber=True)
         self.set('option', 'flowcontinue', True, clobber=True)
+        self.set('option', 'steplist', [], clobber=True)
```
troibe, line 4364: How are these changes related to the gzipped files or loaded tools?

troibe (approved):
The gzip part looks good.
The common show tools part gets around explicitely setting a tool in the flowgraph right?

siliconcompiler/core.py
gadfort: At some point we need to go through and make these things private.

PR #1867, opened 2023-08-15
Set -nodisplay for ssh

troibe:
**What?**
Automatically enable `-nodisplay` for ssh or other configurations without display on linux.

**Why?**
On ssh it's one extra step to enable the nodisplay option. So this improves usability.

**How?**
We check the the `$DISPLAY` and `$WAYLAND_DISPLAY` variables on linux.

**Test:**
Checks that nodisplay is set on linux with x11 or wayland.
Additionally checks there are no regressions on other platforms.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3924,14 +3924,14 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         self.write_manifest(filepath)
 
     ###########################################################################
-    def _ssh_nodisplay(self):
+    def _check_display(self):
         '''
-        Automatically disable display on ssh for Linux systems
+        Automatically disable display for Linux systems without desktop environment
         '''
         if not self.get('option', 'nodisplay') and sys.platform == 'linux' \
                 and 'DISPLAY' not in os.environ and 'WAYLAND_DISPLAY' not in os.environ:
             self.logger.warning('Environment variable $DISPLAY or $WAYLAND_DISPLAY not set')
-            self.logger.warning('Running with option -nodisplay enabled')
+            self.logger.warning("Setting ['option', 'nodisplay'] to True")
             self.set('option', 'nodisplay', True)
 
     ###########################################################################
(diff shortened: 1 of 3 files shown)


siliconcompiler/core.py
```diff
@@ -3923,6 +3923,17 @@ def _finalize_run(self, steplist, environment, status={}):
         filepath = os.path.join(self._getworkdir(), f"{self.get('design')}.pkg.json")
         self.write_manifest(filepath)
 
+    ###########################################################################
+    def _ssh_nodisplay(self):
```
gadfort, line 3927: maybe call this `_check_display()`

siliconcompiler/core.py
```diff
@@ -3923,6 +3923,17 @@ def _finalize_run(self, steplist, environment, status={}):
...
+        '''
+        Automatically disable display on ssh for Linux systems
+        '''
+        if not self.get('option', 'nodisplay') and sys.platform == 'linux' \
+                and 'DISPLAY' not in os.environ and 'WAYLAND_DISPLAY' not in os.environ:
+            self.logger.warning('Environment variable $DISPLAY or $WAYLAND_DISPLAY not set')
+            self.logger.warning('Running with option -nodisplay enabled')
```
gadfort, line 3934: lets be consistent with the key phrasing and and say something like:
```
Setting ['option', 'nodisplay'] to True
```

siliconcompiler/core.py
```diff
@@ -3923,6 +3923,17 @@ def _finalize_run(self, steplist, environment, status={}):
         filepath = os.path.join(self._getworkdir(), f"{self.get('design')}.pkg.json")
         self.write_manifest(filepath)
 
+    ###########################################################################
+    def _ssh_nodisplay(self):
+        '''
+        Automatically disable display on ssh for Linux systems
```
gadfort, line 3929: it's not only ssh, just when the environment is not setup.

tests/core/test_ssh_nodisplay.py
```diff
@@ -0,0 +1,72 @@
+import siliconcompiler
```
gadfort, line 1: rename file to something like, test_check_display

gadfort (changes requested):
Looks good, I think you just need to rename the functions a bit.

PR #1869, opened 2023-08-15
Track all possible Slurm job states

WRansohoff:
We have seen some intermittent failures in the Slurm tests, which appeared to be caused by the scheduler module encountering unexpected states when it looks at the job queue.

This change adds two more non-error states which are uncommon, but may be reported during normal Slurm runs.

### DIFF
--- a/siliconcompiler/scheduler.py
+++ b/siliconcompiler/scheduler.py
@@ -136,7 +136,8 @@ def _deferstep(chip, step, index, status):
         jobout = jobcheck.stdout.decode()
 
         # With native autoscaling, job can be 'PENDING', 'CONFIGURING', or 'COMPLETING'
-        # while the scale-up/down scripts are running.
+        # while the scale-up/down scripts are running. List of possible job states:
+        # https://slurm.schedmd.com/squeue.html#SECTION_JOB-STATE-CODES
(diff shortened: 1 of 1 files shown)


siliconcompiler/scheduler.py
```diff
@@ -140,7 +140,9 @@ def _deferstep(chip, step, index, status):
         if 'RUNNING' in jobout or \
            'PENDING' in jobout or \
            'CONFIGURING' in jobout or \
-           'COMPLETING' in jobout:
+           'COMPLETING' in jobout or \
+           'SIGNALING' in jobout or \
+           'STAGE_OUT' in jobout:
```
gadfort, line 145: is there a definitive list of these somewhere? if so, lets add a link to it and make sure all are accounted for.
Also the watchdog timer is gone, so we can remove that code too.

gadfort (changes requested):

gadfort (changes requested):
It seems like there are a lot more code slurm can report, I think it would be prudent to add them so we dont' have to do this one by one

WRansohoff:
Good point - I've updated the logic to track all possible Slurm job states, split into 'active' and 'inactive' categories.

gadfort (approved):

PR #1870, opened 2023-08-15
Add blackbox support and fix place in OpenROAD

gadfort:
Adds:
- blackbox read support for yosys

Changes:
- adds type checking to tcl escape code, need to revisit to correct escaping
- corrects the instance placement in OpenROAD to avoid faulty `place_cell` command.

### DIFF
--- a/siliconcompiler/tools/openroad/scripts/sc_floorplan.tcl
+++ b/siliconcompiler/tools/openroad/scripts/sc_floorplan.tcl
@@ -302,7 +302,17 @@ if {[dict exists $sc_cfg constraint component]} {
 
     set inst [[ord::get_db_block] findInst $name]
     if { $inst == "NULL" } {
-      utl::error FLW 1 "Could not find instance: $name"
+      utl::warn FLW 1 "Could not find instance: $name"
+
+      if { $cell == "" } {
+        utl::error FLW 1 "Unable to create instance for $name as the cell has not been specified"
+      } else {
+        set master [ord::get_db findMaster $cell]
+        if { $master == "NULL" } {
+          utl::error FLW 1 "Unable to create $name, $cell is not a valid type"
+        }
+        set inst [odb::dbInst_create [ord::get_db_block] $master $name]
+      }
     }
     set master [$inst getMaster]
     set height [ord::dbu_to_microns [$master getHeight]]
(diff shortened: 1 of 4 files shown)


siliconcompiler/tools/openroad/scripts/sc_floorplan.tcl
```diff
@@ -302,7 +302,17 @@ if {[dict exists $sc_cfg constraint component]} {
...
+      utl::warn FLW 1 "Could not find instance: $name"
+
+      if { $cell == "" } {
+        utl::error FLW 1 "Unable to create instance for $name as the cell has not been specified"
+      } else {
+        set master [ord::get_db findMaster $cell]
+        if { $master == "NULL" } {
+          utl::error FLW 1 "Unable to create $name, $cell is not a valid type"
+        }
+        set inst [odb::dbInst_create [ord::get_db_block] $master $name]
```
troibe, line 314: Just a general question:
I guess the instance is just an instantiation of a specific cell but
What is the master?

siliconcompiler/tools/openroad/scripts/sc_floorplan.tcl
```diff
@@ -314,16 +324,9 @@ if {[dict exists $sc_cfg constraint component]} {
...
-    if { $cell != "" } {
-      lappend place_args "-cell" $cell
-    }
-
-    place_cell -inst_name $name \
-      -origin "$x_loc $y_loc" \
-      -orient [$transform_final getOrient] \
-      -status FIRM \
-      {*}$place_args
+    $inst setOrient [$transform_final getOrient]
```
troibe, line 327: So before we placed a cell but now we place an instance of said cell?
So cells don't have to be instanciated first?

troibe (approved):
Looks good as far as I can tell.
So the usecase for the blackbox files are cells where we don't get the internal structure from the library provider?

siliconcompiler/tools/openroad/scripts/sc_floorplan.tcl
gadfort: the master is the cell type (its what it is called in ODB)

gadfort:
> What is currently the issue with escaping?

The tcl code when escaping is "renaming" the component, I'm still debugging that to see if there is a way around it

siliconcompiler/tools/openroad/scripts/sc_floorplan.tcl
gadfort: the place_cell code doesn't work right, so I'm juts bypassing it.

PR #1872, opened 2023-08-15
fix failure tests to pass based on new exit logic

gadfort:
fixes two failing CI tests based on the new exit quickly logic
also adds them to the quick list since they take almost no time to run.


troibe (approved):
Were these part of the daily tests or how come we missed them?

gadfort:
They were in the daily tests and not the regular tool tests. That's why it was missed.

PR #1876, opened 2023-08-17
Add note to getting started for proper execution

gadfort:


troibe (approved):
Confirmed it locally for me.
Is the double tab definitely the right indent?

gadfort:
> Confirmed it locally for me.
> Is the double tab definitely the right indent?

In the docs it becomes a single indent

PR #1878, opened 2023-08-17
remove extra logger messages

gadfort:


WRansohoff:
Looks good to me.

I think the mocked-up aiohttp response object needs to be updated to include a callable `.json` member for the failing tests to pass, if you're going to call that method in the client.

gadfort:
@WRansohoff yes you are correct. I just changed them to return response objects to avoid the type differences

WRansohoff (approved):
Ah, good idea using a `requests.Response`. I originally tried to use the same `aiohttp` responses that the server returns, but those only work in async blocks.

PR #1880, opened 2023-08-17
add missing version requirement to altair

gadfort:


troibe (approved):
That one is a bit weird to me put if it's really required for correct dependency solution then it's fine with me.

PR #1881, opened 2023-08-17
ensure surelog marks required keys

gadfort:
This is needed to properly create a testcase for surelog


troibe (approved):
Looks good.
Restructuring and adding the required keys are the chagnes I'm seeing.

PR #1882, opened 2023-08-17
add : cntrl+c capturing to remote run 

gadfort:
Allows a user to do Control+C to exit out of SC and gives the required information to resume the connection

### DIFF
--- a/siliconcompiler/remote/client.py
+++ b/siliconcompiler/remote/client.py
@@ -10,6 +10,7 @@ import urllib.parse
 import tarfile
 import tempfile
 import multiprocessing
+import sys
 
 from siliconcompiler._metadata import default_server
 from siliconcompiler import utils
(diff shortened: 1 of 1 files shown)


siliconcompiler/remote/client.py
```diff
@@ -274,7 +275,26 @@ def remote_run(chip):
...
+            chip._get_flowgraph_entry_nodes(flow=chip.get('option', 'flow'))[0]
+        entry_manifest = os.path.join(chip._getworkdir(step=entry_step, index=entry_index),
+                                      'outputs',
+                                      f'{chip.design}.pkg.json')
+        reconnect_cmd = f'sc-remote -jobid {jobid} -cfg {entry_manifest} -reconnect'
+        cancel_cmd = f'sc-remote -jobid {jobid} -cancel'
+        chip.logger.info('Disconnecting from remote job')
+        chip.logger.info(f'To reconnect to this job use: {reconnect_cmd}')
+        chip.logger.info(f'To cancel this job use: {cancel_cmd}')
+        sys.exit(0)
```
WRansohoff, line 293: One reason we usually raise `SiliconCompilerError` instead of calling `sys.exit(1)` is that `sys.exit` can't be caught with blanket `except:` blocks. So, if somebody calls `chip.run()` in a script which has post-run logic, this exit will prevent most of that logic from running unless they specifically catch the `except SystemExit:` case.

That's probably fine, but how would you feel about `raise SiliconCompilerError("Job canceled by user keyboard interrupt")`?

WRansohoff (approved):
We should be submitting the logic to move the job ID into the schema next week, but this looks like a good idea.

Just one comment which I could go either way on.

siliconcompiler/remote/client.py
gadfort: it's a good point. The stacktrace is kind of messy but I can try it out and see what it ends up looking like

PR #1884, opened 2023-08-17
add _remove to schema object to help remove default keys

gadfort:
This is useful when removing previously set constraints like components or timing corners.

Only allows removal of user definable keys (ie. default keys)


nmoroze (approved):
LGTM. I feel like I wouldn't be opposed to making this part of the public API, but that could be done later.

PR #1888, opened 2023-08-18
wrap tarfile.read with try/except in the event of error

gadfort:


gadfort:
@WRansohoff just checking in on this

WRansohoff (approved):
I made a similar change in the `('option', 'resume')` PR draft that we ended up punting on. The function should return in the event of an extraction error, and do the `os.remove` call in a `finally:` block so that it doesn't stick around on the user's system.

Should be good to go with the fixes I just committed, once we see that the CI tests pass.

WRansohoff (changes requested):
Don't call 'return' in the 'finally' block - that will prevent the following logic from running when the extraction succeeds.

gadfort:
I removed the finally block since calling return prevents finally from executing

WRansohoff:
I don't think that's right - try running this in a REPL:

```
def abc():
   print('a')
   try:
     raise ValueError('b')
   except:
     return
   finally:
     print('b')
```

Note that both print statement execute. `finally:` blocks get run before the exception logic.

WRansohoff (approved):
Thanks - I know the syntax is a little confusing, a block called "finally" sounds like it would happen after everything else. It's a useful feature to keep in mind for cleanup logic, though.

PR #1896, opened 2023-08-21
deprecate support for python3.6 and python3.7 as they are EOL

gadfort:
This removes support for 3.6 and 3.7.
Changes:
- removes wheels builds for those versions
- updates docs to use minimum version of 3.8

### DIFF
--- a/.github/workflows/daily_ci.yml
+++ b/.github/workflows/daily_ci.yml
@@ -29,7 +29,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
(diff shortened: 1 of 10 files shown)


gadfort:
@nmoroze @WRansohoff are there any code we should revisit now that 3.6 and 3.7 are not going to supported?
The `copytree` comes to mind, but if you think there are other pleases, just let me know so we can get that code cleaned up if possible/needed.

nmoroze:
Nothing much comes to mind right now, but will update if I think of anything else.

troibe (approved):
Looks good to me. I don't see how this could break anything on our side.
Just make sure every user is aware of this.

nmoroze (approved):

PR #1897, opened 2023-08-21
ensure allkeys works when called on a leaf

gadfort:


nmoroze (approved):

nmoroze (commented):
Might be good to add a test, but looks good to me!

PR #1901, opened 2023-08-23
Add 'resume' functionality to remote runs

WRansohoff:
Since the server runs remote jobs using the core `chip.run()` call, we only need some minor client-side changes to make the remote flow work with `('option', 'resume')`.

This change will let us resume partially-completed jobs on a server, functionality which has been helpful with large local jobs in the past. It also cleans up some logging when the server returns empty task archives, which it will do for any completed tasks that get uploaded in order to resume a job.

The feature works by re-submitting completed tasks from a local build directory, so users will only be able to resume jobs on servers which return the full `outputs/` directory for each flowgraph node. Otherwise, the local build directories won't contain the processed design files which are needed by steps that come later in the flowgraph.

Most of the changes are just indentation. The main `remote_preprocess` modifications are in lines 123-133.

### DIFF
--- a/siliconcompiler/remote/client.py
+++ b/siliconcompiler/remote/client.py
@@ -120,44 +120,55 @@ def remote_preprocess(chip, steplist):
     flow = chip.get('option', 'flow')
     remote_steplist = steplist.copy()
     entry_steps = chip._get_flowgraph_entry_nodes(flow=flow)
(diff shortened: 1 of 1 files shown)


WRansohoff:
Closing this in favor of a more complex proposal which will let the server resume runs without needing to return their full results to the client after every step.

PR #1902, opened 2023-08-24
Move job IDs into the core Schema

WRansohoff:
Like we've discussed offline, the job ID values could easily be moved into the core schema, instead of using the ephemeral `chip.status` dictionary, which we would like to get rid of eventually.

This PR implements the client-side changes required to do that. I bumped the schema's minor version number instead of its patch number, because v0.35.0 manifests will not work with servers that are using the new version for remote runs.

I also updated the `sc-remote` CLI app to replace its `-jobid` CLI parameter with `-cfg`, which was already required for reconnecting to a running job. That required a little bit of extra logic to insert the job ID values into the manifests produced by the locally-run entry steps, because the client doesn't know what its job ID will be until it submits a job and receives a response from the server.

### DIFF
--- a/siliconcompiler/remote/client.py
+++ b/siliconcompiler/remote/client.py
@@ -11,9 +11,9 @@ import tarfile
 import tempfile
 import multiprocessing
 
+from siliconcompiler import utils, SiliconCompilerError
 from siliconcompiler._metadata import default_server
-from siliconcompiler import utils
-from siliconcompiler import SiliconCompilerError
+from siliconcompiler.schema import Schema
 
 
 # Client / server timeout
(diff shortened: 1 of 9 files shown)


WRansohoff:
Marking this as a draft until I finalize the new schema parameter, mainly whether it should have pernode set to `never` or `optional`.

gadfort:
The jobid should be under record, we just need to fix the places where we assume that it's always by step/index

siliconcompiler/remote/client.py
```diff
@@ -355,6 +354,27 @@ def check_progress(chip):
...
+def update_entry_manifests(chip):
+    '''
+    Helper method to update locally-run manifests to include remote job ID.
+    '''
+
+    flow = chip.get('option', 'flow')
+    jobid = chip.get('record', 'jobid')
+    design = chip.get('design')
+
+    entry_nodes = chip._get_flowgraph_entry_nodes(flow=flow)
```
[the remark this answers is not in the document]
WRansohoff: Yeah, I like the idea of having one `[design].pkg.json` manifest in the root build directory which we keep up to date as the job progresses. It shouldn't be too complicated to implement, we just need to be careful to avoid race conditions with parallel flowgraphs.

siliconcompiler/remote/client.py
```diff
@@ -355,6 +354,27 @@ def check_progress(chip):
...
 
 
+###################################
+def _update_entry_manifests(chip):
+    '''
+    Helper method to update locally-run manifests to include remote job ID.
+    '''
+
+    flow = chip.get('option', 'flow')
+    jobid = chip.get('record', 'remoteid')
```
troibe, line 364: Having `jobid`, `remoteid` and `job_hash` gets a bit confusing to me at this point.

siliconcompiler/remote/client.py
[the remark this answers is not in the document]
troibe: Sure. Created SC-612 for it.

troibe:
This looks good to me.
Just one tiny comment on the naming.
Also make sure things are merged in in the right order between server and client...

siliconcompiler/remote/client.py
gadfort: at some point we should go through and ensure the naming is consistent (remoteid) is the correct term now

gadfort (approved):

siliconcompiler/remote/client.py
troibe: SC-615

PR #1909, opened 2023-08-27
improve chisel support

gadfort:
Adds:
- support to take in a build.sbt and provide arguments to the runMain command


nmoroze (approved):
Looks good, 2 small pieces of feedback:

- Not super sure about `input, hll, chisel` for `sbt` files, since they're really a build configuration file, and this feels a little like treating it as a design source. I do like using `chip.input()` for it though, so I'm not really sure of a great alternative (maybe replacing `hll` with something like `config`?)
- I'm curious if the assumption about `src/` being next to `build.sbt` always holds for Chisel/scala? If not, would be good to document. Might also be good to document the fact that the first/a single SBT file gets read from `input, hll, chisel` only.

gadfort:
the directory structure is: https://www.scala-sbt.org/1.x/docs/Directories.html

PR #1912, opened 2023-08-28
add option to get report images with screenshot

gadfort:


troibe (approved):
Looks good to me.
Now that I think about it is there some nice way to integrate tcl scripts with the python tests we have?

gadfort:
@troibe what do you mean about integrating tcl with python?

troibe:
> @troibe what do you mean about integrating tcl with python?

As in adding a unit test that checks if this feature calls this tcl function.

troibe:
Added SC-611

PR #1916, opened 2023-08-29
Fix normalization of list to tuples for list-of-tuples

nmoroze:
This PR fixes a bug where setting a single "scalar" value for a list-of-tuples field using a list rather than a tuple is considered invalid. There's a bit of ambiguity here when parsing the type/value from the "outside-in", so I think the easiest solution without restructuring this code is to speculatively consider the value in each possible way and see which works.

### DIFF
--- a/siliconcompiler/schema/schema_obj.py
+++ b/siliconcompiler/schema/schema_obj.py
@@ -524,9 +524,19 @@ class Schema:
     @staticmethod
     def _normalize_value(value, sc_type, error_msg, allowed_values):
         if sc_type.startswith('['):
-            if not isinstance(value, list):
-                value = [value]
             base_type = sc_type[1:-1]
+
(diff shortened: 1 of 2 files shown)


gadfort (approved):
LGTM

troibe (commented):
The logic looks good to me.
I just don't really understand why everything is compared to a list of tuples in the test case...
Probably I'm missing something there.

nmoroze:
> The logic looks good to me. I just don't really understand why everything is compared to a list of tuples in the test case... Probably I'm missing something there.

What's going on is that `set()/add()` provide some flexibility in how values are provided, but the values always get normalized before being stored. So there are a few legal ways to `set()` the keypath in the test, but querying the value back always gives a list-of-tuples, since that's its datatype.

PR #1917, opened 2023-08-30
Split up run()

troibe:
**What?**
Refactor of `run()` no functionality changes.
Also solves SC-586

**Why?**
SC-557, SC-562 and others are difficult to do since `_runtask()` and `run()` are huge function in which changes could easy break something important.

**How?**
We split up `run()` to make editing and testing the steps of the flow easier.

**Test**
So far just the existing test cases.
I could add further ones if necessary.
[Successful run on remote server](https://github.com/zeroasiccorp/scserver/actions/runs/6023567892/job/16340572454)

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3960,6 +3960,25 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
             self.logger.warning("Setting ['option', 'nodisplay'] to True")
             self.set('option', 'nodisplay', True)
 
+    def _increment_jobs(self):
+        '''
+        Auto-update jobname if ['option', 'jobincr'] is True
+        Do this before initializing logger so that it picks up correct jobname
+        '''
+        if self.get('option', 'jobincr'):
+            workdir = self._getworkdir()
+            if os.path.isdir(workdir):
+                # Strip off digits following jobname, if any
+                stem = self.get('option', 'jobname').rstrip('0123456789')
+
+                designdir = os.path.dirname(workdir)
+                jobid = 0
+                for job in os.listdir(designdir):
+                    m = re.match(stem + r'(\d+)', job)
+                    if m:
+                        jobid = max(jobid, int(m.group(1)))
+                self.set('option', 'jobname', f'{stem}{jobid + 1}')
+
     def _filter_steplist(self):
         # Run steps if set, otherwise run whole graph
         if self.get('arg', 'step'):
@@ -4268,21 +4287,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                 self.error(f"{key} must be set before calling run()",
                            fatal=True)
 
-        # Auto-update jobname if ['option', 'jobincr'] is True
-        # Do this before initializing logger so that it picks up correct jobname
-        if self.get('option', 'jobincr'):
-            workdir = self._getworkdir()
-            if os.path.isdir(workdir):
-                # Strip off digits following jobname, if any
-                stem = self.get('option', 'jobname').rstrip('0123456789')
-
-                designdir = os.path.dirname(workdir)
-                jobid = 0
-                for job in os.listdir(designdir):
-                    m = re.match(stem + r'(\d+)', job)
-                    if m:
-                        jobid = max(jobid, int(m.group(1)))
-                self.set('option', 'jobname', f'{stem}{jobid + 1}')
+        self._increment_jobs()
 
         # Re-init logger to include run info after setting up flowgraph.
         self._init_logger(in_run=True)
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3864,13 +3864,13 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                 os.remove(path)
 
     ###########################################################################
-    def _setup_task(self, step, index):
+    def _setup_node(self, step, index):
 
         self.set('arg', 'step', step)
         self.set('arg', 'index', index)
         tool, task = self._get_tool_task(step, index)
 
-        # Run task setup.
+        # Run node setup.
         try:
             setup_step = getattr(self._get_task_module(step, index), 'setup', None)
         except SiliconCompilerError:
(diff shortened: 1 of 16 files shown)


siliconcompiler/core.py
```diff
@@ -3960,47 +3960,11 @@ def _check_display(self):
             self.logger.warning("Setting ['option', 'nodisplay'] to True")
             self.set('option', 'nodisplay', True)
 
-    ###########################################################################
-    def run(self):
+    def _increment_jobs(self):
```
gadfort, line 3963: increment_job_name

siliconcompiler/core.py
```diff
@@ -4015,20 +3979,12 @@ def run(self):
...
-        # Re-init logger to include run info after setting up flowgraph.
-        self._init_logger(in_run=True)
-
-        # Check if flowgraph is complete and valid
-        flow = self.get('option', 'flow')
-        if not self._check_flowgraph(flow=flow):
-            self.error(f"{flow} flowgraph contains errors and cannot be run.",
-                       fatal=True)
-
+    def _filter_steplist(self):
```
gadfort, line 3982: _get_flow_steplist (or something like that)

siliconcompiler/core.py
```diff
@@ -4039,9 +3995,14 @@ def run(self):
                 if os.path.isdir(cur_job_dir):
                     shutil.rmtree(cur_job_dir)
 
-        # List of indices to run per step. Precomputing this ensures we won't
-        # have any problems if [arg, index] gets clobbered, and reduces logic
-        # repetition.
+            return steplist
+
+    def _precompute_indexlist(self, steplist, flow):
```
gadfort, line 4000: _get_flow_indexlist (or something like that)

siliconcompiler/core.py
```diff
@@ -4090,210 +4053,286 @@ def run(self):
...
-        for envvar in self.getkeys('option', 'env'):
-            val = self.get('option', 'env', envvar)
-            os.environ[envvar] = val
+    def _load_remote_config(self):
+        '''
+        Load the remote storage config into the status dictionary.
+        '''
+        if self.get('option', 'credentials'):
+            # Use the provided remote credentials file.
+            cfg_file = os.path.abspath(self.get('option', 'credentials'))
```
gadfort, line 4062: I think all the code related to running the remote flow should go into the client.py and we should just call that and pass in the chip object

siliconcompiler/core.py
```diff
@@ -4090,210 +4053,286 @@ def run(self):
...
+
+        # Delete the job's data from the server.
+        delete_job(self)
+        # Restore logger
+        self._init_logger(in_run=True)
+        # Restore steplist
+        if pre_remote_steplist['set']:
+            self.set('option', 'steplist', pre_remote_steplist['steplist'])
+
+    def _prepare_tasks(self, tasks_to_run, processes, flow, status, steplist, indexlist):
```
gadfort, line 4119: might be nice to simplify the call since steplist and indexlist can be computed quickly.

siliconcompiler/core.py
```diff
@@ -4050,7 +4011,9 @@ def run(self):
                 indexlist[step] = self.get("option", 'indexlist')
             else:
                 indexlist[step] = self.getkeys('flowgraph', flow, step)
+        return indexlist
 
+    def _resume_steps(self, flow, steplist, indexlist):
```
gadfort, line 4016: rename this to make it clear, it's doing the preprocessing for the run. 
Also steplist and index list can be gotten from the helper methods and just makes this signature simpler

siliconcompiler/core.py
```diff
@@ -4090,210 +4053,286 @@ def run(self):
...
-            tasks_to_run = {}
-            processes = {}
-            # Ensure we use spawn for multiprocessing so loggers initialized correctly
-            multiprocessor = multiprocessing.get_context('spawn')
-            for step in steplist:
-                for index in indexlist[step]:
-                    nodename = f'{step}{index}'
-                    if status[nodename] != TaskStatus.PENDING:
-                        continue
+    def _remote_process(self, steplist):
```
gadfort, line 4087: move to client.py

siliconcompiler/core.py
gadfort, line 4119: I think this next set of three should be renamed: prepare_task, check_node_dependencies, and lanch_task makes it hard to follow the naming, you are preparing a node, so lets keep that naming pattern at least.

gadfort (changes requested):
I think where we can avoid passing information around we should do so and just use the helper methods to to get it.
The method names in general make it unclear what their purpose is, so in most cases they need to renamed.

siliconcompiler/core.py
troibe: I think that was part of another refactor Noah suggested as well.
Did that now.
Should I also rename `TaskStatus` to `Node Status`?

siliconcompiler/core.py
troibe: Oh ok on a first glance it wasn't obvious to me that that was the purpose.

siliconcompiler/core.py
troibe: I was not doing this on purpose since there were comments saying indexlist or steplist might get clobbered.
Is that really an issue?

siliconcompiler/core.py
troibe: What do you think of `_reset_flow_nodes`?
Same thing as above. Can steplist and indexlist change between function calls?

siliconcompiler/core.py
gadfort: I don't know, but the merging logic is a mess, so I suppose it's possible, although it shouldn't be.
Lets keep it until we can verify if it's an issue.

siliconcompiler/core.py
gadfort: _reset_flow_nodes is fine

siliconcompiler/core.py
gadfort: NodeStatus is fine

gadfort:
@troibe looks good, just need to resolve the conflicts

troibe:
> @troibe looks good, just need to resolve the conflicts

Done now! Merging this stuff is horrible though.

gadfort:
I think the refactor will help deal with the merging, now that things are broken up, we can move it out of the core when we have time

gadfort (approved):

troibe:
[New run on remote server](https://github.com/zeroasiccorp/scserver/actions/runs/6051635945/job/16423522422)

PR #1918, opened 2023-08-30
Remove skipstep option

troibe:
Unless I'm missing something major there is no place where this option is actually used in the run logic.
Part of SC-613


troibe:
Do I need to update the schema version in any way for this?

gadfort:
If you need to update the last json then you need to change the 0.x.0 if you didn't, you can just increment the patch version

troibe:
Ok incremented patch version without changing `last major` schema.

gadfort (approved):

PR #1919, opened 2023-08-30
switch default for lef generation to bloating occupied layers

gadfort:


troibe (approved):
LGTM
Why the change?

gadfort:
It's the wrong setting

PR #1921, opened 2023-08-31
Reuse docker image

troibe:
**What?**
This commit introduces a github action that takes the tool as an input and returns the latest version of its docker image.

**Why?**
We currently copy and paste the code to retrieve the latest version of the `sc_tools` images.
This is also the case for sclib, zerosoc and scscerver.

**Successful run:**
https://github.com/siliconcompiler/siliconcompiler/actions/runs/6039399625

### DIFF
--- a/.github/workflows/docker_image.yml
+++ b/.github/workflows/docker_image.yml
@@ -12,7 +12,7 @@ on:
         default: 'tools'
         type: string
       sc_version:
-        description: 'SC Version tag'
+        description: 'SC version tag'
         required: false
         type: string
     outputs:
--- a/.github/workflows/docker_image.yml
+++ b/.github/workflows/docker_image.yml
@@ -11,6 +11,10 @@ on:
         required: false
         default: 'tools'
         type: string
+      sc_version:
+        description: 'SC Version tag'
+        required: false
+        type: string
     outputs:
       sc_tool:
         description: 'Name and newest digest for the docker image of the input tool'
@@ -33,6 +37,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: siliconcompiler/siliconcompiler
+          ref: ${{ inputs.sc_version }}
 
       - name: Get newest docker image for the input tool
         id: docker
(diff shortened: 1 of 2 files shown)


.github/workflows/docker_image.yml
```diff
@@ -0,0 +1,41 @@
...
+  workflow_call:
+    inputs:
+      tool:
+        description: 'Name of the tool or "tools" for all tools'
+        required: true
+        type: string
+    outputs:
+      sc_tool:
+        description: 'Name and newest digest for the docker image of the input tool'
+        value: ${{ jobs.docker_image.outputs.sc_tool }}
```
gadfort, line 16: can you add an input for sc version as the repos can control that via a key?

troibe: Sure added that now.
Didn't knwo the `builder.py` script gives you the latest version for each tool when using an older commit of sc as well.

.github/workflows/docker_image.yml
```diff
@@ -0,0 +1,47 @@
...
+  workflow_call:
+    inputs:
+      tool:
+        description: 'Name of the tool'
+        required: false
+        default: 'tools'
+        type: string
+      sc_version:
+        description: 'SC Version tag'
+        required: false
```
gadfort, line 16: does this need a default value of 'main'?

.github/workflows/docker_image.yml
```diff
@@ -0,0 +1,47 @@
...
+on:
+  workflow_call:
+    inputs:
+      tool:
+        description: 'Name of the tool'
+        required: false
+        default: 'tools'
+        type: string
+      sc_version:
+        description: 'SC Version tag'
```
gadfort, line 15: SC version tag

.github/workflows/docker_image.yml
troibe: No runs successfully like that.

.github/workflows/docker_image.yml
troibe: Woops in

gadfort (approved):

PR #1928, opened 2023-09-01
fix handling of booleans and single value parameters which are defined on a pernode basis

gadfort:
Fixes:
- when specifying multiple breakpoints only one was selected because it's defined on a pernode basis and not a list.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -411,6 +411,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
             # Fetch fields from leaf cell
             helpstr = schema.get(*keypath, field='shorthelp')
             typestr = schema.get(*keypath, field='type')
+            pernodestr = schema.get(*keypath, field='pernode')
 
             # argparse 'dest' must be a string, so join keypath with commas
             dest = '_'.join(keypath)
@@ -421,16 +422,25 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
             if not switchlist or any(switch in switchlist for switch in switchstrs):
                 used_switches.update(switchstrs)
                 if typestr == 'bool':
-                    parser.add_argument(*switchstrs,
-                                        nargs='?',
-                                        metavar=metavar,
-                                        dest=dest,
-                                        const='true',
-                                        help=helpstr,
-                                        default=argparse.SUPPRESS)
-                # list type arguments
-                elif re.match(r'\[', typestr):
-                    # all the rest
+                    # Boolean type arguments
+                    if pernodestr == 'never':
+                        parser.add_argument(*switchstrs,
+                                            nargs='?',
+                                            metavar=metavar,
+                                            dest=dest,
+                                            const='true',
+                                            help=helpstr,
+                                            default=argparse.SUPPRESS)
+                    else:
+                        parser.add_argument(*switchstrs,
+                                            metavar=metavar,
+                                            nargs='?',
+                                            dest=dest,
+                                            action='append',
+                                            help=helpstr,
+                                            default=argparse.SUPPRESS)
+                elif re.match(r'\[', typestr) or pernodestr != 'never':
+                    # list type arguments
                     parser.add_argument(*switchstrs,
                                         metavar=metavar,
                                         dest=dest,
@@ -462,13 +472,13 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
 
         # Iterate from index 1, otherwise we end up with script name as a
         # 'source' positional argument
-        for item in sys.argv[1:]:
+        for argument in sys.argv[1:]:
             # Split switches with one character and a number after (O0,O1,O2)
-            opt = re.match(r'(\-\w)(\d+)', item)
+            opt = re.match(r'(\-\w)(\d+)', argument)
             # Split assign switches (-DCFG_ASIC=1)
-            assign = re.search(r'(\-\w)(\w+\=\w+)', item)
+            assign = re.search(r'(\-\w)(\w+\=\w+)', argument)
             # Split plusargs (+incdir+/path)
-            plusarg = re.search(r'(\+\w+\+)(.*)', item)
+            plusarg = re.search(r'(\+\w+\+)(.*)', argument)
             if opt:
                 scargs.append(opt.group(1))
                 scargs.append(opt.group(2))
@@ -479,7 +489,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                 scargs.append(assign.group(1))
                 scargs.append(assign.group(2))
             else:
-                scargs.append(item)
+                scargs.append(argument)
 
         parser.add_argument('-version', action='version', version=_metadata.version)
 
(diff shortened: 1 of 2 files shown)


siliconcompiler/core.py
```diff
@@ -559,6 +577,7 @@ def create_cmdline(self,
                 args = [free_keys.pop(0) if key == 'default' else key for key in keypath]
 
                 # Remainder is the value we want to set, possibly with a step/index value beforehand
+                sctype = self.get(*keypath, field='type')
                 pernode = self.get(*keypath, field='pernode')
                 step, index = None, None
                 if pernode == 'required':
```
troibe, line 583: This one is also unexpected to me.
I also couldn't find the possible values of `pernode` in the documentation.

siliconcompiler/core.py
```diff
@@ -513,7 +523,11 @@ def create_cmdline(self,
 
         # Set loglevel if set at command line
         if 'option_loglevel' in cmdargs.keys():
-            self.logger.setLevel(cmdargs['option_loglevel'])
+            log_level = cmdargs['option_loglevel']
+            if isinstance(log_level, list):
```
troibe, line 527: Is this in case multiple nodes set the loglevel?

siliconcompiler/core.py
```diff
@@ -513,7 +523,11 @@ def create_cmdline(self,
...
+                # if multiple found, pick the first one
+                log_level = log_level[0]
+            self.logger.setLevel(log_level.split()[-1])
```
troibe, line 530: Why do we have to split this?
I thought loglevel allowed values are:
`"NOTSET", "INFO", "DEBUG", "WARNING", "ERROR", "CRITICAL"`

siliconcompiler/core.py
```diff
@@ -421,16 +422,25 @@ def create_cmdline(self,
...
+                                            default=argparse.SUPPRESS)
+                    else:
+                        parser.add_argument(*switchstrs,
+                                            metavar=metavar,
+                                            nargs='?',
+                                            dest=dest,
+                                            action='append',
+                                            help=helpstr,
+                                            default=argparse.SUPPRESS)
+                elif re.match(r'\[', typestr) or pernodestr != 'never':
```
troibe, line 442: Why `or` instead of `and`?

siliconcompiler/core.py
```diff
@@ -421,16 +422,25 @@ def create_cmdline(self,
...
-                                        metavar=metavar,
-                                        dest=dest,
-                                        const='true',
-                                        help=helpstr,
-                                        default=argparse.SUPPRESS)
-                # list type arguments
-                elif re.match(r'\[', typestr):
-                    # all the rest
+                    # Boolean type arguments
+                    if pernodestr == 'never':
```
troibe, line 426: Is this definitely the value for setting `-pernode` to false?

siliconcompiler/core.py
gadfort: looks like that is missing the docs, I'll add it

siliconcompiler/core.py
gadfort: yes, if you set to be different on different steps and indices we have to pick one before the full parsing is complete.

siliconcompiler/core.py
gadfort: if set on a pernode basis the value might look like: `syn DEBUG` or `place 0 INFO`

siliconcompiler/core.py
gadfort: because both need a list behavior

siliconcompiler/core.py
gadfort: yes

siliconcompiler/core.py
```diff
@@ -574,12 +593,28 @@ def create_cmdline(self,
...
-                    elif len(items) == 3:
-                        step, index, val = items
-                    elif len(items) == 2:
-                        step, val = items
+                    if sctype == 'bool':
+                        if len(items) == 3:
+                            step, index, val = items
+                        elif len(items) == 2:
+                            step, val = items
+                            if val != 'true' and val != 'false':
```
troibe, line 601: Is index being checked if it's legal at this point?

siliconcompiler/core.py
```diff
@@ -574,12 +593,28 @@ def create_cmdline(self,
...
+                                index = val
+                                val = True
+                        elif len(items) == 1:
+                            val, = items
+                            if val != 'true' and val != 'false':
+                                step = val
```
troibe, line 607: Is there a check if this step exists at this point?

siliconcompiler/core.py
```diff
@@ -537,6 +551,10 @@ def create_cmdline(self,
 
             # Cycle through all items
             for item in vals:
+                if item is None:
```
troibe, line 554: (Why) Is `item = ''` preferred over `item = None`?

siliconcompiler/core.py
troibe: So the typestr would be not starting with `\[` for an option that is set for multiple nodes?

siliconcompiler/core.py
gadfort: no, there is no way to check that until `run()`

siliconcompiler/core.py
gadfort: not until `run()`

siliconcompiler/core.py
gadfort: because later we blindly call item.split() and if that is a `None` it raises an exception

siliconcompiler/core.py
troibe: Does the actual logging later switch between different granularity levels?
If not we should probably add a little warning that we just picked the first.

siliconcompiler/core.py
gadfort: `re.match(r'\[', typestr)` this is checking if it's a list of some kind (indicating we need to allow lists), the second is checking if `pernode` is used and if it is, we need to allow multiple values as well

siliconcompiler/core.py
troibe: Then does the logging actually switch between different granularity levels per index?

siliconcompiler/core.py
troibe: Ok if it's checked at that point that is also fine.

siliconcompiler/core.py
troibe: Ok if it's checked at that point that is also fine.

siliconcompiler/core.py
gadfort: yes, the logger is reinitialized between nodes to honor these.
This will be corrected later, when we set the loglevel below, this is only the logging level while parsing the values, I'm inclined to not add additional outputs about it. If anything we could pick the highest and use that but I don't think that would be for this PR

siliconcompiler/core.py
troibe: Ok I guess.

siliconcompiler/core.py
gadfort: it should

gadfort: if not, then thats a bug

siliconcompiler/core.py
troibe: Mixed this up as I didn't realise this is still talking about a list in the arguments...Not the list we later get when retrieving an option.

troibe (approved):
LGTM
Sorry for the many comments...Took some time to understand.

gadfort:
@troibe no problem, some of the logic probably needs some improvement to make it easier to follow as well.

PR #1933, opened 2023-09-02
Fix fpga docs and make cli switches a list in the docs 

gadfort:


siliconcompiler/sphinx_ext/dynamicgen.py
```diff
@@ -615,7 +615,7 @@ def document_task(self, path, toolmodule, taskmodule, taskname, toolname):
         # Annotate the target used for default values
         if chip.valid('option', 'target') and chip.get('option', 'target'):
             p = docutils.nodes.inline('')
-            target = chip.get('option', 'target')
+            target = chip.get('option', 'target').split('.')[-1]
```
troibe, line 618: Why was this required?

gadfort: The target gets recorded as the python module name, but the reference just needs the module name

troibe:
I found the differences but I don't quite understand why it was failing to generate?

Before:
image
image
image
image
image
image

After:
image
image
image
image
image
image

troibe (approved):

gadfort:
It fails if the fpga flow is not configured correctly.

PR #1934, opened 2023-09-03
Small changes to enable resuming remote job runs.

WRansohoff:
Client-side changes that we discussed offline to make the `('option', 'resume')` parameter work with remote jobs. Most of the changes are server-side.

* Remove post-job `delete_job` call. It has been replaced by `sc-remote -cfg [...] -delete`, and a pruning job which deletes jobs older than a threshold every few hours.
* Add remote ID to the `remote_run/` call if it was pulled in via an existing manifest.
* Avoid uploading job data when resuming a job.
* Avoid overwriting the remote ID parameter when resuming a job. (I'm open to suggestions if anybody has a better idea for this; `if record != 'remoteid':` felt too hack-y)

This change (and the associated serverside modifications) will also let us send partially-completed local jobs to a server for finishing, with:

* `('option', 'resume') = True`
* `('option', 'remote') = True`
* `('record', 'remoteid') = None`


siliconcompiler/core.py
```diff
@@ -4034,7 +4034,9 @@ def _reset_flow_nodes(self, flow, steplist, indexlist):
                     for metric in self.getkeys('metric'):
                         self._clear_metric(step, index, metric)
                     for record in self.getkeys('record'):
-                        self._clear_record(step, index, record)
+                        # Job-wide record values may need to be preserved across runs.
+                        if self.get('record', record, field='pernode') != 'required':
```
troibe, line 4038: Are there other records than the one required for `resume` that might be set to `required`?

troibe:
With my upcoming changes where the steplist gets replaced with from/to it would be great if you could add some test case(s).
Either here or over in scserver.

siliconcompiler/remote/client.py
```diff
@@ -446,11 +456,14 @@ def _request_remote_run(chip):
     Helper method to make a web request to start a job stage.
     '''
 
-    upload_file = tempfile.TemporaryFile(prefix='sc', suffix='remote.tar.gz')
-    with tarfile.open(fileobj=upload_file, mode='w:gz') as tar:
-        tar.add(chip._getworkdir(), arcname='')
-    # Flush file to ensure everything is written
-    upload_file.flush()
+    should_resume = (chip.get('option', 'resume') and chip.get('record', 'remoteid'))
```
troibe, line 459: Maybe using `remote_resume` would be more consistent

siliconcompiler/core.py
WRansohoff: I don't think so, but making a one-off exception for that record felt sort of messy. I'm definitely open to other ways of avoiding clearing this record if you have ideas.

siliconcompiler/remote/client.py
```diff
@@ -446,11 +456,14 @@ def _request_remote_run(chip):
...
+    # Only package and upload the entry steps if starting a new job.
```
troibe, line 460: starting a new remote run or resuming a local run remotely

siliconcompiler/remote/client.py
```diff
@@ -476,21 +489,26 @@ def _request_remote_run(chip):
     # part of the HTTP spec, so we need to manually follow the trail.
     post_params = {
         'chip_cfg': chip.schema.cfg,
-        'params': __build_post_params(chip)
+        'params': __build_post_params(chip,
+                                      job_hash=chip.get('record', 'remoteid'))
```
troibe, line 493: Is there a reason to keep `job_hash` still around?
Instead of renaming that to `remotid` as well?

troibe (changes requested):
LGTM besides the testing and a couple of minor things

siliconcompiler/core.py
gadfort: maybe make an optional parameter to `_clear_record` to preserve the remoteid?

siliconcompiler/remote/client.py
```diff
@@ -136,24 +136,25 @@ def _remote_preprocess(chip, steplist):
...
-        run_task = multiprocessor.Process(target=chip._runtask,
-                                          args=(local_step, index, {}))
-        run_task.start()
-        run_task.join()
-        if run_task.exitcode != 0:
-            # A 'None' or nonzero value indicates that the Process target failed.
-            ftask = f'{local_step}{index}'
-            chip.error(f"Could not start remote job: local setup task {ftask} failed.",
-                       fatal=True)
+        if not chip.get('option', 'resume'):
```
gadfort, line 139: what happens if resume is on, but the remoteid is not set?

tests/flows/test_resume.py
```diff
@@ -36,3 +37,45 @@ def test_resume(gcd_chip):
     # Ensure flow finished successfully
     assert gcd_chip.find_result('def', step='place') is not None
     assert gcd_chip.find_result('gds', step='export') is not None
+
+@pytest.mark.eda
+@pytest.mark.timeout(600)
+def test_resume_remote(gcd_chip, unused_tcp_port):
+    '''Test that jobs can be resumed on a remote server.
+    '''
```
gadfort, line 45: move to remote tests

siliconcompiler/remote/client.py
```diff
@@ -283,27 +284,35 @@ def remote_process(chip, steplist):
...
+            steplist = cur_steplist if cur_steplist else chip.list_steps()
+        _remote_preprocess(chip, steplist)
 
     # Run the job on the remote server, and wait for it to finish.
     # Set logger to indicate remote run
     chip._init_logger(step='remote', index='0', in_run=True)
     _remote_run(chip)
 
-    # Delete the job's data from the server.
-    delete_job(chip)
```
gadfort, line 306: since we want this one quickly, could you make a separate PR with just this so we can get this in quickly.

gadfort (changes requested):

siliconcompiler/remote/client.py
WRansohoff: It will upload the current local run to be resumed on the remote server.

siliconcompiler/remote/client.py
WRansohoff: Sure.

siliconcompiler/remote/client.py
WRansohoff: We should rename it to `remoteid` - that will be a separate change, since the "job_hash" string is also baked into the API specs.

tests/flows/test_resume.py
WRansohoff: Sure, I guess it makes more sense there.

siliconcompiler/remote/client.py
WRansohoff: https://github.com/siliconcompiler/siliconcompiler/pull/1936

siliconcompiler/core.py
```diff
@@ -4996,22 +4996,33 @@ def _record_metric(self, step, index, metric, value, source, source_unit=None):
             self.add('tool', tool, 'task', task, 'report', metric, source, step=step, index=index)
 
     #######################################
-    def _clear_metric(self, step, index, metric):
+    def _clear_metric(self, step, index, metric, preserve=[]):
```
gadfort, line 4999: preserve needs to be None, putting mutable objects in the defaults for keywords is discouraged.

siliconcompiler/core.py
```diff
@@ -4996,22 +4996,33 @@ def _record_metric(self, step, index, metric, value, source, source_unit=None):
...
+
         flow = self.get('option', 'flow')
         tool, task = self._get_tool_task(step, index, flow=flow)
 
         self.unset('metric', metric, step=step, index=index)
         self.unset('tool', tool, 'task', task, 'report', metric, step=step, index=index)
 
     #######################################
-    def _clear_record(self, step, index, record):
+    def _clear_record(self, step, index, record, preserve=[]):
```
gadfort, line 5016: same

siliconcompiler/remote/client.py
gadfort: we have a ticket open to correct the naming overall

gadfort (changes requested):
LGTM, just need to fix the `[]` arguments, I'll defer to @troibe for final approval

siliconcompiler/remote/client.py
troibe: SC-615

troibe (approved):
Looks ood wit the changes.

gadfort (approved):

PR #1935, opened 2023-09-05
fix CI to patch home() to ensure OS testing works

gadfort:
https://github.com/siliconcompiler/siliconcompiler/actions/runs/6084794388/job/16507538745


troibe (approved):
Looks good

PR #1937, opened 2023-09-06
Get flowgraph nodes

troibe:
**Why?**
When working more with the flowgraph it was a bit unclear at times what is a node vs step in existing code.
Additonally this should make it more clear at which points the steplist and indexlist are considered in the flowgraph.
Last minor benefit is that it saves one level of indentation.

**What?**
Introduce a function that gets all nodes `(step, index)` in the flowgraph and use it in existing code.

**Testing:**
[Remote server run](https://github.com/zeroasiccorp/scserver/actions/runs/6096289187/job/16542015595)

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1716,7 +1716,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         # 2. Check library names
         libraries = set()
         for val, step, index in self.schema._getvals('asic', 'logiclib'):
-            if (in_step, in_index) in flowgraph_nodes:
+            if (step, index) in flowgraph_nodes:
                 libraries.update(val)
 
         for library in libraries:
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1689,7 +1689,8 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
             self.logger.error(f"flowgraph {flow} not defined.")
 
         indexlist = self.get('option', 'indexlist')
-        for (step, index) in self._get_flowgraph_nodes(flow, steplist=steplist, indexlist=indexlist):
+        flowgraph_nodes = self._get_flowgraph_nodes(flow, steplist=steplist, indexlist=indexlist)
+        for (step, index) in flowgraph_nodes:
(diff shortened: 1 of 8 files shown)


siliconcompiler/core.py
```diff
@@ -1688,43 +1688,35 @@ def check_manifest(self):
...
+                self.logger.error(f'{step}{index} relies on {in_step}{in_index}, '
+                                  'but this task has not been run and is not in the '
+                                  'current steplist.')
+                error = True
 
         # 2. Check library names
         libraries = set()
         for val, step, index in self.schema._getvals('asic', 'logiclib'):
-            if step in steplist and index in indexlist[step]:
+            if step in steplist and (not indexlist or in_index in indexlist):
```
gadfort, line 1719: why not check for `(step,index) in nodes` instead of using the index list

tests/core/test_graph.py
```diff
@@ -64,6 +64,25 @@ def test_graph():
...
+    chip.node(flow, 'premin', fake_out, index=1)
+    chip.node(flow, 'domin', minimum)
+    chip.node(flow, 'postmin', fake_in)
+
+    chip.edge(flow, 'premin', 'domin', tail_index=0)
+    chip.edge(flow, 'premin', 'domin', tail_index=1)
+    chip.edge(flow, 'domin', 'postmin')
+
+    assert chip._get_flowgraph_nodes(flow) == \
+        [('premin', '0'), ('premin', '1'), ('domin', '0'), ('postmin', '0')]
```
gadfort, line 83: this this order always true?

siliconcompiler/core.py
gadfort: it seems like the get_flowgraph_nodes will return based on the indexlist, so why not just use that where you have checks for indexlist?

troibe: Just replaced it there.
At some other points there is some logic around the lines of: `if` at least one index of this step or all indices of this step
There I didn't find a prettier replacement yet.

tests/core/test_graph.py
troibe: For now yes it is the order from the schema and it is a list so it is a fixed order.

gadfort (approved):

PR #1941, opened 2023-09-07
fix missing " in screenshot for OR

gadfort:


troibe (approved):
How did you even find this?

gadfort:
it tries to `source .../sc_write_images.tcl"` which is not a file

PR #1942, opened 2023-09-07
Flowgraph nodes remaining instances

troibe:
Cleanup from https://github.com/siliconcompiler/siliconcompiler/pull/1937
Replace any remaining uses of `self.getkeys('flowgraph',)` with `self._get_flowgraph_nodes()`.
Any now remaining instances of `self.getkeys('flowgraph',)` require the rewrite with `from/to`.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -4032,29 +4032,31 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         # manifests from _runtask() actually updates values.
         should_resume = self.get("option", 'resume')
         node_list = self._get_flowgraph_nodes(flow, steplist=steplist, indexlist=indexlist)
+        for (step, index) in self._get_flowgraph_nodes(flow):
+            stepdir = self._getworkdir(step=step, index=index)
+            cfg = f"{stepdir}/outputs/{self.get('design')}.pkg.json"
+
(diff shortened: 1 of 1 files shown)


siliconcompiler/core.py
```diff
@@ -4033,29 +4032,31 @@ def _reset_flow_nodes(self, flow, steplist, indexlist):
...
-                    self.set('flowgraph', flow, step, index, 'status', None)
-
-                    # Reset metrics and records
-                    for metric in self.getkeys('metric'):
-                        self._clear_metric(step, index, metric)
-                    for record in self.getkeys('record'):
-                        self._clear_record(step, index, record)
-                elif os.path.isfile(cfg):
-                    self.set('flowgraph', flow, step, index, 'status', NodeStatus.SUCCESS)
+                if self.get('flowgraph', flow, step, index, 'status') == NodeStatus.SUCCESS:
```
gadfort, line 4058: why is this still needed?

troibe: We are not using the flowgraph for determining the success of a step yet.
So for now I left the "all indices have to fail" assumption.

troibe: In my next pull request I'm planning to change all those spots where the `step/index` is accessed manually like this one.

gadfort (approved):

PR #1943, opened 2023-09-07
Fix client error

gadfort:


siliconcompiler/remote/client.py
```diff
@@ -665,9 +665,12 @@ def fetch_results(chip, node, results_path=None):
 
     # Note: the server should eventually delete the results as they age out (~8h), but this will
     # give us a brief period to look at failed results.
-    if results_code:
+    if not node and results_code:
```
troibe, line 668: I don't see how this would occur.
If you just want to mute it then I guess completely remove this
and just keep
```
if results_code:
    return
```

gadfort: this condition only applies if no node is specified.

siliconcompiler/remote/client.py
```diff
@@ -678,7 +681,7 @@ def fetch_results(chip, node, results_path=None):
         with tarfile.open(results_path, 'r:gz') as tar:
             tar.extractall(path=(node if node else ''))
     except tarfile.TarError as e:
-        chip.error(f'Failed to extract data from {results_path}: {e}')
+        chip.logger.error(f'Failed to extract data from {results_path}: {e}')
```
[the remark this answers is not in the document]
gadfort: future work , but yes that might be the right approach

troibe (approved):
If this is an emergency fix that actually fixes things then go ahead.
If it's just muting then:
I think either both failing to download and failing to extract should be treated as errors or as warnings.

siliconcompiler/remote/client.py
troibe: I'm just saying I don't see any call that would leave node unspecified

siliconcompiler/remote/client.py
```diff
@@ -665,9 +665,12 @@ def fetch_results(chip, node, results_path=None):
 
     # Note: the server should eventually delete the results as they age out (~8h), but this will
     # give us a brief period to look at failed results.
-    if results_code:
+    if not node and results_code:
         chip.error("Sorry, something went wrong and your job results could not be retrieved. "
```
troibe, line 669: This would still raise an exception with stacktrace...

siliconcompiler/remote/client.py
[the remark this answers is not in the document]
troibe: SC-627

siliconcompiler/remote/client.py
gadfort: yes, this just needs to be reworked once we start cleaning up the server

PR #1944, opened 2023-09-07
fix flow docs build

gadfort:
I noticed the flow was not building correctly in the docs. 
This fixes the flowgraph writing to correctly handle mismatching names and reenabled lost docs for steps

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
 extend-exclude = third_party,setup/deps,_skbuild
 max-line-length = 100
-ignore =
+extend-ignore =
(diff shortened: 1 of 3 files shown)


.flake8
```diff
@@ -1,4 +1,4 @@
 [flake8]
 extend-exclude = third_party,setup/deps,_skbuild
 max-line-length = 100
-ignore =
+extend-ignore =
```
troibe, line 4: can be removed completely?

siliconcompiler/sphinx_ext/dynamicgen.py
```diff
@@ -405,18 +405,20 @@ def display_config(self, chip, modname):
...
-                    continue
-                schema = Schema(cfg=cfg)
-                schema.prune()
-                pruned = schema.cfg
-                if prefix not in step_cfg:
-                    step_cfg[prefix] = {}
-                step_cfg[prefix][step] = pruned
-
-            section += build_schema_value_table(step_cfg, self.env.docname, skip_zero_weight=True)
+            cfg = chip.getdict('flowgraph', chip.design, step)
```
troibe, line 408: Just building the main flowgraph is enough here?
Not sure what the original intention was with building the additional flowgraphs as well...

troibe (approved):
Looks good to me.

PR #1946, opened 2023-09-08
Cleanup remote printout

gadfort:


troibe (approved):
This way the default logger still shows the trace on the commandline right?

gadfort:
if called via the CLI, only the error message is printed, not the stack trace

PR #1947, opened 2023-09-08
Add -noinfo and -nonote

troibe:


troibe:
@petergrossmann21 Do you have a nice testcase that produces both infos and notes?

siliconcompiler/tools/surelog/surelog.py
```diff
@@ -57,6 +57,14 @@ def setup(chip):
     if no_write_cache == ['true']:
         options.append('-nowritecache')
 
+    no_info = chip.get('tool', tool, 'task', task, 'var', 'disable_info', step=step, index=index)
```
gadfort, line 60: Can you set them to false by default and add a help string so these options get documented

gadfort (changes requested):

petergrossmann21:
@troibe Any HDL codebase of any complexity should generate quite a few of each.  Look especially for things in your existing test suite that contain generate statements.

siliconcompiler/tools/surelog/surelog.py
troibe: If understand the other `var` options properly then it should already be off by default.
Added the docstring.

gadfort (approved):

PR #1954, opened 2023-09-11
Save cfg on error at input merge and final merge

troibe:
Currently the only places where the manifest gets merged is when collecting the inputs on the start of a node or at the and of a run. In both these cases the manifest gets only merged on success this merges it regardless of success or failure.
This pull request saves the manifest regardless of whether the steps have failed.
@gadfort Are you sure we want to change the behaviour in this way?


troibe:
Abandoning because trying to merge manifests at the end of a node execution led to synchronization issues.
Additionally there were some issues with merging in the central manifest to per node manifests.
Some values in the schema were overwritten that should not have gotten overwritten.

Final implementation with input merge and final merge can be found here.
https://github.com/siliconcompiler/siliconcompiler/pull/1968

PR #1955, opened 2023-09-11
correct tcl in OR to find the right master cell

gadfort:


troibe (approved):
Any linting or other checks we can enable on tcl files that might help catching these kind of things?

gadfort:
I looked at tcl linting a while ago, there really isn't any any good solutions out there I could find. @nmoroze had expressed some interest in it as well. I'd be happy to add it if we can find something that works

PR #1956, opened 2023-09-11
Fix handling if libcorner in yosys and openroad to allow for multiple libcorners.

gadfort:


troibe:
Is there anything to the readthedocs fail?

siliconcompiler/tools/openroad/openroad.py
```diff
@@ -398,38 +395,39 @@ def get_metric_sources(metric):
...
+    keypaths = {}
     for constraint in chip.getkeys('constraint', 'timing'):
-        pexcorner = chip.get('constraint', 'timing', constraint, 'pexcorner',
-                             step=step, index=index)
-        if pexcorner:
-            corners.add(pexcorner)
+        corners = chip.get('constraint', 'timing', constraint, 'libcorner', step=step, index=index)
+        for corner in corners:
+            if chip.valid('library', lib, 'output', corner, delaymodel):
+                keypaths[constraint] = ('library', lib, 'output', corner, delaymodel)
```
troibe, line 408: Are the corners sorted in any way?
The last overwrite is the last/most ??? corner that is valid?

siliconcompiler/tools/openroad/openroad.py
```diff
@@ -398,38 +395,39 @@ def get_metric_sources(metric):
...
 
-    return list(corners)
+        if constraint not in keypaths:
+            keypaths[constraint] = ('library', lib, 'output', corners[0], delaymodel)
```
troibe, line 411: Didn't we already see that corners[0] leads to an invalid keypath in the for loop above?
This sets the `keypath[constraint]` to an invalid path as far as I can tell which I'm not sure is intended...

siliconcompiler/tools/openroad/openroad.py
```diff
@@ -122,20 +122,17 @@ def setup(chip):
...
             chip.add('tool', tool, 'task', task, 'require',
                      ",".join(['library', lib, 'output', stackup, 'lef']),
                      step=step, index=index)
         for lib in macrolibs:
-            for corner in corners:
-                if chip.valid('library', lib, 'output', corner, delaymodel):
-                    chip.add('tool', tool, 'task', task, 'require',
-                             ",".join(['library', lib, 'output', corner, delaymodel]),
+            for timing_key in get_library_timing_keypaths(chip, lib).values():
+                if chip.valid(*timing_key):
```
troibe, line 134: Ok I see here is another validity check.
If you remove the possiby invalid keypath from `get_library_timing_keypaths()` I think you could remove this one.
If not then you should probably also check validity for the `targetlibs`.

siliconcompiler/tools/openroad/openroad.py
```diff
@@ -438,27 +436,21 @@ def get_corner_by_check(chip, check):
                                  step=step, index=index):
             continue
 
-        libcorner = chip.get('constraint', 'timing', constraint, 'libcorner',
-                             step=step, index=index)
-        if libcorner:
-            return libcorner[0]
+        return constraint
```
troibe, line 439: This can now be rewritten to:
```suggestion
    if check in chip.get('constraint', 'timing', constraint, 'check',
                                 step=step, index=index):
        return constraint
```

siliconcompiler/tools/openroad/openroad.py
```diff
@@ -468,22 +460,20 @@ def build_pex_corners(chip):
...
     for constraint in chip.getkeys('constraint', 'timing'):
-        libcorner = chip.get('constraint', 'timing', constraint, 'libcorner',
-                             step=step, index=index)
         pexcorner = chip.get('constraint', 'timing', constraint, 'pexcorner',
                              step=step, index=index)
 
-        if not libcorner or not pexcorner:
+        if not pexcorner:
             continue
-        corners[libcorner[0]] = pexcorner
```
troibe, line 478: Is this the fundamental reason that before only one `libcorner` was possible?

siliconcompiler/tools/yosys/syn_asic.py
```diff
@@ -320,14 +310,24 @@ def get_synthesis_corner(chip):
...
+        return ('library', lib, 'option', 'file', 'yosys_synthesis_libraries')
 
+    step = chip.get('arg', 'step')
+    index = chip.get('arg', 'index')
+    delaymodel = chip.get('asic', 'delaymodel', step=step, index=index)
+    for corner in corners:
+        if chip.valid('library', lib, 'output', corner, delaymodel):
+            return ('library', lib, 'output', corner, delaymodel)
+
+    return ('library', lib, 'output', corners[0], delaymodel)
```
troibe, line 327: As in `openroad.py` returning invalid library key is fine here?

siliconcompiler/tools/yosys/syn_asic.py
```diff
@@ -303,7 +293,7 @@ def get_synthesis_corner(chip):
     syn_corner = chip.get('tool', tool, 'task', task, 'var', 'synthesis_corner',
                           step=step, index=index)
     if syn_corner:
-        return syn_corner[0]
+        return syn_corner
```
troibe, line 296: Aren't these `syn_corners`?

troibe (changes requested):
Biggest confusion to me are passing potentially invalid keypaths around.
Additionally a couple of smaller suggestions

siliconcompiler/tools/openroad/openroad.py
gadfort: not sorted, this is just ensuring that the liberty models are marked as required

siliconcompiler/tools/openroad/openroad.py
gadfort: this is to ensure a path is marked required so execution will fail on check_manifest.

siliconcompiler/tools/openroad/openroad.py
gadfort: macros don't always have a timing model, but the target libs must have one

siliconcompiler/tools/openroad/openroad.py
gadfort: it was a misunderstanding when this was implemented about the libcorners

siliconcompiler/tools/openroad/scripts/sc_apr.tcl
```diff
@@ -153,15 +152,18 @@ source "$sc_refdir/sc_procs.tcl"
 ###############################
 
 # Read Liberty
-utl::info FLW 1 "Defining timing corners: $sc_corners"
-define_corners {*}$sc_corners
+utl::info FLW 1 "Defining timing corners: $sc_scenarios"
```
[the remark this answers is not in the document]
gadfort: within the tool itself, they are called corners.
The scenarios can be different, but openroad doesn't support that properly

siliconcompiler/tools/yosys/syn_asic.py
gadfort: same

gadfort:
@troibe the invalid keypaths are on purpose to ensure we can mark something as required for SC to check

troibe (approved):
> this is to ensure a path is marked required so execution will fail on check_manifest.

I don't quite understand why it's better to fail during `check_manifest` instead of where the keypath would actually be used.

PR #1806, opened 2023-07-27
Exit early on failure for local runs

troibe:
**What?**
With these changes the local runs fail quicker.

**Why?**
It makes it easier for the user to understand which step failed.
It only displays the last step and the tool logs only up until the error occured.
Addtionally the runtime decreases since we don't invoke unnecessary tools.

**How?**
For builtin tasks we check if any of the input dependencies successfully finished.
For non-builtin in tasks we check if all of the input dependencies successfully finished.

**Before:**
```
| INFO    | parasteps_0  | floorplan  | 0  | Tool 'openroad' found with version 'v2.0-8997' in directory '/usr/local/bin'
| ERROR   | parasteps_0  | floorplan  | 0  | Version check failed for openroad. Check installation.
| ERROR   | parasteps_0  | floorplan  | 0  | Found version v2.0-8997, did not satisfy any version specifier set >=v2.0-9413.
| ERROR   | parasteps_0  | floorplan  | 0  | Halting step 'floorplan' index '0' due to errors.
| ERROR   | parasteps_0  | physyn     | 0  | Halting step due to previous error in floorplan0
| ERROR   | parasteps_0  | physyn     | 0  | Halting step 'physyn' index '0' due to errors.
| ERROR   | parasteps_0  | place      | 0  | Halting step due to previous error in physyn0
| ERROR   | parasteps_0  | place      | 0  | Halting step 'place' index '0' due to errors.
| ERROR   | parasteps_0  | place      | 1  | Halting step due to previous error in physyn0
| ERROR   | parasteps_0  | place      | 1  | Halting step 'place' index '1' due to errors.
| INFO    | parasteps_0  | placemin   | 0  | Running builtin task 'minimum'
| ERROR   | parasteps_0  | placemin   | 0  | No inputs selected after running builtin
| ERROR   | parasteps_0  | placemin   | 0  | Halting step 'placemin' index '0' due to errors.
| ERROR   | parasteps_0  | cts        | 0  | Halting step due to previous error in placemin0
| ERROR   | parasteps_0  | cts        | 0  | Halting step 'cts' index '0' due to errors.
| ERROR   | parasteps_0  | cts        | 1  | Halting step due to previous error in placemin0
| ERROR   | parasteps_0  | cts        | 1  | Halting step 'cts' index '1' due to errors.
| INFO    | parasteps_0  | ctsmin     | 0  | Running builtin task 'minimum'
| ERROR   | parasteps_0  | ctsmin     | 0  | No inputs selected after running builtin
[...]

PR #1958, opened 2023-09-12
Use node tuple instead of string

troibe:
After some other code refactoring fixing [this](https://github.com/siliconcompiler/siliconcompiler/pull/1806#discussion_r1279656792) suggestion from Noah become fairly easy.

No functional changes just readability/reliability.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -4086,15 +4086,12 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
             if status[node] != NodeStatus.PENDING:
                 continue
 
-            node_inputs = self.get('flowgraph', flow, step, index, 'input')
-            inputs = [(step, index) for step, index in node_inputs]
-
             if (self._get_in_job(step, index) != jobname):
                 # If we specify a different job as input to this task,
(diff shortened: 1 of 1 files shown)


siliconcompiler/core.py
```diff
@@ -4082,29 +4082,28 @@ def _prepare_nodes(self, nodes_to_run, processes, flow, status, steplist, indexl
...
         for (step, index) in flowgraph_nodes:
-            nodename = f'{step}{index}'
-            if status[nodename] != NodeStatus.PENDING:
+            node = (step, index)
+            if status[node] != NodeStatus.PENDING:
                 continue
 
             node_inputs = self.get('flowgraph', flow, step, index, 'input')
-            inputs = [f'{step}{index}' for step, index in node_inputs]
+            inputs = [(step, index) for step, index in node_inputs]
```
gadfort, line 4090: is this needed? seems like a nop

gadfort (commented):
LGTM

siliconcompiler/core.py
troibe: Oh very true. Even better then.

gadfort (approved):

PR #1959, opened 2023-09-12
Lint - Fix unknown proc openroad_pin_access_args

troibe:
Just did an initial run of the [`ttclcheck` linter](https://wiki.tcl-lang.org/page/ttclcheck).
It's the linter used in Vivado.
Still haven't figured out completly how to set up openroad and other dependencies but I found my first real error within seconds.
Hopefully we can set it up properly and integrate it soon...


troibe:
@gadfort Do you know any better tcl shell than `tclsh`?
That one doesn't even support my arrow keys. :smile:

gadfort (approved):

troibe:
> Do you know any better tcl shell than tclsh?
That one doesn't even support my arrow keys. 😄

https://wiki.tcl-lang.org/page/tclreadline

PR #1968, opened 2023-09-15
Merge and write manifest on error

troibe:
**What?**
- Write output manifest on all steps including failures.
- Write the schema when possible
- Replace file existence as criteria for node success when resuming

**Why?**
Currently the manifest does not capture the schema after every node execution.
In particular failed nodes do not have a written manifest and won't be part of the final schema.
It also is one step further to eventually get rid of the `status` dictionary.

**How?**
We now write the manifest on non fatal errors which are captured by `_haltstep`.
This requires reading the manifest to get the node status instead of just checking if the manifest exists for resuming.
If the node does not get executed at all because its input dependencies failed then we don't write a manifest.

**Testing**
- tests/flowgraph.py
- tests/flowstatus.py
- tests/flows/test_resume.py
- **new:** tests/core/test_error_manifest.py

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -3957,7 +3957,6 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                 schema = Schema(manifest=lastcfg)
                 if schema.get('flowgraph', flow, step, index, 'status') == NodeStatus.SUCCESS:
                     stat_success = True
-            # Merge in manifest if the task was successful.
             if os.path.isfile(lastcfg):
                 self._read_manifest(lastcfg, clobber=False, partial=True)
             if stat_success:
@@ -4064,9 +4063,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                 for record in self.getkeys('record'):
                     self._clear_record(step, index, record)
             elif os.path.isfile(cfg):
-                chip_node_status = Chip('node_succeess')
-                chip_node_status._read_manifest(cfg)
-                node_status = chip_node_status.get('flowgraph', flow, step, index, 'status')
+                node_status = Schema(manifest=cfg).get('flowgraph', flow, step, index, 'status')
                 self.set('flowgraph', flow, step, index, 'status', node_status)
             else:
                 self.set('flowgraph', flow, step, index, 'status', NodeStatus.ERROR)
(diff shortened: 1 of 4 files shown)


gadfort (approved):
LGTM, probably would benefit from @nmoroze review as well before merging

siliconcompiler/core.py
```diff
@@ -3956,8 +3958,9 @@ def _finalize_run(self, steplist, environment, status={}):
                 if schema.get('flowgraph', flow, step, index, 'status') == NodeStatus.SUCCESS:
                     stat_success = True
             # Merge in manifest if the task was successful.
```
nmoroze, line 3960: Remove/change comment now that this isn't tied to task success

siliconcompiler/core.py
```diff
@@ -4061,7 +4064,10 @@ def _reset_flow_nodes(self, flow, steplist, indexlist):
                 for record in self.getkeys('record'):
                     self._clear_record(step, index, record)
             elif os.path.isfile(cfg):
-                self.set('flowgraph', flow, step, index, 'status', NodeStatus.SUCCESS)
+                chip_node_status = Chip('node_succeess')
+                chip_node_status._read_manifest(cfg)
+                node_status = chip_node_status.get('flowgraph', flow, step, index, 'status')
```
nmoroze, line 4069: Could this be done more cleanly with a `Schema` object? (bit of a nit-pick, but don't need the fake design name then for example)

siliconcompiler/core.py
```diff
@@ -3956,8 +3958,9 @@ def _finalize_run(self, steplist, environment, status={}):
                 if schema.get('flowgraph', flow, step, index, 'status') == NodeStatus.SUCCESS:
                     stat_success = True
             # Merge in manifest if the task was successful.
-            if stat_success:
+            if os.path.isfile(lastcfg):
```
nmoroze, line 3961: Should/does it ever occur that this file doesn't exist? If it shouldn't happen, might be worth an assertion or error check to make sure we don't mask a bug, rather than silently continuing.

nmoroze (approved):
Looks good, a couple small comments but feel free to merge once addressed.

siliconcompiler/core.py
troibe: Yes it can occur and is fine.
The scenario will be that an input node (or the input of the input etc) failed.
In that case we don't write an empty manifest...

siliconcompiler/core.py
troibe: Oh of course I was thinking initially that directly accessing the Schema might be discouraged or sth.

PR #1970, opened 2023-09-17
remove value key from single value keys in manifest viewer

gadfort:


troibe (approved):
I don't really understand why but sure.

gadfort:
This corrects the implementation to reflect the original plan.

PR #1973, opened 2023-09-18
speeds up .valid by simply checking the schema dictionary instead of requiring the allkeys()

gadfort:
Setup node for OpenROAD is noticeably slow. This appears to be from all the calls to `valid()`
This change speeds that up by ~178x.

Before profile:
```
         3556918 function calls (3413821 primitive calls) in 3.751 seconds
```
After profile:
```
         58632 function calls (53019 primitive calls) in 0.021 seconds
```

Also removes seemingly unused `valid_keypaths` argument from `.valid`

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -850,7 +850,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         return fullstr
 
     ###########################################################################
-    def valid(self, *keypath, valid_keypaths=None, default_valid=False):
+    def valid(self, *keypath, default_valid=False):
         """
         Checks validity of a keypath.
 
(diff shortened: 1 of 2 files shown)


siliconcompiler/core.py
```diff
@@ -850,15 +850,14 @@ def help(self, *keypath):
...
-    def valid(self, *keypath, valid_keypaths=None, default_valid=False):
+    def valid(self, *keypath, default_valid=False):
         """
         Checks validity of a keypath.
 
         Checks the validity of a parameter keypath and returns True if the
         keypath is valid and False if invalid.
 
         Args:
-            keypath(list str): Variable length schema key list.
```
nmoroze, line 861: Did you mean to remove this? (looks like the actual arg you deleted was just never documented...)

nmoroze (approved):
LGTM, great speedup. One minor note on docstring to look at, otherwise good to merge.

PR #1976, opened 2023-09-20
Replace step/indexlist with from/to

troibe:
**What?**
Instead of executing a list of steps/indices we now specify where in the flowgraph we start `-from` and up to which point `-to` we execute.

**Why?**
Using the step and indexlist we didn't actually traverse the flowgraph so some flowgraph semantic was not captured before.

**How?**
We traverse the flowgraph to identify all nodes that can be reached and the nodes that successfully executed.

**Testing**
Used existing tests so far but can add more.


siliconcompiler/remote/client.py
```diff
@@ -169,10 +170,11 @@ def _remote_preprocess(chip, steplist):
     # we need to send inputs up to the server.
     chip._collect()
 
-    # Set 'steplist' to only the remote steps, for the future server-side run.
+    # Set 'from/to' to only the remote steps, for the future server-side run.
     chip.unset('arg', 'step')
     chip.unset('arg', 'index')
-    chip.set('option', 'steplist', remote_steplist, clobber=True)
+    chip.set('option', 'from', [remote_nodelist[0][0]], clobber=True)
+    chip.set('option', 'to', [remote_nodelist[-1][0]], clobber=True)
```
gadfort, line 177: I don't think you can assume any ordering here.
You probably need to use exit nodes as the 'to' and the outputs from the entry nodes as a 'from'

siliconcompiler/remote/client.py
```diff
@@ -300,12 +302,19 @@ def remote_process(chip, steplist):
...
-        'set': chip.schema._is_set(chip.schema._search('option', 'steplist')),
+    cur_from_nodes = chip.get('option', 'from')
+    cur_to_nodes = chip.get('option', 'to')
+    pre_remote_from_nodes = {
+        'from_nodes': cur_from_nodes,
+        'set': chip.schema._is_set(chip.schema._search('option', 'from')),
+    }
+    pre_remote_to_nodes = {
+        'to_nodes': cur_to_nodes,
+        'set': chip.schema._is_set(chip.schema._search('option', 'to')),
```
gadfort, line 313: fundamentally, this needs to be fixed when we re-work the server.

siliconcompiler/scheduler.py
```diff
@@ -111,7 +111,8 @@ def _defernode(chip, step, index):
         with open(script_path, 'w') as sf:
             sf.write('#!/bin/bash\n')
             sf.write(f'sc -cfg {shlex.quote(cfg_file)} -builddir {shlex.quote(buildir)} '
-                     f'-arg_step {shlex.quote(step)} -arg_index {shlex.quote(index)}\n')
+                     f'-arg_step {shlex.quote(step)} -arg_index {shlex.quote(index)} '
+                     f'-from {shlex.quote(step)} -to {shlex.quote(step)}\n')
```
gadfort, line 115: why is this needed, if we provide `arg_step` and `arg_index`

siliconcompiler/schema/schema_cfg.py
```diff
@@ -2461,28 +2461,29 @@ def schema_option(cfg):
             and per index basis. During execution, the default behavior is to
             copy inputs from the current job.""")
 
-    scparam(cfg, ['option', 'steplist'],
+    scparam(cfg, ['option', 'from'],
             sctype='[str]',
             scope='job',
-            shorthelp="Compilation step list",
-            switch="-steplist <step>",
+            shorthelp="Compilation steps to start from",
```
gadfort, line 2467: start flowgraph execution from

siliconcompiler/schema/schema_cfg.py
```diff
@@ -2461,28 +2461,29 @@ def schema_option(cfg):
...
-                     "api: chip.set('option', 'indexlist', '0')"],
+            shorthelp="Compilation steps to end with",
+            switch="-to <step>",
+            example=[
+                "cli: -to 'syn'",
+                "api: chip.set('option', 'to', 'syn')"],
             schelp="""
-            List of indices to execute. The default is to execute all
-            indices for each step of a run.""")
+            List of steps to end execution with. The default is to go
```
gadfort, line 2485: clarify that this is inclusive

tests/apps/test_sc.py
```diff
@@ -11,7 +11,7 @@ def test_design_inference(scroot, monkeypatch):
     source = os.path.join(scroot, 'tests', 'data', 'heartbeat.v')
     # only run import, makes this quicker
 
-    monkeypatch.setattr('sys.argv', ['sc', source, '-steplist', 'import', '-strict'])
+    monkeypatch.setattr('sys.argv', ['sc', source, '-to', 'import', '-from', 'import', '-strict'])
```
gadfort, line 14: you don't need `-from` here

tests/core/test_archive.py
```diff
@@ -69,25 +69,6 @@ def test_archive(chip):
...
-        contents = f.getnames()
-
-    for item in ('build/oh_parity/job0/oh_parity.pkg.json',
-                 'build/oh_parity/job0/import/0/reports',
-                 'build/oh_parity/job0/import/0/outputs',
-                 'build/oh_parity/job0/import/0/import.log'):
-        assert item in contents
-
-    for item in contents:
-        assert not item.startswith('build/oh_parity/job0/syn')
```
gadfort, line 88: why is this removed? this doesn't change

tests/core/test_check_flowgraph.py
```diff
@@ -9,19 +9,19 @@ def test_check_flowgraph_steplist_not_set():
...
 
-    chip.set('option', 'steplist', ['import', 'syn'])
+    chip.set('option', 'from', ['syn2'])
 
-    assert chip._check_flowgraph()
+    assert not chip._check_flowgraph()
 
 
-def test_check_flowgraph_extra_steps():
+def test_check_flowgraph_extra_to_steps():
```
gadfort, line 21: test for then the from and to are not connected is also needed.

tests/core/test_summary.py
```diff
@@ -23,10 +23,10 @@ def gcd_with_metrics(gcd_chip):
             for metric in gcd_chip.getkeys('flowgraph', flow, step, index, 'weight'):
                 gcd_chip.set('flowgraph', flow, step, index, 'status', NodeStatus.SUCCESS)
                 gcd_chip.set('metric', metric, str(dummy_data), step=step, index=index)
-                prev_step = steps.index(step) - 1
+                prev_step = steps.index((step, index)) - 1
```
gadfort, line 26: this test should be corrected (it looks to have been wrong before too), just picking the previous does not guarantee its the previous node.

tests/flows/test_flowstatus.py
```diff
@@ -13,12 +14,12 @@
 
 from siliconcompiler.tools.builtin import nop
 from siliconcompiler.tools.builtin import minimum
+from siliconcompiler import SiliconCompilerError
 
 
 @pytest.mark.eda
 @pytest.mark.quick
 @pytest.mark.parametrize('steplist', [
-    ['import', 'place'],
```
gadfort, line 21: why is this one removed?

tests/flows/test_flowstatus.py
```diff
@@ -63,7 +64,8 @@ def test_flowstatus(scroot, steplist):
     chip.node(flow, 'cts', cts)
     chip.edge(flow, 'placemin', 'cts')
 
-    chip.set('option', 'steplist', steplist)
+    chip.set('option', 'from', [steplist[0]])
+    chip.set('option', 'to', [steplist[-1]])
```
gadfort, line 68: you could just make the list above to and from so it's clearer

siliconcompiler/core.py
```diff
@@ -2525,13 +2524,12 @@ def __archive_job(self, tar, job, steplist, index=None, include=None):
...
 
-        indexlist = [index] if index else None
-        for (step, idx) in self._get_flowgraph_nodes(flow, steplist=steplist, indexlist=indexlist):
+        for (step, idx) in flowgraph_nodes:
             self.logger.info(f'Archiving {step}{idx}...')
             self._archive_node(tar, step, idx, include=include)
 
     ###########################################################################
-    def archive(self, jobs=None, step=None, index=None, include=None, archive_name=None):
+    def archive(self, jobs=None, include=None, archive_name=None):
```
gadfort, line 2532: this should not change, if step and or index is provided, only those should be archived.

siliconcompiler/core.py
```diff
@@ -3082,37 +3064,28 @@ def summary(self, steplist=None, show_all_indices=False,
...
-        All step keys from the flowgraph dictionary are collected and the
-        distance from the root node (ie. without any inputs defined) is
-        measured for each step. The step list is then sorted based on
-        the distance from root and returned.
+        Returns an ordered list of flowgraph nodes which will be executed.
+        This takes the from/to options into account.
 
         Returns:
-            A list of steps sorted by distance from the root node.
+            A list of steps serialized in a legal execution order.
```
gadfort, line 3073: is sorting needed?

siliconcompiler/core.py
```diff
@@ -4036,12 +3998,12 @@ def _get_indexlist(self):
...
         return None
 
-    def _reset_flow_nodes(self, flow, steplist, indexlist):
+    def _reset_flow_nodes(self, flow, nodes):
         # Reset flowgraph/records/metrics by probing build directory. We need
         # to set values to None for steps we may re-run so that merging
         # manifests from _runtask() actually updates values.
         should_resume = self.get("option", 'resume')
-        node_list = self._get_flowgraph_nodes(flow, steplist=steplist, indexlist=indexlist)
+        node_list = nodes
```
gadfort, line 4006: remove node_list use if it's not needed

siliconcompiler/core.py
```diff
@@ -4169,30 +4144,26 @@ def _launch_nodes(self, nodes_to_run, processes, status):
...
-
-            if not index_succeeded:
-                self.error('Run() failed, see previous errors.', fatal=True)
+    def _check_nodes_status(self, flow, status, from_nodes, to_nodes):
+        def success(node):
+            return status[node] == NodeStatus.SUCCESS
+        reached_nodes = set(self._reachable_flowgraph_nodes(flow, from_nodes, cond=success))
+        unreached_nodes = to_nodes.difference(reached_nodes)
+        if unreached_nodes:
+            self.error(f'Run() failed - These final nodes could not be reached: {unreached_nodes}',
```
gadfort, line 4153: should be formatted as `cts0, cts1, and cts2`

gadfort (changes requested):

tests/flows/test_flowstatus.py
gadfort: okay, can you add a unit test to ensure a forked run will be marked a failure?

siliconcompiler/remote/client.py
troibe: Honestly I don't know if this actually needed.
There was some logic like that too in siliconcompiler itself for the step/indexlist...

siliconcompiler/core.py
troibe: Removed it from the description

tests/flows/test_flowstatus.py
troibe: `test_long_branch` tests the same disjoint branches with failures and I changed the behavior there.
I think this would be duplicate.

siliconcompiler/remote/client.py
troibe: Changed it so it starts off at any successor of an entry node.

siliconcompiler/remote/client.py
troibe: I'll try if I can get it to work without.

siliconcompiler/scheduler.py
troibe: Sure I can just make it such that `arg_step` overwrites `from` and `to` it's value.

troibe:
So I looked into it but at least if we allow multiple disjoint flowgraphs within the same flowgraph,
then this will only catch it if `from`/`to` are explicitely defined.
By default the two single nodes `import` and `syn` without a connection are both `entry` and `exit` nodes.
Therefore the default implied `from=['import', 'syn']` and `to=['import', 'syn']`.
I don't think this is probably not what a novice users intends so we should really reconsider whether we want to allow multiple disjoint flowgraphs within the same flowgraph,

If we explicitely set `from=['import']` and `to=['syn']`. It will catch as expected.

siliconcompiler/core.py
```diff
@@ -3852,10 +3854,9 @@ def _setup_node(self, step, index):
         else:
             self.error(f'setup() not found for tool {tool}, task {task}', fatal=True)
 
-        # Need to clear index, otherwise we will skip setting up other indices.
-        # Clear step for good measure.
-        self.set('arg', 'step', None)
-        self.set('arg', 'index', None)
+        # Need to restore step/index, otherwise we will skip setting up other indices.
```
troibe, line 3857: This is the preserving of `step/index` args that I was just talking about in our standup.

siliconcompiler/remote/client.py
```diff
@@ -166,10 +167,9 @@ def _remote_preprocess(chip, remote_nodelist):
     # we need to send inputs up to the server.
     chip._collect()
 
-    # Set 'from/to' to only the remote steps, for the future server-side run.
-    chip.unset('arg', 'step')
-    chip.unset('arg', 'index')
-    chip.set('option', 'to', to_nodes)
+    # Recover step/index
+    chip.set('arg', 'step', preset_step)
```
troibe, line 171: Same thing here.

siliconcompiler/core.py
```diff
@@ -1852,6 +1852,16 @@ def _check_flowgraph(self, flow=None):
                 self.logger.error(f'{step} is not defined in the {flow} flowgraph')
                 error = True
 
+        from_nodes = self._get_execution_entry_nodes(flow)
+        to_nodes = self._get_execution_exit_nodes(flow)
+        reachable_nodes = set(self._reachable_flowgraph_nodes(flow, set(from_nodes)))
+        unreachable_nodes = set(to_nodes).difference(reachable_nodes)
+        if unreachable_nodes:
+            unreachable_nodes_formated = list(
+                map(lambda node: f'{node[0] + node[1]}', unreachable_nodes))
```
gadfort, line 1861: can this be `f'{node[0]}{node[1]}'`

siliconcompiler/core.py
```diff
@@ -1852,6 +1852,16 @@ def _check_flowgraph(self, flow=None):
...
+            self.logger.error(f'These final nodes can not be reached: {unreachable_nodes_formated}')
```
gadfort, line 1862: Maybe update to say: `These final nodes in {flow} can not be reached`? I'll leave that up to you

siliconcompiler/core.py
```diff
@@ -2556,9 +2566,16 @@ def archive(self, jobs=None, include=None, archive_name=None):
...
-
         if not archive_name:
-            archive_name = f"{design}_{jobname}.tgz"
+            if step and index:
+                archive_name = f"{design}_{jobname}_{step}{index}.tgz"
+                flowgraph_nodes = [(step, index)]
+            elif step:
+                archive_name = f"{design}_{jobname}_{step}.tgz"
+                flow = self.get('option', 'flow')
+                flowgraph_nodes = self._get_flowgraph_nodes(flow=flow, steplist=[step])
```
gadfort, line 2576: steplist? maybe update the function signature to be "steps="?

gadfort (commented):
I think this looks good. Please check the documentation to see if we discuss steplist or indexlist there?

troibe:
@gadfort The last commit removes the remaining mentions of steplist and makes it more clear if all `flowgraph_nodes` are used or just the `nodes_to_execute`.
Please proofread the docstring for `run()`.
I think we should be pretty much ready now.

siliconcompiler/core.py
```diff
@@ -4070,25 +3983,33 @@ def _reset_flow_nodes(self, flow, steplist, indexlist):
...
                         self.set('flowgraph', flow, step, index, 'status', None)
                         for metric in self.getkeys('metric'):
                             self._clear_metric(step, index, metric)
                         for record in self.getkeys('record'):
                             self._clear_record(step, index, record)
 
-    def _prepare_nodes(self, nodes_to_run, processes, flow, status, steplist, indexlist):
+    def clean_build_dir(self):
+        if not self.get('option', 'resume') and not self.get('arg', 'step') \
+                and not self.get('option', 'from') and not self.get('record', 'remoteid'):
```
gadfort, line 3998: can you make a ticket to track the remoteid thing? I'm not 100% sure this is correct, but I can't think of how it is wrong.

siliconcompiler/core.py
```diff
@@ -4222,20 +4145,20 @@ def _local_process(self, flow, status, steplist, indexlist):
...
         '''
         Executes tasks in a flowgraph.
 
-        The run function sets up tools and launches runs for every index
-        in a step defined by a steplist. The steplist is taken from the schema
-        steplist parameter if defined, otherwise the steplist is defined
-        as the list of steps within the schema flowgraph dictionary. Before
-        starting  the process, tool modules are loaded and setup up for each
+        The run function sets up tools and launches runs for every node
+        in the flowgraph starting with 'from' steps and ending at 'to' steps.
```
gadfort, line 4158: steps -> nodes

gadfort (approved):
LGTM, just correct the docstring for `run()`

gadfort:
@troibe please do not merge this as this needs to go along with the #1896

siliconcompiler/core.py
```diff
@@ -4289,12 +4211,84 @@ def run(self):
...
+        return nodes_sorted
+
+    ###########################################################################
+    def nodes_to_execute(self, flow=None):
+        '''
+        Returns an ordered list of flowgraph nodes which will be executed.
+        This takes the from/to options into account.
+
+        Returns:
+            A list of steps that will get executed during run().
```
nmoroze, line 4249: steps -> nodes?

siliconcompiler/report/summary_table.py
```diff
@@ -69,7 +69,7 @@ def _show_summary_table(chip, flow, steplist, show_all_indices):
 
         libraries = set()
         for val, step, _ in chip.schema._getvals('asic', 'logiclib'):
-            if not step or step in steplist:
+            if not step or step in flowgraph_nodes:
```
nmoroze, line 72: I think you need to incorporate index here, e.g.

```python
for val, step, index in chip.schema._getvals(...):
     if not step or (step, index) in flowgraph_nodes:
```

siliconcompiler/scheduler.py
```diff
@@ -112,11 +112,6 @@ def _defernode(chip, step, index):
             sf.write('#!/bin/bash\n')
             sf.write(f'sc -cfg {shlex.quote(cfg_file)} -builddir {shlex.quote(buildir)} '
                      f'-arg_step {shlex.quote(step)} -arg_index {shlex.quote(index)}\n')
-            # In case of error(s) which prevents the SC build script from completing, ensure the
-            # file mutex for job completion is set in shared storage. This lockfile signals the
-            # server to mark the job done, without putting load on the cluster reporting/accounting
-            # system.
-            sf.write(f'touch {os.path.dirname(output_file)}/done')
```
nmoroze, line 119: Seems like unrelated change. Why is this being removed?

tests/apps/test_sc.py
```diff
@@ -11,7 +11,7 @@ def test_design_inference(scroot, monkeypatch):
     source = os.path.join(scroot, 'tests', 'data', 'heartbeat.v')
     # only run import, makes this quicker
 
-    monkeypatch.setattr('sys.argv', ['sc', source, '-steplist', 'import', '-strict'])
+    monkeypatch.setattr('sys.argv', ['sc', source, '-from', 'import', '-strict'])
```
nmoroze, line 14: Should this be `-to import` instead? (won't break test either way, but should make it faster)

tests/flows/test_from_to.py
```diff
@@ -22,33 +22,36 @@ def test_steplist(gcd_chip):
...
     gcd_chip.run()
     assert gcd_chip.find_result('def', step='floorplan')
 
 
 @pytest.mark.eda
-def test_steplist_keep_reports(gcd_chip):
+def test_from_to_keep_reports(gcd_chip):
     '''Regression test for making sure that reports from previous steps are
-    still mapped when a script is re-run with a steplist.'''
+    still mapped when a script is re-run with afrom/to.'''
```
nmoroze, line 40: typo: afrom->from

nmoroze (changes requested):
Found a couple subtle things that I think be fixed, but overall this looks great, thanks!

Another high-level note is I noticed a lot of casting to `set()` where I think it may not be needed. I think it would be cleaner to remove, but it does seem safer and not dangerous from a functionality perspective to keep those... maybe someday in the future if we add type annotations it will be more comfortable to get rid of those :)

siliconcompiler/scheduler.py
troibe: That was to make debugging easier and is now merged: https://github.com/siliconcompiler/siliconcompiler/pull/1950

siliconcompiler/core.py
```diff
@@ -3022,12 +3026,12 @@ def _dashboard(self, wait=True, port=None, graph_chips=None):
         return dash
 
     ###########################################################################
-    def summary(self, steplist=None, show_all_indices=False,
+    def summary(self, from_nodes=None, to_nodes=None, show_all_indices=False,
```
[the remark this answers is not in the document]
troibe: Sure I can do that. Maybe I should that logic back in if `summary()` is considered to be part of the python API?
Currently that can be done with `chip.set('option', 'from', ['step'])` so removing it also works fine.

siliconcompiler/core.py
```diff
@@ -4114,27 +4034,32 @@ def _check_node_dependencies(self, node, deps, status):
...
+        deps_was_successful = {}
         while len(nodes_to_run) > 0 or len(running_nodes) > 0:
             # Check for new nodes that can be launched.
             for node, deps in list(nodes_to_run.items()):
                 # TODO: breakpoint logic:
                 # if node is breakpoint, then don't launch while len(running_nodes) > 0
 
-                self._check_node_dependencies(node, deps, status)
+                deps_was_successful[node] = self._check_node_dependencies(node, deps, status,
+                                                                          deps_was_successful)
```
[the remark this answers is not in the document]
troibe: Thanks for pointing it out. The idea is to operate on the top level dictionary so I'll make it consistent with that

siliconcompiler/core.py
troibe: You mean 'from' nodes and 'to' nodes?
So far those options are step based so writing nodes here might be confusing to the user.

siliconcompiler/apps/sc_remote.py
```diff
@@ -129,13 +129,14 @@ def main():
...
         environment = copy.deepcopy(os.environ)
         entry_nodes = chip._get_flowgraph_entry_nodes(chip.get('option', 'flow'))
-        for node in entry_nodes:
-            remote_steps.remove(node[0])
-        chip.set('option', 'steplist', remote_steps)
+        flow = chip.get('option', 'flow')
+        entry_nodes = chip._get_flowgraph_entry_nodes(flow)
+        for entry_node in entry_nodes:
+            outputs = chip._get_flowgraph_node_outputs(flow, entry_node)
+            chip.set('option', 'from', list(map(lambda node: node[0], outputs)))
```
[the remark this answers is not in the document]
troibe: Yes that should be handled correctly. Internally [when checking the flowgraph entry nodes](https://github.com/siliconcompiler/siliconcompiler/blob/from-to-rebase/siliconcompiler/core.py/#L4729-L4730) we just get the nodes that are [in these from steps](https://github.com/siliconcompiler/siliconcompiler/blob/from-to-rebase/siliconcompiler/core.py/#L4714-L4716). So if there are duplicates in the from steps it doesn't make a difference. (Maybe it would be a tiny bit slower if we had huge graphs)

siliconcompiler/core.py
gadfort: good point

siliconcompiler/core.py
troibe: SC-636

siliconcompiler/core.py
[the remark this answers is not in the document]
nmoroze: `summary()` is considered part of the API, but I'm overall happy with simplifying the API if we don't see a good use case for this (easier to add back in later than to delete). 

The one use-case I recall of the `steplist` arg was in `sc.py`, but I can't remember why we used it there when I'm pretty sure just reading from the schema is effectively the same functionality wise.

siliconcompiler/report/summary_table.py
```diff
@@ -69,7 +69,7 @@ def _show_summary_table(chip, flow, flowgraph_nodes, show_all_indices):
 
         libraries = set()
         for val, step, _ in chip.schema._getvals('asic', 'logiclib'):
-            if not step or step in flowgraph_nodes:
+            if not step or (step, index) in flowgraph_nodes:
```
nmoroze, line 72: Watch out, also need to add in `index` to the for loop (otherwise this is using a stale value of index I believe)

nmoroze (changes requested):
Thanks for addressing the feedback! Just one more thing.

siliconcompiler/report/summary_table.py
troibe: Thanks for catching that one.

nmoroze (approved):

PR #1981, opened 2023-09-21
Add [option, softlib]

nmoroze:
Adds support for soft libraries instead of using macrolib/logiclib for pulling in Verilog sources, ydirs, etc. 

@gadfort do you have a strong opinion about adding support to all frontends before merge?


aolofsson:
This might need some discussion. Is softlib meant to replace logiclib and macrolib completely?

nmoroze:
Nope, would be along with them. Originally opened this to solve an immediate problem, but closing since I think I have another solution in the meantime. 

I do think we'll need something like this at some point, but we can save it for a larger discussion.

PR #1987, opened 2023-09-22
Use flowgraph to see if verilator is entry node

troibe:
In case this is not an entry steps and the file is not there then we would error out now instead of searching for sources now.
@nmoroze Was this the kind of behaviour you were thinking of?


nmoroze (approved):
Yup, this is what I was thinking in that ticket. Adding @gadfort to make sure he's onboard (sorry if this was already discussed, but I think this ticket has been around for a while and can't remember if we got full buy-in).

troibe:
Closing this as @gadfort suggested it would be better to have a more holistic approach to integrating input and output files as part of the flow

PR #1988, opened 2023-09-22
fix daily CI to use correct ref

gadfort:


troibe (approved):
Sure we can write it explicitly.
The default ref before was `main`.
Shouldn't that result in the same commit id?

gadfort:
@troibe not when this is triggered by a tool builds, then the ref is their branches and it ends up unable to test what we are looking for.

troibe:
Oh so you specfically don't want it to be main in those cases.
Makes sense then.

PR #1989, opened 2023-09-22
update openroad and allow for better control of heatmaps

gadfort:
Adds controls to heatmaps to set number of bins
Only emit heatmaps with data in them.
Always show clock tree.

### DIFF
--- a/siliconcompiler/tools/openroad/scripts/sc_procs.tcl
+++ b/siliconcompiler/tools/openroad/scripts/sc_procs.tcl
@@ -116,10 +116,7 @@ proc sc_pin_placement { args } {
 ###########################
 
 proc sc_has_gui {} {
-  if {[expr [llength [info procs save_image]] > 0]} {
-      return true
-  }
-  return false
+  return [gui::supported]
 }
 
 ###########################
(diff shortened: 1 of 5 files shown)


siliconcompiler/tools/openroad/scripts/sc_procs.tcl
```diff
@@ -116,10 +116,7 @@ proc sc_pin_placement { args } {
 ###########################
 
 proc sc_has_gui {} {
-  if {[expr [llength [info procs save_image]] > 0]} {
-      return true
-  }
-  return false
+  return [gui::supported]
```
troibe, line 119: Is this a new attribute in openroad or comming from our siliconcompiler tcl?

siliconcompiler/tools/openroad/scripts/sc_write_images.tcl
```diff
@@ -124,6 +166,7 @@ proc sc_image_routing_congestion {} {
     "Routing" \
     "routing_congestion.png" \
     "routing congestion"
+    0
```
troibe, line 169: So we are not using your new logic from the `sc_image_heatmap()` proc here?

troibe (approved):
Looks reasonable to me

siliconcompiler/tools/openroad/scripts/sc_write_images.tcl
gadfort: the routing congestion is a little more limited so we cannot use that logic yet

PR #1992, opened 2023-09-23
ensure CLI apps use similar switches for show and dashboard to find manifests

### LINKED ISSUE #1990: Post flow consistency
It would nice for the post run tools to have consistent command line switches and behavior. 

Example -design exists on sc-show but not in sc-dashboard.
Example, sc-dashboard gives help with no arg, sc-show runs through error.

gadfort:
Closes #1990

- Allows both `sc-show` and `sc-dashboard` to use the same switches to load the initial manifest
- Wrap all CLI `create_cmdline` with try/except to provide error message to users without backtrace.

### DIFF
--- a/siliconcompiler/apps/_common.py
+++ b/siliconcompiler/apps/_common.py
@@ -1,0 +1,70 @@
+import glob
+import os
+
+
+def manifest_find_switches():
+    return ['-design',
+            '-cfg',
+            '-arg_step',
+            '-arg_index',
+            '-jobname']
+
+
+def load_manifest(chip, src_file):
+    manifest = None
+    if (src_file is not None) and (not chip.get('option', 'cfg')):
+        # only autoload manifest if user doesn't supply manually
+        manifest = _get_manifest(os.path.dirname(src_file))
+        if not manifest:
+            design = os.path.splitext(os.path.basename(src_file))[0]
+            chip.logger.error(f'Unable to automatically find manifest for design {design}. '
+                              'Please provide a manifest explicitly using -cfg.')
+            return False
+    elif not chip.get('option', 'cfg'):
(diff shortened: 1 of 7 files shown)


siliconcompiler/apps/_common.py
```diff
@@ -0,0 +1,69 @@
+import glob
+import os
+
+
+def manifest_find_switches():
```
troibe, line 5: Nit: maybe just `manifest_switches()` would be more accurate as their is no finding going on.

siliconcompiler/apps/sc_dashboard.py
```diff
@@ -40,19 +41,26 @@ def main():
...
+        chip.logger.error(e)
+        return 1
 
     # Error checking
     design = chip.get('design')
     if design == UNSET_DESIGN:
         chip.logger.error('Design not loaded')
         return 1
 
+    if not load_manifest(chip, None):
```
troibe, line 61: Maybe I'm missing something but weren't we loading the manifest before to show the information?
Just wondering because we might be loading the manifest twice now...

troibe (approved):
Looks good

siliconcompiler/apps/sc_dashboard.py
gadfort: I'm not sure I follow

gadfort: here we will only load the manifest if it was not provided via the -cfg flag

troibe: What I meant is: Doesn't `chip.create_cmdline()` already load the manifest from the `cfg`?

troibe: Sorry misread your answer

PR #1993, opened 2023-09-23
provide siliconcompiler cli app as alias for sc

gadfort:
because `sc` is a system alias in Windows powershell, this provides a full name `siliconcompiler` cli app which is an alias to `sc` to allow for easier use on Windows.


aolofsson (approved):
Nice solution!

Issue #1986, opened 2023-09-22
Could not find remote server configuration: defaulting to https://server.siliconcompiler.com


yyh-sjtu:
<img width="772" alt="image" src="https://github.com/siliconcompiler/siliconcompiler/assets/131600287/9ce5cb8b-7ebd-4980-a9e1-140ade91e08a">
After I executed the command "sc -remote -target "asic_demo"", the shell told me that "Could not find remote server configuration" and the "https://server.siliconcompiler.com" cannot be visited either. 
<img width="760" alt="image" src="https://github.com/siliconcompiler/siliconcompiler/assets/131600287/90790a82-770b-4e23-a838-19b032a18f20">

gadfort:
That is just a notice that your job will be submitted to the default server (it does not have a public facing website, SC will upload the job to it and it will run it and return the results).
The failure you are encountering seems to relate to an issue with Surelog. What OS and python versions are you using? and did you install Surelog separately?

yyh-sjtu:
Thank you very much for reply. The OS is Windows, and the version of python is 3.10.0. I haven't install Surelog, because I don't know much about the siliconcompiler. I think it can run without any other tools, if I choose the remote option.

yyh-sjtu:
> That is just a notice that your job will be submitted to the default server (it does not have a public facing website, SC will upload the job to it and it will run it and return the results). The failure you are encountering seems to relate to an issue with Surelog. What OS and python versions are you using? and did you install Surelog separately?

Thank you very much for reply. The OS is Windows, and the version of python is 3.10.0. I haven't install Surelog, because I don't know much about the siliconcompiler. I think it can run without any other tools, if I choose the remote option.

gadfort:
It looks like the surelog that we bundle with Windows did not compile properly. We'll look into that.

yyh-sjtu:
> It looks like the surelog that we bundle with Windows did not compile properly. We'll look into that.
OK. Thank you very much. You reply helps me a lot. I think I can choose linux OS to use siliconcompiler.

yyh-sjtu:
> It looks like the surelog that we bundle with Windows did not compile properly. We'll look into that.
OK. Thank you very much. Your reply helps me a lot. I think I can choose linux OS to use siliconcompiler.

PR #1994, opened 2023-09-25
build surelog with static zlib

gadfort:
Build surelog with static linking for zlib dependency: https://github.com/siliconcompiler/siliconcompiler/suites/16508799539/artifacts/943209333

Closes: #1986


nmoroze (approved):
Works for me! Thanks for fixing this.

PR #1996, opened 2023-09-25
build two artifacts for wheels build to keep siliconcompiler artifact small

gadfort:
Wheels: https://github.com/siliconcompiler/siliconcompiler/actions/runs/6303016707


nmoroze (approved):
Looks good, one clarification is that the dependencies artifact also includes SC wheels themselves, so you can update the instructions to say you only need to click on "dependencies". Might also be worth changing the name of the artifact to something like "sc_plus_dependencies" to make that clear.

gadfort:
https://github.com/siliconcompiler/siliconcompiler/actions/runs/6304784562

PR #1998, opened 2023-09-25
Update yosys to make use of new -dont_use flag to ABC

gadfort:
This removes the need to modify the libraries for ABC to mark cells as dont_use, but instead use the new -dont_use.

### DIFF
--- a/siliconcompiler/tools/yosys/syn_asic.tcl
+++ b/siliconcompiler/tools/yosys/syn_asic.tcl
@@ -72,13 +72,16 @@ proc determine_keep_hierarchy { iter cell_limit } {
 # DESIGNER's CHOICE
 ####################
 
+set sc_logiclibs        [dict get $sc_cfg asic logiclib]
+set sc_macrolibs        [dict get $sc_cfg asic macrolib]
+
 set sc_libraries        [dict get $sc_cfg tool $sc_tool task $sc_task {file} synthesis_libraries]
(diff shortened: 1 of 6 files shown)


siliconcompiler/tools/yosys/syn_asic.tcl
```diff
@@ -265,8 +268,14 @@ if {$script != ""} {
...
 }
+set abc_dont_use []
+foreach lib "$sc_logiclibs $sc_macrolibs" {
+    foreach cell [dict get $sc_cfg library $lib asic cells dontuse] {
+        lappend abc_dont_use -dont_use $cell
+    }
+}
 
-yosys abc -liberty $sc_dff_library {*}$abc_args
+yosys abc {*}$abc_args {*}$abc_dont_use
```
troibe, line 278: Not sure if removing the `-liberty` flag was intended.
In case you want to add it back in:
```suggestion
yosys abc -liberty $sc_dff_library {*}$abc_args {*}$abc_dont_use
```

troibe (approved):
Looks good. I'm not sure if you accidentally removed that `-liberty` flag.
Or if it is intentional since we do the processing now in siliconcompiler.

siliconcompiler/tools/yosys/syn_asic.tcl
gadfort: this is intentional, it's just passing in a library ABC doesnt actually need.

PR #2000, opened 2023-09-27
Check tool versions at beginning of run

troibe:
**Why?**
Currently even on a local run we only check if the tool version is correct and the tool exists when executing a node.
This leads to failures in the middle of the run and is annoying to the user.

**What?**
If the nodes are executed locally then we check at the beginning of the run if all tools are present and have right version.

**How?**
We use the same tool version check that we use when executing a node but run it also at the start of the run.

**Issues:**
Some tools like `iverilog` produce their binary while executing the flow.
These binaries can't be version checked in the beginning of the run.

**Before:**
```-------------------------------------------------------------------------------- Captured stdout call --------------------------------------------------------------------------------
| INFO    | /home/martin-zeroasic/coding/siliconcompiler/examples/gcd/gcd.v inferred as rtl/verilog
| INFO    | /home/martin-zeroasic/coding/siliconcompiler/examples/gcd/gcd.sdc inferred as constraint/sdc
| INFO    | job0 | --        | - | Setting up node import0 with surelog/parse
| INFO    | job0 | --        | - | Setting up node syn0 with yosys/syn_asic
| INFO    | job0 | --        | - | Setting up node floorplan0 with openroad/floorplan
| INFO    | job0 | --        | - | Setting up node place0 with openroad/place
| INFO    | job0 | --        | - | Setting up node cts0 with openroad/cts
| INFO    | job0 | --        | - | Setting up node route0 with openroad/route
| INFO    | job0 | --        | - | Setting up node dfm0 with openroad/dfm
| INFO    | job0 | --        | - | Setting up node export0 with klayout/export
| INFO    | job0 | --        | - | Setting up node export1 with openroad/export
| INFO    | job0 | --        | - | Checking manifest before running.
| INFO    | job0 | import    | 0 | Tool 'surelog' found with version '1.71' in directory '/usr/local/bin'
| INFO    | job0 | import    | 0 | Running in /tmp/pytest-of-martin-zeroasic/pytest-24/test_py_read_manifest0/build/gcd/job0/import/0
| INFO    | job0 | import    | 0 | surelog -nocache +libext+.sv+.v -parse -nouhdm /home/martin-zeroasic/coding/siliconcompiler/examples/gcd/gcd.v -top gcd
| INFO    | job0 | import    | 0 | Computing hash value for [tool,surelog,task,parse,output]
| INFO    | job0 | import    | 0 | Computing hash value for [input,rtl,verilog]
| INFO    | job0 | import    | 0 | Finished task in 0.59s
| INFO    | job0 | syn       | 0 | Tool 'yosys' found with version '0.31+16' in directory '/usr/local/bin'
| INFO    | job0 | syn       | 0 | Running in /tmp/pytest-of-martin-zeroasic/pytest-24/test_py_read_manifest0/build/gcd/job0/syn/0
| INFO    | job0 | syn       | 0 | yosys -c /home/martin-zeroasic/coding/siliconcompiler/siliconcompiler/tools/yosys/sc_syn.tcl
| INFO    | job0 | syn       | 0 | Computing hash value for [tool,yosys,task,syn_asic,output]
| INFO    | job0 | syn       | 0 | Computing hash value for [library,nangate45,output,typical,nldm]
| INFO    | job0 | syn       | 0 | Computing hash value for [library,nangate45,option,file,yosys_addermap]
| INFO    | job0 | syn       | 0 | Finished task in 2.4s
| INFO    | job0 | floorplan | 0 | Tool 'openroad' found with version 'v2.0-9907' in directory '/usr/local/bin'
| ERROR   | job0 | floorplan | 0 | Version check failed for openroad. Check installation.
| ERROR   | job0 | floorplan | 0 | Found version v2.0-9907, did not satisfy any version specifier set >=v2.0-10370.
| ERROR   | job0 | floorplan | 0 | Halting step 'floorplan' index '0' due to errors.
```

**After:**
```

-------------------------------------------------------------------------------- Captured stdout call --------------------------------------------------------------------------------
| INFO    | gcd.v inferred as rtl/verilog
| INFO    | rtl2gds | --        | - | Setting up node import0 with surelog/parse
| INFO    | rtl2gds | --        | - | Tool 'surelog' found with version '1.71' in directory '/usr/local/bin'
| INFO    | rtl2gds | --        | - | Setting up node syn0 with yosys/syn_asic
| INFO    | rtl2gds | --        | - | Tool 'yosys' found with version '0.31+16' in directory '/usr/local/bin'
| INFO    | rtl2gds | --        | - | Setting up node floorplan0 with openroad/floorplan
| INFO    | rtl2gds | --        | - | Tool 'openroad' found with version 'v2.0-9907' in directory '/usr/local/bin'
| ERROR   | rtl2gds | --        | - | Version check failed for openroad. Check installation.
| ERROR   | rtl2gds | --        | - | Found version v2.0-9907, did not satisfy any version specifier set >=v2.0-10370.
| ERROR   | rtl2gds | --        | - | Halting step 'floorplan' index '0' due to errors.
```

**Testing:**
- `tests/core/test_error_manifest.py` and `tests/core/test_fail_early.py` needed to be added to the eda tests since we are now checking earlier if their executable is present
- `tests/core/test_version_early.py`

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -4211,6 +4211,18 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         for (step, index) in flowgraph_nodes:
             # Setting up tool is optional
             self._setup_node(step, index)
+            # Env vars are necessary as test_multiple_tools.py requires it for its version check
+            self._set_env_vars(step, index)
+            tool, task = self._get_tool_task(step, index, flow)
+            # Icarus compiles its executable during the run so we can't check its version
+            if (tool, task) != ('execute', 'exec_input'):
+                run_func = getattr(self._get_task_module(step, index, flow=flow), 'run', None)
+                try:
+                    self._check_tool_version(step, index, run_func)
+                # Convert sys.exit(1) from haltstep() in version check to SilicoCompilerError
+                except SystemExit:
+                    self.error('Pre-run version check failed. Please update your tools.',
+                               fatal=True)
 
         # Check validity of setup
         self.logger.info("Checking manifest before running.")
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -4196,6 +4196,17 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
             if status[node] != NodeStatus.PENDING:
                 self.set('flowgraph', flow, step, index, 'status', status[node])
 
+    def _pre_run_version_check(self, flow, flowgraph_nodes):
+        # Don't print information messages produced by _check_tool_version()
+        prev_log_level = self.logger.level
+        self.logger.setLevel(logging.WARNING)
+
+        for (step, index) in flowgraph_nodes:
+            run_func = getattr(self._get_task_module(step, index, flow=flow), 'run', None)
+            self._check_tool_version(step, index, run_func)
+
+        self.logger.setLevel(prev_log_level)
+
     def _local_process(self, flow, status, steplist, indexlist):
         # Populate status dict with any flowgraph status values that have already
         # been set.
@@ -4211,18 +4222,6 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         for (step, index) in flowgraph_nodes:
             # Setting up tool is optional
             self._setup_node(step, index)
-            # Env vars are necessary as test_multiple_tools.py requires it for its version check
-            self._set_env_vars(step, index)
-            tool, task = self._get_tool_task(step, index, flow)
-            # Icarus compiles its executable during the run so we can't check its version
-            if (tool, task) != ('execute', 'exec_input'):
-                run_func = getattr(self._get_task_module(step, index, flow=flow), 'run', None)
-                try:
-                    self._check_tool_version(step, index, run_func)
-                # Convert sys.exit(1) from haltstep() in version check to SilicoCompilerError
-                except SystemExit:
-                    self.error('Pre-run version check failed. Please update your tools.',
-                               fatal=True)
 
         # Check validity of setup
         self.logger.info("Checking manifest before running.")
@@ -4235,6 +4234,8 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
             self.error('Manifest check failed. See previous errors.', fatal=True)
         if self._error:
             self.error('Implementation errors encountered. See previous errors.', fatal=True)
+
+        self._pre_run_version_check(flow, flowgraph_nodes)
 
         nodes_to_run = {}
         processes = {}
--- a/siliconcompiler/tools/execute/execute.py
+++ b/siliconcompiler/tools/execute/execute.py
@@ -11,5 +11,6 @@ def setup(chip):
 
     tool, task = chip._get_tool_task(step, index)
 
-    chip.set('tool', tool, 'exe', ":exe:", clobber=False)
+    # Use dummy bash executable 'true' to pass pre-run version check
+    chip.set('tool', tool, 'exe', 'true', clobber=False)
     chip.set('tool', tool, 'task', task, 'option', [], step=step, index=index, clobber=False)
--- a/tests/core/test_error_manifest.py
+++ b/tests/core/test_error_manifest.py
@@ -1,9 +1,12 @@
 import siliconcompiler
 from siliconcompiler.tools.surelog import parse
-from siliconcompiler._common import SiliconCompilerError
 import os
 
+import pytest
 
+
+@pytest.mark.quick
+@pytest.mark.eda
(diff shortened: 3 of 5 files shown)


siliconcompiler/core.py
```diff
@@ -4211,6 +4211,18 @@ def _local_process(self, flow, status, steplist, indexlist):
         for (step, index) in flowgraph_nodes:
             # Setting up tool is optional
             self._setup_node(step, index)
+            # Env vars are necessary as test_multiple_tools.py requires it for its version check
+            self._set_env_vars(step, index)
```
gadfort, line 4215: if this is needed it should go into the `check_tool_version` function, don't forget to remove these variables after this is done

siliconcompiler/core.py
```diff
@@ -4211,6 +4211,18 @@ def _local_process(self, flow, status, steplist, indexlist):
...
+            tool, task = self._get_tool_task(step, index, flow)
+            # Icarus compiles its executable during the run so we can't check its version
+            if (tool, task) != ('execute', 'exec_input'):
```
gadfort, line 4218: I think `check_tool_version` should handle this by checking if there is a version to test

siliconcompiler/core.py
```diff
@@ -4211,6 +4211,18 @@ def _local_process(self, flow, status, steplist, indexlist):
...
+                run_func = getattr(self._get_task_module(step, index, flow=flow), 'run', None)
+                try:
+                    self._check_tool_version(step, index, run_func)
+                # Convert sys.exit(1) from haltstep() in version check to SilicoCompilerError
```
gadfort, line 4222: can this be changed to a `raise SiliconCompilerError` instead of `sys.exit(1)`?

siliconcompiler/core.py
```diff
@@ -4211,6 +4211,18 @@ def _local_process(self, flow, status, steplist, indexlist):
...
             # Setting up tool is optional
             self._setup_node(step, index)
+            # Env vars are necessary as test_multiple_tools.py requires it for its version check
+            self._set_env_vars(step, index)
+            tool, task = self._get_tool_task(step, index, flow)
+            # Icarus compiles its executable during the run so we can't check its version
+            if (tool, task) != ('execute', 'exec_input'):
+                run_func = getattr(self._get_task_module(step, index, flow=flow), 'run', None)
+                try:
+                    self._check_tool_version(step, index, run_func)
```
gadfort, line 4221: It might be nice to only get errors from this function when calling in this context? Otherwise it's a lot of printing.

Also this could be done in parallel to speed things up a bit

gadfort (changes requested):
I think this should be a separate function we can all after node setup and after manifest checking.

siliconcompiler/core.py
troibe: Currently it just fails and says the executable could not be found.
I can move this check into `check_tool_version` though.

siliconcompiler/core.py
troibe: Sure for now it is definitely printing the version checks here **and** when running the node so I could not log that here.
I don't think parallelizing here seems worth the additional complexity.

siliconcompiler/core.py
gadfort: I don't think this check is correct, we should find a better method then exclude a single node

siliconcompiler/core.py
gadfort: I think we should check this on some larger graphs on the VMs.

siliconcompiler/core.py
troibe: Just one test uses this to overwrite the version of a tool.
I think it's probably better to move this into the test instead of `check_tool_version` because the tool version shouldn't depend on the env vars in the normal case.

siliconcompiler/core.py
troibe: I'll look what verilator does differently in its tool setup since that should also produce an executable dynamically.

siliconcompiler/core.py
troibe: Sure I'll check if that works. That might be more elegant.

siliconcompiler/core.py
troibe: Fixed it in icarus by using the dummy executable `true` instead of a non existing executable.

siliconcompiler/core.py
troibe: Created a task for this: SC-638

siliconcompiler/core.py
troibe: Changed it back.
I think the reason we use `sys.exit(1)` is that this is not an error/exception in siliconcompiler so we should not show a stacktrace.

PR #2003, opened 2023-09-28
Return copy instead of reference for chip.get()

troibe:
**What?**
Always copy the list or other objects that would otherwise returned by reference from the schema.

**Why?**
Benazeer was most recently running into accidental edits of lists in the schema.
I have personally noticed this in the past.

**Issues:**
I'm not sure if this is "deep" enough.
Maybe this should even be on the `schema.get()` level.

**Testing:**
tests/core/test_get_copy.py

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -923,12 +923,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                     )
                     return None
 
-            # Prevent accidental modifications of the schema content by not passing a reference
-            result = self.schema.get(*keypath, field=field, job=job, step=step, index=index)
-            try:
-                return result.copy()
-            except AttributeError:
-                return result
+            return self.schema.get(*keypath, field=field, job=job, step=step, index=index)
         except (ValueError, TypeError) as e:
             self.error(str(e))
             return None
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -923,7 +923,12 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                     )
                     return None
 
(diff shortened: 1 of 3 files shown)


siliconcompiler/core.py
```diff
@@ -923,7 +923,12 @@ def get(self, *keypath, field='value', job=None, step=None, index=None):
                     )
                     return None
 
-            return self.schema.get(*keypath, field=field, job=job, step=step, index=index)
+            # Prevent accidental modifications of the schema content by not passing a reference
+            result = self.schema.get(*keypath, field=field, job=job, step=step, index=index)
+            try:
+                return result.copy()
+            except AttributeError:
+                return result
```
gadfort, line 931: the schema object should make the copy, not the chip object

gadfort (changes requested):

siliconcompiler/schema/schema_obj.py
```diff
@@ -134,13 +134,21 @@ def _read_manifest(filepath):
...
         Returns a schema parameter field.
 
         See :meth:`~siliconcompiler.core.Chip.get` for detailed documentation.
         """
+        # Prevent accidental modifications of the schema content by not passing a reference
+        result = self.__get(*keypath, field=field, job=job, step=step, index=index)
+        try:
+            return result.copy()
+        except AttributeError:
+            return result
```
gadfort, line 148: can we use:
```
import copy
...
return copy.copy(result)
```

instead? seems cleaner than try/except

troibe: Sure that works too.

troibe: Thanks for the suggestion I was looking for something like that.
Done now

gadfort (approved):

PR #2005, opened 2023-09-28
Windows pathlib fix

WRansohoff:
This change updates the Schema's normalization logic to always store paths in a POSIX format.

Windows machines will still be able to find files using the Schema's POSIX paths, even with absolute paths, thanks to the `pathlib.Path` parsing in `core.py`. I added [a daily CI test](https://github.com/siliconcompiler/siliconcompiler/actions/runs/6343604134/job/17231856210) to verify that.

While this logic won't change the path's meaning, it will strip leading `./` characters for relative paths; I had to adjust a couple of CI tests to account for that, but they still use the `==` operator on `Path` objects.

### DIFF
--- a/.github/workflows/daily_ci.yml
+++ b/.github/workflows/daily_ci.yml
@@ -118,3 +118,22 @@ jobs:
           cd $GITHUB_WORKSPACE/zerosoc
           ./build.py --core-only
           ./build.py --top-only
+
+  windows_unit_tests:
+    name: Windows Unit Tests
+    runs-on: windows-latest
+    steps:
+      - name: Checkout SiliconCompiler
+        uses: actions/checkout@v3
+
+      - name: Install Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
+          architecture: x64
+
+      - name: Install SC Without Surelog
+        run: READTHEDOCS=True pip3 install -r requirements.txt .
+
+      - name: Run Windows Unit Tests
+        run: pytest tests/schema/test_windows_paths.py
--- a/siliconcompiler/schema/schema_obj.py
+++ b/siliconcompiler/schema/schema_obj.py
@@ -570,7 +570,7 @@ class Schema:
 
         if sc_type in ('file', 'dir'):
             if isinstance(value, (str, pathlib.Path)):
-                return str(value)
+                return pathlib.Path(value).as_posix()
             else:
                 raise TypeError(error_msg)
 
--- a/tests/core/test_create_cmdline.py
+++ b/tests/core/test_create_cmdline.py
@@ -1,4 +1,5 @@
 import re
+import pathlib
 import pytest
 
 import siliconcompiler
@@ -272,7 +273,15 @@ def test_cli_examples(monkeypatch):
 
     for kp, step, index, val in expected_data:
         print("Check", kp, c.schema.get(*kp, step=step, index=index), val)
-        assert c.schema.get(*kp, step=step, index=index) == val
+        stype = c.schema.get(*kp, field='type')
+        if stype in ('file', 'dir'):
+            assert pathlib.Path(c.schema.get(*kp, step=step, index=index)) == pathlib.Path(val)
+        elif stype in ('[file]', '[dir]'):
+            actual_paths = c.schema.get(*kp, step=step, index=index)
+            for i in range(len(val)):
+                assert pathlib.Path(actual_paths[i]) == pathlib.Path(val[i])
+        else:
+            assert c.schema.get(*kp, step=step, index=index) == val
 
 
 def test_invalid_switch():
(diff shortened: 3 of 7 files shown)


siliconcompiler/schema/schema_obj.py
```diff
@@ -570,7 +570,7 @@ def _normalize_value(value, sc_type, error_msg, allowed_values):
 
         if sc_type in ('file', 'dir'):
             if isinstance(value, (str, pathlib.Path)):
-                return str(value)
+                return pathlib.Path(value).as_posix()
```
gadfort, line 573: we should still store it as a string (otherwise way too many things might break)

tests/core/test_create_cmdline.py
```diff
@@ -272,7 +273,15 @@ def _mock_read_manifest(chip, manifest, clobber=False, clear=False):
...
-        assert c.schema.get(*kp, step=step, index=index) == val
+        stype = c.schema.get(*kp, field='type')
+        if stype in ('file', 'dir'):
+            assert pathlib.Path(c.schema.get(*kp, step=step, index=index)) == pathlib.Path(val)
+        elif stype in ('[file]', '[dir]'):
+            actual_paths = c.schema.get(*kp, step=step, index=index)
+            for i in range(len(val)):
+                assert pathlib.Path(actual_paths[i]) == pathlib.Path(val[i])
+        else:
+            assert c.schema.get(*kp, step=step, index=index) == val
```
gadfort, line 284: this should not need to change (that would imply that this change is not backwards compatible)

tests/core/test_setget.py
```diff
@@ -85,7 +86,16 @@ def test_setget():
             # arbitrary step/index to avoid error
             step, index = 'syn', '0'
         result = chip.get(*keypath, step=step, index=index)
-        assert result == value, f'Expected value {value} from keypath {keypath}. Got {result}.'
+        stype = chip.get(*keypath, field='type')
```
gadfort, line 89: same

tests/schema/test_windows_paths.py
```diff
@@ -0,0 +1,33 @@
+import os
+import pathlib
+
+from siliconcompiler import Chip
+
+
+def test_windows_path_compat():
```
gadfort, line 7: spell out compatibility
also this wasn't where the failure occurred, we should have a test ensure that is also working

.github/workflows/daily_ci.yml
```diff
@@ -118,3 +118,22 @@ jobs:
...
+    runs-on: windows-latest
+    steps:
+      - name: Checkout SiliconCompiler
+        uses: actions/checkout@v3
+
+      - name: Install Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
+          architecture: x64
```
gadfort, line 133: is this needed?

.github/workflows/daily_ci.yml
```diff
@@ -118,3 +118,22 @@ jobs:
...
+
+      - name: Install SC Without Surelog
+        run: READTHEDOCS=True pip3 install -r requirements.txt .
```
gadfort, line 136: you could grab the cached version of surelog.
remove the READTHEDOCS=True thing, since this is a hack to avoid running the leflib build. But that should be easy to run.

.github/workflows/daily_ci.yml
```diff
@@ -118,3 +118,22 @@ jobs:
...
+
+      - name: Run Windows Unit Tests
+        run: pytest tests/schema/test_windows_paths.py
```
gadfort, line 139: why not run all tests?

gadfort (changes requested):

siliconcompiler/schema/schema_obj.py
WRansohoff: `.as_posix()` returns a string, not a Path.

tests/core/test_create_cmdline.py
WRansohoff: Python cannot guarantee that the exact original string is returned from a cross-platform path conversion, only that the path representation has not changed.

There's no limit to how many strings could refer to the same path. `tapcell.tcl`, `./tapcell.tcl`, `././tapcell.tcl`, `.//.//tapcell.tcl`, ...

.github/workflows/daily_ci.yml
WRansohoff: I just did it this way because Surelog isn't needed for a unit test, and functional tests cannot run on windows because the other tools are not compatible.

.github/workflows/daily_ci.yml
WRansohoff: OpenROAD is not available on Windows.

WRansohoff:
Closing in favor of #2022

PR #2007, opened 2023-09-29
Fix schema patch version updates

troibe:
**What?**
Read in version from `Schema` instead of `Chip`.
This allows the test to check the version of the `last_major.json`.

**How?**
Updated the schema patch version to demonstrate that it works.
Also renamed `last_major.json` to `last_minor.json` because that is what it captures.

**Why?**
Issue described in https://github.com/siliconcompiler/siliconcompiler/pull/2006


siliconcompiler/schema/schema_cfg.py
```diff
@@ -11,7 +11,7 @@
 except ImportError:
     from siliconcompiler.schema.utils import trim
 
-SCHEMA_VERSION = '0.36.0'
+SCHEMA_VERSION = '0.36.1'
```
gadfort, line 14: can you revert this back to `.0`

gadfort (changes requested):
LGTM, just change the version back to 0.36.0

siliconcompiler/schema/schema_cfg.py
troibe: done

gadfort (approved):

PR #2008, opened 2023-09-29
Add prune

troibe:
**What?**
Extends the functionality in: https://github.com/siliconcompiler/siliconcompiler/pull/1976
The naming is `prune` instead of `exclude` as that is what was the outcome of the siliconcompiler meeting two weeks ago.

**Testing:**
I implemented the specification as best as I could understand it.
The most important part to me and what required changes is that the correctness of the paths in the flowgraph is based on reaching all final steps but within those final steps it doesn't matter how many nodes (at least one) was reached.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -4234,7 +4234,6 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                 # If node is builtin and not all input nodes were pruned, mark visited
                 if self._is_builtin(tool, task) \
                         and self._get_pruned_node_inputs(flow, current_node):
-                    print(self._get_pruned_node_inputs(flow, current_node))
                     nodes_sorted.append(current_node)
                     visited_nodes.add(current_node)
                     current_nodes.remove(current_node)
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1687,7 +1687,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
         for (step, index) in nodes_to_execute:
             in_job = self._get_in_job(step, index)
 
-            for in_step, in_index in self._get_flowgraph_node_inputs(flow, (step, index)):
+            for in_step, in_index in self._get_pruned_node_inputs(flow, (step, index)):
                 if in_job != self.get('option', 'jobname'):
                     workdir = self._getworkdir(jobname=in_job, step=in_step, index=in_index)
                     cfg = os.path.join(workdir, 'outputs', f'{design}.pkg.json')
@@ -1854,7 +1854,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
                 error = True
 
         unreachable_steps = self._unreachable_steps_to_execute(flow)
-        if self._unreachable_steps_to_execute(flow):
+        if unreachable_steps:
             self.logger.error(f'These final steps in {flow} can not be reached: '
                               f'{list(unreachable_steps)}')
             error = True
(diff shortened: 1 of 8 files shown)


troibe:
I'm happy about any further testing we can add to this one as the per node flowgraph combined with the per step success is still about confusing to me.

siliconcompiler/schema/schema_cfg.py
```diff
@@ -2485,6 +2485,18 @@ def schema_option(cfg):
             Inclusive list of steps to end execution with. The default is to go
             to all exit steps in the flow graph.""")
 
+    scparam(cfg, ['option', 'prune'],
+            sctype='[str]',
```
gadfort, line 2489: i think we would want this to be [(str, str)] instead

siliconcompiler/schema/schema_cfg.py
```diff
@@ -2485,6 +2485,18 @@ def schema_option(cfg):
...
+            scope='job',
+            shorthelp="Prune branches starting with",
+            switch="-prune <node>",
+            example=[
+                "cli: -prune 'syn0'",
```
gadfort, line 2494: -prune 'syn 0'

siliconcompiler/schema/schema_cfg.py
```diff
@@ -2485,6 +2485,18 @@ def schema_option(cfg):
...
+                "api: chip.set('option', 'prune', 'syn0')"],
```
gadfort, line 2495: chip.set('option', 'prune', ('syn', '0'))

gadfort (changes requested):
Overall looks good, just needs to switch the type form str to (str, str) in the schema (and fix where it's used)

siliconcompiler/schema/schema_cfg.py
```diff
@@ -2485,6 +2485,18 @@ def schema_option(cfg):
             Inclusive list of steps to end execution with. The default is to go
             to all exit steps in the flow graph.""")
 
+    scparam(cfg, ['option', 'prune'],
+            sctype='[str]',
+            scope='job',
+            shorthelp="Prune branches starting with",
```
gadfort, line 2491: Prune flowgraph...

gadfort:
https://github.com/siliconcompiler/siliconcompiler/actions/runs/6372810647/job/17295978850

siliconcompiler/core.py
```diff
@@ -1853,15 +1853,10 @@ def _check_flowgraph(self, flow=None):
...
-        to_nodes = self._get_execution_exit_nodes(flow)
-        reachable_nodes = set(self._reachable_flowgraph_nodes(flow, set(from_nodes)))
-        unreachable_nodes = set(to_nodes).difference(reachable_nodes)
-        if unreachable_nodes:
-            unreachable_nodes_formated = list(
-                map(lambda node: f'{node[0]}{node[1]}', unreachable_nodes))
-            self.logger.error(f'These final nodes in {flow} can not be reached: '
-                              f'{unreachable_nodes_formated}')
+        unreachable_steps = self._unreachable_steps_to_execute(flow)
+        if self._unreachable_steps_to_execute(flow):
```
nmoroze, line 1857: I think you can reuse `unreachable_steps` here

siliconcompiler/core.py
```diff
@@ -4235,9 +4228,13 @@ def _nodes_to_execute(self, flow, from_nodes, to_nodes):
...
                     current_nodes.update(outputs)
             if current_nodes == current_nodes_copy:
-                raise Exception(
-                    f'Input paths coming from {inputs.difference(visited_nodes)} '
-                    f'to {current_node} are missing.')
+                # Only fail if step was not visited
+                if not any(filter(lambda node: node[0] == current_node[0], visited_nodes)):
+                    raise SiliconCompilerError(
+                        f'Flowgraph connection from {inputs.difference(visited_nodes)} '
+                        f'to {current_node} are missing. '
```
nmoroze, line 4235: nit: are -> is

nmoroze (commented):
Overall looks good! I think it might be good to add a test case that pruning one node that feeds into a builtin min/max still allows the rest of the flow to proceed without that node.

PR #2013, opened 2023-10-01
switch server to basic schema

gadfort:
Adds:
- basic schema based on Schema class for the server (mirror of it's current settings)
- Helper function in Schema to override default schema configuration for subclasses.


siliconcompiler/remote/server.py
```diff
@@ -22,33 +25,34 @@ class Server:
...
-        nfsmount_abspath = [os.path.abspath(nfs_path) for nfs_path in self.cfg['nfsmount']['value']]
-        self.cfg['nfsmount']['value'] = nfsmount_abspath
+        self.schema = ServerSchema(logger=self.logger)
 
         # Set up a dictionary to track running jobs.
         self.sc_jobs = {}
 
+    def run(self):
+        if not os.path.exists(self.nfs_mount):
+            raise FileNotFoundError(f'{self.nfs_mount} could not be found.')
```
WRansohoff, line 53: Could this be a warning, with the directory being created if it doesn't exist?

`nfs_mount` was probably not a great choice for the parameter name; local directories also work when the server is executing jobs locally, as in our CI tests. It might make sense to change the parameter name to something like `jobs_dir` or `job_storage`.

WRansohoff (approved):
Looks like a great change! The current dev server's configuration implementation predates the `Schema` object.

Just one comment, maybe we should rename the `nfsmount` parameter. At this point, it really reflects an accessible location where job data can be stored.

siliconcompiler/remote/server.py
gadfort: ohh, nevermind, I see what you mean, I think as part of reworking the server we can modify this behavior.

PR #2014, opened 2023-10-01
add tcl footer template to be able to embed the sc_root

gadfort:


siliconcompiler/templates/tcl/manifest.tcl.j2
```diff
@@ -12,3 +21,9 @@ proc sc_top {} {
     }
     return $sc_entrypoint
 }
+
+{% if scroot  %}
```
troibe, line 25: Is there any scenario where we wouldn't want to have this?
Just asking because if it's conditional then we will always have to check if it's present when using it later.

troibe (approved):
Looks good but I think it would be better if scroot would always be set in tcl.

PR #2015, opened 2023-10-02
fix typo in sc-remote app

gadfort:


troibe (approved):
Yep now it's consistent with the other apps.

PR #2021, opened 2023-10-03
read schema manifest despite missing keys

gadfort:
This change will allow us to open older schemas without failing out.
This is useful when attempting review older runs, but the schema has changed since that run and now.


troibe (approved):
Looks good.
I'm not sure if it already makes sense here but we should add a test case for this scenario.

PR #2041, opened 2023-10-12
Replace utils.copytree() with shutil.copytree()

troibe:
**Why?**
We [deprecated python 3.6 and 3.7 ](https://github.com/siliconcompiler/siliconcompiler/commit/84ce9c1f060cbce78877ea55365f7d3e0dbb13ed)so now we can rely on `shutil.copytree()`.

**How?**
Using `shutil.copytree()` and its `copy_function` and `ignore` parameters we can achieve the same functionality as `utils.copytree()`.

### DIFF
--- a/siliconcompiler/utils.py
+++ b/siliconcompiler/utils.py
@@ -8,45 +8,16 @@ from pathlib import Path
 PACKAGE_ROOT = os.path.dirname(os.path.abspath(__file__))
 
 
-def copytree(src, dst, ignore=[], dirs_exist_ok=False, link=False):
-    '''Simple implementation of shutil.copytree to give us a dirs_exist_ok
-    option in Python < 3.8.
-
-    If link is True, create hard links in dst pointing to files in src
-    instead of copying them.
-    '''
-    os.makedirs(dst, exist_ok=dirs_exist_ok)
-
-    for name in os.listdir(src):
-        if name in ignore:
-            continue
-
-        srcfile = os.path.join(src, name)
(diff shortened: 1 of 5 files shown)


siliconcompiler/utils.py
```diff
@@ -8,45 +8,16 @@
...
-    for name in os.listdir(src):
-        if name in ignore:
-            continue
-
-        srcfile = os.path.join(src, name)
-        dstfile = os.path.join(dst, name)
-
-        if os.path.islink(srcfile):
-            # Get the true filepath if its a link
-            srcfile = os.path.realpath(srcfile)
```
gadfort, line 29: I think you still need this in the `link_symlink_copy` code

gadfort (changes requested):

siliconcompiler/utils.py
troibe: Just to make sure I'm getting this right.
`os.path.islink()` can only detect symlinks afaik so the behavior of `utils.copytree()` was this:

If the `srcfile` is a symlink then we copy over the `real` file when `link` is `False`.
If `link` is true we convert the symlink into a hardlink to the `real` file, if that does not work we try sticking with a `symlink`, if that also fails we copy over the `real` file.
Is this the behavior we want?

Python's internal `copytree()` behavior with `symlinks=True` is to recreate the symlink. With `symlinks=False` it copies the `real` file behind the symlink.

gadfort (approved):

PR #2043, opened 2023-10-12
Add line numbers to parsed logs

troibe:
**Why?**
This makes it easier to figure out where the warning and errors come from in the original log.

@gadfort I added it to infos and the logger output as well.
Do you think that's a good idea or should I remove it?

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -2757,7 +2757,7 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
             right_align = len(str(line_count))
             # Start at the beginning of file again
             f.seek(0)
-            for num, line in enumerate(f, 1):
+            for num, line in enumerate(f, start=1):
                 for suffix in checks:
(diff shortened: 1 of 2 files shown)


siliconcompiler/core.py
```diff
@@ -2764,15 +2764,15 @@ def check_logfile(self, jobname=None, step=None, index='0',
                     if string is not None:
                         matches[suffix] += 1
                         # always print to file
-                        print(string.strip(), file=checks[suffix]['report'])
+                        print(f'{num}: {string.strip()}', file=checks[suffix]['report'])
```
gadfort, line 2767: I think just format the line once.
Also, I think it would be helpful if you right align the numbers (you can probably prefetch the total number of lines in the file to help with that:
```
  1: blah
 11: blah
111: blah blah
```

gadfort (changes requested):

gadfort (approved):

PR #2046, opened 2023-10-14
only run release artifact on release flows

gadfort:
https://github.com/siliconcompiler/siliconcompiler/actions/runs/6519197312


troibe (approved):
Ah sorry didn't see that that wasn't part of the same `if` predicate as the one I added when reviewing. Hope you didn't have to manually remove too much.

gadfort:
it just fails luckily

PR #2049, opened 2023-10-16
Completly switch to Python internal symlinking

troibe:
**What?**
Cleanup based on discussion here: https://github.com/siliconcompiler/siliconcompiler/pull/2041#discussion_r1357214986

**Why & How?**
We should not try changing symlinks into hardlinks like we currently do.
With the `symlinks=True` the behavior of `shutil.copytree()` is to recreate the symlink (with the same metadata) instead of copying over the content.


troibe:
After discussing this we decided we do want to replace hardlinks with symlinks instead of copying them if we can't create another hardlink to the file.
The python behavior would be to copy the hardlink content.
So abandoning this now.

PR #2057, opened 2023-10-24
Error and warning parsing should split the print of errors and warnings

troibe:
**What?**
Order matched suffixes in logfiles from tools as follows `[..., 'warnings', 'errors']`.
Also show a small summary of the number of each suffix.

**Why?**
Improves readability of log output.

**How?**
Reorder the way the log files are scanned.

**Example output:**
```
| WARNING | job0 | export | 0 | 28: [WARNING] no fill config file specified
| INFO    | job0 | export | 0 | Number of warnings: 1
| INFO    | job0 | export | 0 | Number of errors: 0
```

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -2751,14 +2751,22 @@ If you are sure that your working directory is valid, try running `cd $(pwd)`.""
             checks[suffix]['args'] = regexes
             matches[suffix] = 0
 
+        # Order suffixes as follows: [..., 'warnings', 'errors']
+        ordered_suffixes = list(filter(lambda key:
+                                       key not in ['warnings', 'errors'], checks.keys()))
+        if 'warnings' in checks:
+            ordered_suffixes.append('warnings')
+        if 'errors' in checks:
+            ordered_suffixes.append('errors')
+
         # Looping through patterns for each line
         with open(logfile, errors='ignore_with_warning') as f:
             line_count = sum(1 for _ in f)
             right_align = len(str(line_count))
             # Start at the beginning of file again
             f.seek(0)
-            for num, line in enumerate(f, start=1):
-                for suffix in checks:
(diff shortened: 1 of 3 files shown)


siliconcompiler/core.py
```diff
@@ -2779,7 +2787,9 @@ def check_logfile(self, jobname=None, step=None, index='0',
                             else:
                                 self.logger.info(f'{suffix}: {line_with_num}')
 
-        for suffix in checks:
+        for suffix in ordered_suffixes:
+            if display:
+                self.logger.info(f'Number of {suffix}: {matches[suffix]}')
```
gadfort, line 2792: I think we still want the summary, just not the details

siliconcompiler/core.py
```diff
@@ -2751,14 +2751,22 @@ def check_logfile(self, jobname=None, step=None, index='0',
...
             line_count = sum(1 for _ in f)
             right_align = len(str(line_count))
-            # Start at the beginning of file again
-            f.seek(0)
-            for num, line in enumerate(f, start=1):
-                for suffix in checks:
+            for suffix in ordered_suffixes:
+                # Start at the beginning of file again
+                f.seek(0)
+                for num, line in enumerate(f, start=1):
```
gadfort, line 2769: Can you make it such that warning lines print before the error lines.

gadfort (changes requested):

siliconcompiler/core.py
troibe: That's already the case as is.
I adjusted the test now to make it more obvious.
We now have an error in the tool log before a warning.
In the siliconcompiler log the warning comes before the error.

siliconcompiler/core.py
troibe: now printing regardless of `display`

gadfort (approved):

PR #2058, opened 2023-10-26
ensure klayout honors save format

gadfort:


siliconcompiler/tools/klayout/klayout_export.py
```diff
@@ -45,6 +45,8 @@
 def gds_export(design_name, in_def, in_files, out_file, tech, allow_missing, config_file='',
                seal_file='',
                timestamps=True):
+    from tools.klayout.klayout_utils import get_write_options  # noqa E402
```
troibe, line 48: What is the reason for the dynamic import here and in the other file?

troibe (approved):
Looks good.
For documentation purposes which issue where you running into that this fixes?

gadfort:
Without this, klayout always saves a GDS, regardless of the datatype requested.

siliconcompiler/tools/klayout/klayout_export.py
gadfort: klayout seems to need this, we can try to revisit later since this might be fixed with a full path to the package

PR #2060, opened 2023-10-26
Remove valid bit from schema

troibe:
**What?**
Remove the valid bit from the schema. 
Change the `patch` number in the schemaversion because it was never implemented to begin with.

**Why?**
Now we have `-prune` to specify which paths/nodes in the flowgraph will not be run.
Therefore attaching the valid bit to nodes in the flowgraph is not necessary.


aolofsson (approved):
Makes sense!

gadfort (approved):

PR #2061, opened 2023-10-26
fix openroad routing script to correctly set unidirectional layers

gadfort:


siliconcompiler/tools/openroad/scripts/sc_route.tcl
```diff
@@ -30,13 +30,13 @@ insert_fillers
...
   foreach via [dict exists $sc_cfg tool $sc_tool task $sc_task var drt_default_via] {
+    utl::info FLW 1 "Marking $via a default routing via"
     detailed_route_set_default_via $via
   }
 }
-if {[dict exists $sc_cfg tool $sc_tool task $sc_task var drt_unidirectional_layer]} {
-  foreach layer [dict exists $sc_cfg tool $sc_tool task $sc_task var drt_unidirectional_layer] {
-    detailed_route_set_unidirectional_layer $via
-  }
+foreach layer $openroad_drt_unidirectional_layers {
```
troibe, line 37: So `sc_apr.tcl` is always being run before `sc_route.tcl` and we can rely on this being at least an empty list?

troibe (approved):

siliconcompiler/tools/openroad/scripts/sc_route.tcl
gadfort: Yes

PR #2063, opened 2023-10-30
Fix schema variable in sc_apr.tcl

SeyiAyoARL:
Use the proper schema variable (detail_routing_unidirectional_layer) vs the one currently being used (drt_unidirectional_layer)


gadfort (approved):
Thanks for catching that

PR #2065, opened 2023-11-01
add support for RCX file generation for openroad

gadfort:


siliconcompiler/flows/generate_openroad_rcx.py
```diff
@@ -0,0 +1,70 @@
...
+            flow.edge(flowname, 'bench', 'pex', head_index=corner, tail_index=0)
+
+    prevstep = None
+    for step, task in pipe:
+        if not task:
+            continue
+        flow.node(flowname, step, task)
+        if prevstep:
+            flow.edge(flowname, prevstep, step)
+        prevstep = step
```
troibe, line 62: Propably the for loop is for later extensibility so in that case keep it.
Otherwise it could be simplified to:
```suggestion
    flow.node(flowname, 'bench', rcx_bench)
```

siliconcompiler/tools/openroad/rcx_bench.py
```diff
@@ -0,0 +1,94 @@
...
+    '''
+
+    openroad.setup(chip)
+
+    step = chip.get('arg', 'step')
+    index = chip.get('arg', 'index')
+    tool, task = chip._get_tool_task(step, index)
+
+    chip.set('tool', tool, 'task', task, 'script', 'sc_rcx.tcl', step=step, index=index)
+    chip.set('tool', tool, 'task', task, 'threads', 1, step=step, index=index)
```
troibe, line 15: Just because we were talking about the `threads` in the meeting today. Does this actually do anything here?

gadfort: it will only open OpenROAD with a single thread, this task doesn't support multi threading

siliconcompiler/flows/generate_openroad_rcx.py
gadfort: good call

troibe (approved):
I think everything is properly copied over.
We know this works from internal testing...
We can add a test case in the future.

PR #2074, opened 2023-11-02
Initial dependency support

troibe:
**What?**
- Intitial necessary schema changes to support dependencies (`path`, `name`, `commitid`)
- Logic to download files from a variety of sources
- Saves dependency data to `~/.sc/cache/{name}-{commitid}`

**Why?**
This will allow in the next step to set file names relative to a dependency. No more searching for files or ambiguities.

**Testing:**
All the currently supported paths are tested in `test_dependency.py`


siliconcompiler/dependency.py
```diff
@@ -0,0 +1,170 @@
...
+            'path': e.g. 'ssh://github.com/xyz/xyz/'
+            'commitid': e.g. '3b94aa80506d25d5388131e9f2ecfcf4025ca866',
+            'name': 'your-dependency'
+            or
+            'path': e.g. 'https://github.com/xyz/xyz/archive/',
+            'commitid': e.g. '3b94aa80506d25d5388131e9f2ecfcf4025ca866',
+            'name': 'your-dependency'
+            or
+            'path': e.g. 'https://zeroasic.com/xyz.tar.gz',
+            'name': e.g. 'your-dependency-version-1'
```
gadfort, line 45: is this just a string now?

siliconcompiler/schema/schema_cfg.py
```diff
@@ -2988,6 +2991,53 @@ def schema_option(cfg):
...
+            sctype='str',
+            scope='global',
+            shorthelp="Dependency name",
+            switch="-dependency_name 'dependency <str>'",
+            example=[
+                "cli: -dependency_name 'freepdk45_data freepdk45'",
+                "api: chip.set('dependency', 'freepdk45_data', 'name', 'freepdk45')"],
+            schelp="""
+            Dependency name in cache: name-commitid
+            Should be unique if no commitid specified""")
```
gadfort, line 3036: is this needed if the `dependency` is also a name?

siliconcompiler/schema/schema_cfg.py
```diff
@@ -2988,6 +2991,53 @@ def schema_option(cfg):
...
+    scparam(cfg, ['dependency', dependency, 'commitid'],
+            sctype='str',
+            scope='global',
+            shorthelp="Dependency source commitid",
+            switch="-dependency_commitid 'dependency <str>'",
+            example=[
+                "cli: -dependency_commitid 'freepdk45_data 07ec4aa'",
+                "api: chip.set('dependency', 'freepdk45_data', 'commitid', '07ec4aa')"],
+            schelp="""
+            Dependency source commitid""")
```
gadfort, line 3024: maybe `version` is a bit more generic? I'm not sure about that, just a suggestion.

siliconcompiler/schema/schema_cfg.py
troibe: Just using the name of the dependency would also work of course.
Again might be confusing with the python package name.
For example so far I was using `siliconcompiler_data` as the python package name for the metadata but `siliconcompiler` as the name for the actual content.

siliconcompiler/schema/schema_cfg.py
gadfort: ref is probably good

siliconcompiler/dependency.py
troibe: Yep if it just points to some random tar file on the internet.
In that case we don't have any chance to verify the `name` or `commitid`.
We still need to have a name of where we want to store the cached files.

siliconcompiler/schema/schema_cfg.py
troibe: Then nevermind I can just change it.

tests/dependency/test_dependecy.py
```diff
@@ -0,0 +1,58 @@
+import siliconcompiler
```
gadfort, line 1: rename this file to `test_dependency.py`

gadfort (changes requested):
LGTM, just one minor fix

gadfort (approved):
LGTM

PR #2078, opened 2023-11-05
Remove scpath

troibe:
**What?**
Removes `SCPATH` from schema to rely on dependency management.
Also adjust testcases that relied on SCPATH.

**Why?**
With our new dependency management `SCPATH` is not necessary anymore.

**How?**
For now we since we don't have a public `siliconcompiler_data` or `freepdk45_data` etc. we manually set the dependency data for the targets.
Once we have these up on pip we can remove that.


siliconcompiler/core.py
```diff
@@ -415,6 +416,13 @@ def load_target(self, module, **kwargs):
             Loads the 'freepdk45_demo' target with 5 parallel synthesis tasks
         """
 
+        self.set('dependency', 'siliconcompiler_data',
+                 'path', 'git+https://github.com/siliconcompiler/siliconcompiler')
+        self.set('dependency', 'siliconcompiler_data',
+                 'name', 'siliconcompiler')
+        self.set('dependency', 'siliconcompiler_data',
+                 'commitid', 'dependency-caching-rebase')
+
```
gadfort, line 425: I think we should set this or something like it in the `__init__` method

siliconcompiler/core.py
```diff
@@ -826,7 +835,7 @@ def unset(self, *keypath, step=None, index=None):
             self.logger.debug(f'Failed to unset value for {keypath}: parameter is locked')
 
     ###########################################################################
-    def add(self, *args, field='value', step=None, index=None):
+    def add(self, *args, field='value', step=None, index=None, dependency=''):
```
gadfort, line 838: same

siliconcompiler/core.py
```diff
@@ -858,12 +867,12 @@ def add(self, *args, field='value', step=None, index=None):
...
 
         try:
-            self.schema.add(*args, field=field, step=step, index=index)
+            self.schema.add(*args, field=field, step=step, index=index, dependency=dependency)
         except (ValueError, TypeError) as e:
             self.error(str(e))
 
     ###########################################################################
-    def input(self, filename, fileset=None, filetype=None, iomap=None):
+    def input(self, filename, fileset=None, filetype=None, iomap=None, dependency=''):
```
gadfort, line 875: same

siliconcompiler/core.py
```diff
@@ -935,14 +944,14 @@ def _add_input_output(self, category, filename, fileset, filetype, iomap):
         elif not fileset:
             self.logger.info(f'{filename} inferred as fileset {use_fileset}')
 
-        self.add(category, use_fileset, use_filetype, filename)
+        self.add(category, use_fileset, use_filetype, filename, dependency=dependency)
```
gadfort, line 947: same

siliconcompiler/core.py
```diff
@@ -1123,7 +1132,13 @@ def _find_files(self,
...
             search_paths = self.__convert_paths_to_posix(search_paths)
 
-        for path in paths:
+        while len(dependencies) < len(paths):
+            dependencies.append(None)
+
+        for (dependency, path) in zip(dependencies, paths):
+            if dependency:
+                result.append(os.path.join(dep.path(self, dependency), path))
+                continue
```
gadfort, line 1141: we need some error messaging here is the file is not found in the declared package

siliconcompiler/libs/asap7sc7p5t.py
```diff
@@ -14,24 +14,28 @@ def _setup_lib(chip, libname, suffix):
...
 
     # todo: remove later
     lib.set('option', 'pdk', 'asap7')
 
     # timing
     for corner_name, lib_corner in corners.items():
         lib.add('output', corner_name, 'nldm',
-                libdir + '/nldm/' + libname + '_' + lib_corner + '.lib.gz')
+                libdir + '/nldm/' + libname + '_' + lib_corner + '.lib.gz',
+                dependency='siliconcompiler_data')
```
gadfort, line 26: It might be nice to extend the Library/PDK classes to automatically attach the package information if not declared.

siliconcompiler/schema/schema_cfg.py
```diff
@@ -112,9 +112,11 @@ def scparam(cfg,
             cfg['node']['default']['default']['date'] = []
             cfg['node']['default']['default']['author'] = []
             cfg['node']['default']['default']['filehash'] = []
+            cfg['node']['default']['default']['dependency'] = []
```
gadfort, line 115: package, also need to add this to the schema documentation

siliconcompiler/schema/schema_obj.py
```diff
@@ -187,7 +187,7 @@ def __get(self, *keypath, field='value', job=None, step=None, index=None):
             raise ValueError(f'Invalid field {field}')
 
     ###########################################################################
-    def set(self, *args, field='value', clobber=True, step=None, index=None):
+    def set(self, *args, field='value', clobber=True, step=None, index=None, dependency=''):
```
gadfort, line 190: it might be nice if the schema didn't have this and instead relied on Chip class (there is some awkward integration here that requires the same edits in several places).

siliconcompiler/sphinx_ext/dynamicgen.py
```diff
@@ -218,33 +218,14 @@ def run(self):
...
-        since we can't be sure that's happened yet. Therefore, we have to check
-        each one explicitly.
-
-        However, this could result in duplicate modules being detected once the
-        SCPATH does get updated. Therefore, we check to ensure that SCPATH
-        directories are not equal to the builtins directory before searching it.
-
-        TODO: we want better duplicate resolution (in case the user explicitly
-        declares a duplicate tool), where SCPATH takes priority.
+        This function explicitly searches builtins.
```
gadfort, line 221: some functionality like this might be useful to preserve

tests/core/test_check_manifest.py
```diff
@@ -95,8 +94,7 @@ def test_check_allowed_filepaths_fail(scroot, monkeypatch):
     os.chdir(cwd)
 
     env = {
-        'SC_VALID_PATHS': os.path.join(scroot, 'third_party', 'pdks'),
-        'SCPATH': os.environ['SCPATH']
+        'SC_VALID_PATHS': os.path.join(scroot, 'third_party', 'pdks')
```
gadfort, line 97: is this used somewhere?

troibe: Seems it's fine to remove `SC_VALID_PATHS` too.

siliconcompiler/libs/asap7sc7p5t.py
troibe: If a pdk just is allowed to have one dependency then that sounds like a good idea.

siliconcompiler/core.py
troibe: Added

siliconcompiler/libs/asap7sc7p5t.py
troibe: Adding it like this `PDK(chip, name, dependency='siliconcompiler_data')`.

siliconcompiler/schema/schema_obj.py
troibe: Yes but then I need to inspect the `cfg['type']` of the keypath in `Chip`.
If you think that would be prettier I can do that too.

siliconcompiler/core.py
troibe: I don't think that makes much sense as this is only required for targets.
Otherwise the tests right fully complain that we have set a default that does not correspond to the schema default.
We could set the siliconcompiler as a default in the schema which I tried but couldn't get to work. But I can look more into that.

siliconcompiler/schema/schema_cfg.py
troibe: Sure but wouldn't this be confusing to the user?
The schema will say this file is part of package xyz but then the `package` information will be in the `dependency` section of the schema...

siliconcompiler/core.py
gadfort: this feels a little hacky here, lets leave it and figure it out before merging

siliconcompiler/schema/schema_cfg.py
gadfort: we will likely need to make some more changes to the the schema to be self consistent, but it should be `package`

siliconcompiler/schema/schema_obj.py
gadfort: lets leave this for now

siliconcompiler/sphinx_ext/dynamicgen.py
gadfort: we use this for out internal documentation generation

troibe: Where should we search instead of `SCPATH` then though?
All dependencies?

gadfort: We'll figure it out

siliconcompiler/core.py
```diff
@@ -783,6 +783,8 @@ def set(self, *args, field='value', clobber=True, step=None, index=None, depende
                 on a per-node basis.
             index (str): Index name to set for parameters that may be specified
                 on a per-node basis.
+            package (str): Package that this file/dir dependens on. Available packages
```
gadfort, line 786: dependens -> depends

siliconcompiler/core.py
```diff
@@ -857,6 +859,8 @@ def add(self, *args, field='value', step=None, index=None, dependency=''):
                 on a per-node basis.
             index (str): Index name to modify for parameters that may be specified
                 on a per-node basis.
+            package (str): Package that this file/dir dependens on. Available packages
```
gadfort, line 862: same

siliconcompiler/core.py
```diff
@@ -1137,7 +1141,10 @@ def _find_files(self,
 
         for (dependency, path) in zip(dependencies, paths):
             if dependency:
-                result.append(os.path.join(dep.path(self, dependency), path))
+                depdendency_path = os.path.join(dep.path(self, dependency), path)
+                if not os.path.exists(depdendency_path) and not missing_ok:
+                    self.error(f'Could not find {depdendency_path} in {dependency}.')
```
gadfort, line 1146: just do the path component in this case

gadfort (changes requested):
Looks good, just a minor set of changes

siliconcompiler/core.py
troibe: You mean just check that path?

gadfort (approved):
LGTM

PR #2079, opened 2023-11-07
add data retention to wheels build since we add artifacts to releases

gadfort:


troibe (approved):
For reference the default was 90 days

PR #2081, opened 2023-11-08
correctly read in empty modules in yosys

gadfort:
Fixes issue when empty modules was present in RTL. Declared blackboxes are handled separately in syn_asic.tcl


troibe (approved):
Looks reasonable.
Where were we running into this?

gadfort:
Yeah, attempting to run the tapeout through this was triggering these issues.

PR #2088, opened 2023-11-13
use package for data files in SC

gadfort:


siliconcompiler/core.py
```diff
@@ -1163,17 +1163,17 @@ def _find_files(self,
             dependencies.append(None)
 
         for (dependency, path) in zip(dependencies, paths):
+            if not search_paths:
```
troibe, line 1166: Why would you give this higher priority?

siliconcompiler/core.py
```diff
@@ -81,6 +81,9 @@ def __init__(self, design, loglevel=None):
 
         self.schema = Schema(logger=self.logger)
 
+        self.register_package_source('siliconcompiler',
+                                     'python://siliconcompiler')
```
troibe, line 85: Does this work if python is installed from pip?

gadfort: yes, just gets the location that siliconcompiler is installed to pip or git works

siliconcompiler/core.py
gadfort: because the imported_file_paths need higher priority

siliconcompiler/core.py
troibe: Also if we call `register_sc_data_source(chip)` in every dependency why do we need this at all?

tests/core/test_read_manifest.py
```diff
@@ -23,25 +23,14 @@ def test_read_manifest_fields():
...
-
-
 # Use nostrict mark to prevent changing default value of [option, strict]
 @pytest.mark.nostrict
 def test_modified_schema(datadir):
     '''Make sure schema has not been modified without updating defaults.json'''
 
     # gets default from schema
     chip = siliconcompiler.Chip('test')
+    chip.schema._remove('dependency', 'siliconcompiler')
```
troibe, line 33: This seems to be a bit the idea of this test but I know it fails with the dependency....

tests/core/test_read_manifest.py
```diff
@@ -62,6 +51,7 @@ def test_modified_schema(datadir):
 @pytest.mark.nostrict
 def test_last_schema(datadir):
     chip = siliconcompiler.Chip('test')
+    chip.schema._remove('dependency', 'siliconcompiler')
```
troibe, line 54: Same thing
This seems to be a bit the idea of this test but I know it fails with the dependency....

siliconcompiler/utils.py
```diff
@@ -239,3 +240,9 @@ def default_credentials_file():
     cfg_file = os.path.join(Path.home(), '.sc', 'credentials')
 
     return cfg_file
+
+
+def register_sc_data_source(chip):
+    chip.register_package_source('siliconcompiler_data',
+                                 'git+https://github.com/siliconcompiler/siliconcompiler',
+                                 'v'+sc_version)
```
troibe, line 248: This shouldn't matter much because we don't plan on changing the PDKs and LIBs,
but if we do want to edit them it might be confusing to have to do a release first.

tests/core/test_create_cmdline.py
```diff
@@ -208,6 +208,7 @@ def _mock_read_manifest(chip, manifest, **kwargs):
     monkeypatch.setattr(siliconcompiler.Schema, 'read_manifest', _mock_read_manifest)
 
     chip = siliconcompiler.Chip('test')
+    chip.schema._remove('dependency', 'siliconcompiler')
```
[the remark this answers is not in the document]
gadfort: when we call `allkeys` this will be present, but fail the test because it doesn't match the example in the command line check

tests/core/test_read_manifest.py
gadfort: I'll need to double check this one

troibe (approved):
Initially I was confused because I thought you were setting the `siliconcompiler_data` to `python://siliconcompiler` everywhere which probably wouldn't work.
But since you overwrite that with the path to the cloned git repo whenever you are using a PDK or LIB I think this should be fine.
I'm not sure if we need the `python://siliconcompiler` in the `__init__` of `Chip`.

tests/core/test_read_manifest.py
gadfort: rewrote the test

siliconcompiler/utils.py
gadfort: this will eventually become one of the data repos. Using SC is just a temporary solution until they are ready.

gadfort:
@troibe adding the 'siliconcompiler' package in the init just ensures it's registered without the need for us to do it in tons of places once we can transition tools, etc to use them

PR #2090, opened 2023-11-14
create docker runner labeled latest

gadfort:
provide basic runner image


troibe (approved):
I think this makes sense.
The only thing confusing to me is why this is a separate docker container.
Couldn't we just label `sc_tools` with `latest` on release in addition to its other label?

gadfort:
right now we use sc_tools for ci testing so it it easier to avoid adding the siliconcompiler install to it since it depends on which commit that is running on , the sc_runner is pegged to the last release so we can safely install it there.

PR #2091, opened 2023-11-15
improve sdc parsing for yosys clock period extraction

gadfort:
This is needed if the clock period is specified with a variable instead of a number


siliconcompiler/tools/yosys/syn_asic.py
```diff
@@ -390,11 +390,33 @@ def get_abc_period(chip):
             with open(sdc, 'r') as f:
                 lines = f.read().splitlines()
 
+            # collect simple varaiables in case clock is specified with a variable
```
troibe, line 393: typo: varaiables -> variables

troibe (approved):
This makes sense under the assumption that one line of tcl only has one `set` and `create_clocks`.

PR #2093, opened 2023-11-15
switch from dependency to package

gadfort:


.github/workflows/python_ci.yml
```diff
@@ -29,7 +29,9 @@ jobs:
         if: matrix.os == 'windows-latest'
         run: |
           choco install -y graphviz
-
+          New-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem" -Name "LongPathsEnabled" -Value 1 -PropertyType DWORD -Force
```
troibe, line 32: I did see random failures on Windows if it was due to long paths then thanks so much for fixing that.

siliconcompiler/package.py
```diff
@@ -12,74 +12,48 @@
...
-            or
-            'path': e.g. 'https://github.com/xyz/xyz/archive/',
-            'ref': e.g. '3b94aa80506d25d5388131e9f2ecfcf4025ca866'
-            or
-            'path': e.g. 'https://zeroasic.com/xyz.tar.gz',
-            'ref': e.g. 'your-dependency-version-1'
-            or
-            'path': e.g. 'python://siliconcompiler',
+        package (str): Name of the data source
+        quiet (boolean): Suppress error messages
```
troibe, line 19: Suppress package found messages

tests/package/test_package.py
```diff
@@ -1,17 +1,26 @@
...
+from siliconcompiler import package
 from pathlib import Path
 import pytest
 import logging
+import os
 
 
-def dependency_cache_path(path, ref, chip=None, quiet=True):
+@pytest.fixture(autouse=True)
+def mock_home(monkeypatch):
```
troibe, line 10: What were your thoughts for adding this?

gadfort: I prefer the tests not modify my local setup and run in the pytest directory

troibe: Sounds reasonable. Should we do that globally then?
Any test that uses a target will download or use the cache as well.

gadfort: Good point, can you handle that? just push the changes to this PR

PR #2094, opened 2023-11-16
switch tools refdir to use siliconcompiler package

gadfort:
Removes searching of scroot from SC and relies only on the packaging


troibe (approved):
Nice. Seems like the last missing piece.

PR #2102, opened 2023-11-18
add option/cache to schema

gadfort:
adds control of cache location, this is needed for the server to operate properly.

### DIFF
--- a/siliconcompiler/package.py
+++ b/siliconcompiler/package.py
@@ -8,6 +8,7 @@ import importlib
 import shutil
 from time import sleep
 from siliconcompiler import SiliconCompilerError
+from siliconcompiler.utils import default_cache_dir
 
 
 def path(chip, package, quiet=True):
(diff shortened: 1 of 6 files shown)


siliconcompiler/package.py
```diff
@@ -44,7 +45,13 @@ def path(chip, package, quiet=True):
         return path
 
     # location of the python package
-    cache_path = os.path.join(Path.home(), '.sc', 'cache')
+    cache_path = chip.get('option', 'cache')
+    if cache_path:
+        cache_path = chip.find_files('option', 'cache', missing_ok=True)
+        if not cache_path:
```
troibe, line 51: So can cache path can be both a relative and absolute path?

siliconcompiler/schema/schema_cfg.py
```diff
@@ -2187,6 +2187,20 @@ def schema_option(cfg):
...
+            sctype='file',
+            scope='job',
+            shorthelp="User cache directory",
+            switch="-cache <file>",
+            example=[
+                "cli: -cache /home/user/.sc/cache",
+                "api: chip.set('option', 'cache', '/home/user/.sc/cache')"],
+            schelp="""
+            Filepath to cache used for package data sources. If the
+            credentials parameter is empty, the remote processing client program
```
troibe, line 2200: The credentials part might be a copy paste error?

tests/package/test_package.py
```diff
@@ -3,10 +3,15 @@
...
+import os
 
 
-def cache_path(path, ref, chip=None, quiet=True):
+def cache_path(path, ref, chip=None, quiet=True, cache=None):
     chip = chip or siliconcompiler.Chip('test')
+    chip.set('option', 'cache', cache)
+
+    if not cache:
+        cache = Path.home().joinpath('.sc/cache')
```
troibe, line 14: Can't hurt to use `default_cache_dir()` here as well.

troibe (approved):
Looks good. Two minor minor suggestions.

siliconcompiler/package.py
gadfort: yes don't see why not.

PR #2104, opened 2023-11-18
update to Ubuntu22 for tools builder

gadfort:
Update all images and build scripts to be compatible with ubuntu 22.04


troibe (commented):
I think this all looks good but let's wait until this is synced with the siliconcompiler remote server.

gadfort:
remote is ready to migrate to new OS

PR #2106, opened 2023-11-21
Link server test

troibe:
Enabled the server test again and linked the `ref` and `sc-ref`.


troibe:
`ref` needs to be changed back to main before merging.

.github/workflows/server_ci.yml
```diff
@@ -1,8 +1,6 @@
 on:
   workflow_dispatch:
   push:
-    branches:
-      - main
```
gadfort, line 5: can we make that on pull_request? and switch this be in the tools ci so the tools image is correctly built before it's needed?

gadfort (changes requested):

gadfort (approved):

PR #2110, opened 2023-11-28
correct find files behavior for packages to reject missing files

gadfort:


troibe (approved):
Makes sense. Probably we handle `None` better than non existing paths.

PR #2113, opened 2023-11-30
Add TCL linting

nmoroze:
This PR adds TCL linting using the newly released [tclint](https://github.com/nmoroze/tclint)! 

Right now the checks are mostly related to style. A few things to highlight:

**1:** The one functional-ish fix is in `sc_write_images.tcl`: 

```tcl
catch "analyze_power_grid -net $net -corner $corner" err
``` 

is changed to 

```tcl
catch {analyze_power_grid -net $net -corner $corner} err
```

 which will act differently if either `$net` or `$corner` contains spaces (I think the version with braces is what one intends to do in general, though). 

**2:** I disabled indent checking, since the OR scripts seem to use (mostly) 2 spaces, and the other tools mostly 4 spaces. I'd be happy to open a follow-on PR that adds indent checking in either style, but I'm not sure which you prefer. I'm also planning to add the ability to tweak configuration on a per-file(set) basis, so if you wanted different tools to use a different style, I could prioritize that feature.

**3:** One of the biggest violation categories was long lines, particularly on sets reading from the schema and particularly in sc_apr.tcl. In general I adopted a style of:

```tcl
set name \
   [dict get ...]
```

In a few places I did rewrite code a bit more heavily to make things cleaner. In `sc_apr.tcl`, I created a few intermediate dictionaries when there was a long block with a common prefix, and in `sc_constraints.sdc` I pulled out some of the string formatting into separate variables. 

**4:** There were also a decent number of argument spacing changes. In general `tclint` wants one space between commands/arguments, but I did add an exception that allows aligned set blocks if all values are aligned. However, it only considers contiguous blocks of sets (comments or whitespace will break these). I'd be open to loosening that restriction a bit too. Note that `tclint` doesn't yet parse expressions, so spacing inside e.g. if statement conditions aren't checked yet. 

Please let me know if you have any feedback! No worries at all if you don't think this is ready for integration into the project, I figured a PR would be a good way to demo what I've got.


.github/workflows/lint.yml
```diff
@@ -42,3 +42,15 @@ jobs:
...
+  lint_tcl:
+    name: Lint TCL scripts
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v3
+
+      - name: Lint with tclint
+        run: |
+          python -m pip install tclint==0.1.0
```
gadfort, line 55: can you add this to the testing requirements for sc in the requirements.txt?

.github/workflows/lint.yml
```diff
@@ -42,3 +42,15 @@ jobs:
         if: success() || steps.check-format.conclusion == 'failure'
         run: |
           verible-verilog-lint --rules_config .github/workflows/config/verible.rules `cat files.txt`
+
+  lint_tcl:
+    name: Lint TCL scripts
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v3
```
gadfort, line 51: use v4

tclint.toml
```diff
@@ -0,0 +1,11 @@
+exclude = [
+    'third_party',
+    'siliconcompiler/tools/yosys/syn_strategies.tcl'
```
gadfort, line 3: too messy to include?

gadfort (changes requested):
Looks good overall

tclint.toml
nmoroze: yeah, it's very closely adapted from another file and I don't think it's very well-tested, so didn't want to hack it up. I'll add a comment.

.github/workflows/lint.yml
nmoroze: done, I also tweaked the workflow to use `-c requirements.txt` to install just tclint based on the version in requirements.txt and avoid duplication (without requiring all requirements to be installed). 

https://pip.pypa.io/en/stable/user_guide/#constraints-files

gadfort (approved):

PR #2116, opened 2023-12-04
Add expr to unwrapped Tcl return expressions

nmoroze:
I think I caught a subtle bug in a few Tcl procs caused by supplying an expression directly to `return`. 

It seems like Tcl will always return the last word of the expression:

```tcl
proc foo {} {
  return 0 && 1
}

puts [foo]
```

prints 1 for me. This can be fixed by adding an `expr` command to wrap the expression.

`tclint` 0.1.2 now catches this issue.


gadfort (approved):
Thanks

PR #2120, opened 2023-12-08
Enable tclint indent checks and fix violations

nmoroze:
This uses the new `tclint` fileset configuration feature to set an alternate indent for the OpenROAD scripts, which made it a lot quicker to clean up. There were 2 files in the OpenROAD scripts that used 4 space indent, but I transitioned them over in this PR.


gadfort (approved):
Thanks.

PR #2128, opened 2023-12-13
fix klayout

gadfort:
Switch to importing schema module by file to avoid name collision with other `schema` module.
Fix wheels to correctly install klayout again and re-enable those tests


gadfort:
Wheels run: https://github.com/siliconcompiler/siliconcompiler/actions/runs/7198084049

PR #2148, opened 2024-01-05
FPGA Flow Augmentation -- Flop techmapping

petergrossmann21:
In this PR, we upgrade the FPGA synthesis flow so that it can technology-map flip flops with asychronous sets/resets and synchronous enable signals.

This is done by porting the FPGA Architect reference flow for this scheme to Silicon Compiler.

Key implementation features are:

1.  A smarter dfflegalize pass, which is if/elsed with flags specifying whether an architecture supports asynchronous sets/resets and/or enables.  Any mix of the three flags is allowed, and the new legalize_flops function will figure out the correct legalization strategy based on these flags.
2. Additional hooks to propagate these flags from an FPGA part name definition within its FPGA family module to the corresponding tool/task settings for yosys
3. Adjustments to the yosys syn_fpga module to check for and use these flags to guide the syn_fpga.tcl script
4. A technology mapping step for flip flops that follows the dfflegalize pass
5. Updates to FPGA part name definitions to use the new flags and specify tech libraries
6. An example architecture (no proprietary features) that offers flip flops, multipliers, and memories
7. An example flop technology library (no proprietary features) that maps yosys primitives to generic flop cell names
8. A test case in test/flows that runs synthesis and place and route on the new example architecture with an up/down counter to map to async flops with enables.

Some things that may require adjustment to suit the preferred implementation style of Silicon Compiler:

* It may be preferable to specify the tool/task calls for yosys directly in the FPGA family modules, rather than with the indirection of a per-part name specification.  This would allow the lattice example to be oblivious to these flags since Lattice parts use a separate synthesis flow that is built into yosys, rather than the one used for the VPR example architectures
* Some cleanup may be needed for the case where a flop tech library is not required.  Looking for some leadership input on what expectations there are on this topic.
* Any feedback that helps set this PR up to provide a template for future extensions to DSP and memory mapping are appreciated; PRs for that are forthcoming.


examples/fpga_flow/designs/updowncount/updowncount.v
```diff
@@ -0,0 +1,29 @@
+//updowncount.v
+//Peter Grossmann
+//17 November 2023
+//$Id$
+//$Log$
+
```
gadfort, line 6: remove

examples/fpga_flow/techlib/tech_flops.v
```diff
@@ -0,0 +1,1267 @@
+//tech_flops.v
+//Peter Grossmann
+//19 May 2022
+//$Id$
+//$Log$
+
```
gadfort, line 6: remove

examples/fpga_flow/techlib/tech_flops.v
```diff
@@ -0,0 +1,1267 @@
...
+//This file provides a generic tech mapping option suitable for many
```
gadfort, line 7: might want to rename this file to make it clear it is an example techmap

siliconcompiler/fpgas/vpr_example.py
```diff
@@ -52,10 +53,32 @@ def setup(chip):
             fpga.set('fpga', 'example_arch_X005Y005', 'file', 'graphfile',
                      os.path.join(arch_root, 'example_arch_X005Y005_rr_graph.xml'))
             fpga.set('fpga', 'example_arch_X005Y005', 'var', 'channelwidth', '32')
+            fpga.set('fpga', part_name, 'var', 'flop_async_set', '0')
+            fpga.set('fpga', part_name, 'var', 'flop_async_reset', '0')
+            fpga.set('fpga', part_name, 'var', 'flop_enable', '0')
+            fpga.set('fpga', part_name, 'var', 'legalize_flops', '1')
+            fpga.add('fpga', part_name, 'file', 'yosys_flop_techmap', 'None')
```
gadfort, line 60: don't set the file to None, just don't set it.

siliconcompiler/fpgas/vpr_example.py
```diff
@@ -52,10 +53,32 @@ def setup(chip):
...
 
         if (part_name == 'example_arch_X008Y008'):
             # No RR graph for this architecture to support testing
             fpga.set('fpga', 'example_arch_X008Y008', 'var', 'channelwidth', '32')
+            fpga.set('fpga', part_name, 'var', 'flop_async_set', '0')
+            fpga.set('fpga', part_name, 'var', 'flop_async_reset', '0')
+            fpga.set('fpga', part_name, 'var', 'flop_enable', '0')
+            fpga.set('fpga', part_name, 'var', 'legalize_flops', '1')
+            fpga.add('fpga', part_name, 'file', 'yosys_flop_techmap', 'None')
```
gadfort, line 69: same

siliconcompiler/fpgas/vpr_example.py
```diff
@@ -52,10 +53,32 @@ def setup(chip):
...
+
+        if (part_name == 'example_arch_X014Y014'):
+
+            techlib_root = os.path.join(flow_root, 'techlib')
+
+            fpga.set('fpga', 'example_arch_X014Y014', 'var', 'channelwidth', '80')
```
gadfort, line 75: 'example_arch_X014Y014' -> part_name

siliconcompiler/fpgas/lattice_ice40.py
```diff
@@ -30,6 +30,10 @@ def setup(chip):
 
         fpga.set('fpga', part_name, 'vendor', vendor)
         fpga.set('fpga', part_name, 'lutsize', lut_size)
+        fpga.set('fpga', part_name, 'var', 'flop_async_set', 'None')
+        fpga.set('fpga', part_name, 'var', 'flop_async_reset', 'None')
+        fpga.set('fpga', part_name, 'var', 'flop_enable', 'None')
+        fpga.set('fpga', part_name, 'var', 'legalize_flops', 'None')
```
gadfort, line 36: I think this would be better as a list of ff features, instead of a bunch of TF flags.
The presence to this key would also indicate if the legalization would occur.

siliconcompiler/tools/yosys/syn_fpga.py
```diff
@@ -38,10 +38,75 @@ def setup_fpga(chip):
...
+             ",".join(['fpga', part_name, 'var', 'flop_async_reset']),
+             step=step, index=index)
+
+    chip.add('tool', tool, 'task', task, 'require',
+             ",".join(['fpga', part_name, 'var', 'flop_enable']),
+             step=step, index=index)
+
+    chip.add('tool', tool, 'task', task, 'require',
+             ",".join(['fpga', part_name, 'var', 'legalize_flops']),
+             step=step, index=index)
```
gadfort, line 55: no need to require the feature flags. But you will need to add help string to it so it's clear what these are in the documentation

siliconcompiler/tools/yosys/syn_fpga.py
```diff
@@ -38,10 +38,75 @@ def setup_fpga(chip):
...
+
+    # Do not require tech libraries, as there are some FPGAs that
+    # are so simple (in academia) that they do not require any
```
gadfort, line 58: remove

siliconcompiler/tools/yosys/syn_fpga.py
```diff
@@ -38,10 +38,75 @@ def setup_fpga(chip):
...
+
     chip.add('tool', tool, 'task', task, 'output', design + '_netlist.json', step=step, index=index)
     chip.add('tool', tool, 'task', task, 'output', design + '.blif', step=step, index=index)
 
 
+##################################################
```
gadfort, line 64: I don't think this preprocess adds any value to the readability of the tcl since it's just copying the values over.

siliconcompiler/tools/yosys/syn_fpga.tcl
```diff
@@ -1,10 +1,124 @@
+
+proc legalize_flops { async_reset async_set enable library } {
+
+    if { $library eq "None" } {
```
gadfort, line 4: what does library indicate?

siliconcompiler/tools/yosys/syn_fpga.tcl
```diff
@@ -1,10 +1,124 @@
...
+                    yosys log "Legalize DFFs for synchronous reset only"
+                    yosys dfflegalize \
+                        -cell \$_DFF_P_ 01 \
+                        -cell \$_DFF_P??_ 01 \
+
+                }
+            }
+        }
+    }
+}
```
gadfort, line 97: this can probably be simplified into a set of FFs based on the attributes and then just call the dfflegalize once with `{*}$args` type of thing. We can talk about this

siliconcompiler/tools/yosys/syn_fpga.tcl
```diff
@@ -1,10 +1,124 @@
...
+    [dict get $sc_cfg tool $sc_tool task $sc_task var flop_enable]
+set sc_syn_legalize_flops \
+    [dict get $sc_cfg tool $sc_tool task $sc_task var legalize_flops]
+
+if {[dict exists $sc_cfg tool $sc_tool task $sc_task file flop_techmap]} {
+    set sc_syn_flop_library \
+        [dict get $sc_cfg tool $sc_tool task $sc_task file flop_techmap]
+} else {
+    set sc_syn_flop_library "None"
+}
```
gadfort, line 121: remove

siliconcompiler/tools/yosys/syn_fpga.tcl
```diff
@@ -1,10 +1,124 @@
...
+set sc_syn_lut_size \
+    [dict get $sc_cfg fpga $sc_partname lutsize]
+set sc_syn_flop_async_set \
+    [dict get $sc_cfg tool $sc_tool task $sc_task var flop_async_set]
+set sc_syn_flop_async_reset \
+    [dict get $sc_cfg tool $sc_tool task $sc_task var flop_async_reset]
+set sc_syn_flop_enable \
+    [dict get $sc_cfg tool $sc_tool task $sc_task var flop_enable]
+set sc_syn_legalize_flops \
+    [dict get $sc_cfg tool $sc_tool task $sc_task var legalize_flops]
```
gadfort, line 114: I think a feature list would be easier to read and understand.

siliconcompiler/tools/yosys/syn_fpga.tcl
```diff
@@ -65,8 +185,14 @@ if {[string match {ice*} $sc_partname]} {
     yosys flatten
     yosys clean
 
+    if { $sc_syn_flop_library ne "None" } {
+        yosys techmap -map $sc_syn_flop_library
+    }
```
gadfort, line 190: need post processing call to techmap and opt to cleanup the design afterwards like the ASIC flow

siliconcompiler/tools/yosys/syn_fpga.tcl
```diff
@@ -54,7 +168,13 @@ if {[string match {ice*} $sc_partname]} {
...
     yosys techmap -map +/techmap.v
 
-    yosys dfflegalize -cell \$_DFF_P_ 01 -cell \$_DFF_P??_ 01
+    if { $sc_syn_legalize_flops != 0 } {
+        legalize_flops \
+            $sc_syn_flop_async_reset \
+            $sc_syn_flop_async_set \
+            $sc_syn_flop_enable \
+            $sc_syn_flop_library
+    }
```
gadfort, line 177: probably something like:
```
if { [llength $sc_ff_features] != 0 } {
...
}
```

examples/fpga_flow/techlib/tech_flops.v
gadfort: make it clear this is for the 014x014 case

gadfort (changes requested):

siliconcompiler/tools/yosys/syn_fpga.tcl
petergrossmann21: Trying to build the arguments out of a list produces this error:

`dfflegalize -cell {$_DFF_P_} 01 -cell {$_DFF_PN?_} 01 -cell {$_DFFE_PP_} 01 -cell {$_DFFE_PN?P_} 01 -cell {$_DFFSR_PNN_} 01 -cell {$_DFFSR_PNN_} 01`

There are some issues with needing to escape character the leading \ and $ confounding this; second opinions welcome.  Just eliminating log statements and changing the nested if/else to if ... elseif ... else helps readability to my eye.

petergrossmann21: Ah, perhaps I don't understand the significance of the {*}$args syntax.  Stay tuned

petergrossmann21: Yes; [this stackoverflow thread](https://stackoverflow.com/questions/37926222/what-is-the-significance-of-when-used-with-args-like-args-in-tcl) was helpful.

siliconcompiler/tools/yosys/syn_fpga.tcl
petergrossmann21: The purpose of having a boolean flag to allow/disallow flop legalization is to provide users a backdoor to skip this step for debugging purposes; in general, the step should always be executed in a final flow.

petergrossmann21:
@gadfort I believe this is now ready for a next review cycle, some tests are still running but based on local testing and tests that have completed passing looks highly likely.

petergrossmann21:
@gadfort On further review I still have an issue with integrating my flop techmap to the siliconcompiler_data object.  Is there any way to test this properly on a local machine (as opposed to the hack discussed last week)?

gadfort (changes requested):

gadfort (approved):

PR #2151, opened 2024-01-08
Correctly add `input` to schema in tool template

ubfx:
Clears up a small confusion in the template provided for new tools.


gadfort (approved):
Thanks for catching that and fixing it.

PR #2152, opened 2024-01-09
Fix dead link in README

ubfx:


gadfort (approved):
Thanks for fixing that.

PR #2155, opened 2024-01-17
Update conda environment, introduce conda CI test

ubfx:
Since conda is a very convenient way of setting up the environment for siliconcompiler, it's worth revisiting the environment setup.
This patch adds klayout to conda dependencies as it is now available on litex-hub. It also adds version constraints matching the current constraints in the tool modules.


ubfx:
At least the klayout version from litex-hub seems to be built without Qt bindings, which makes `LayoutView::get_image()` unavailable. This function returns a `QtImage` of the layout which can then be saved. There seems to be no other use of the QtBindings except for saving a screenshot of the layout, which can now be handled via `LayoutView::save_image()` instead.

siliconcompiler/tools/klayout/klayout_show.py
```diff
@@ -90,9 +90,8 @@ def __screenshot(schema, layout_view, output_path):
                                          'show_vertical_resolution',
                                          step=step, index=index)[0])
 
-    gds_img = layout_view.get_image(horizontal_resolution, vertical_resolution)
     print(f'[INFO] Saving screenshot to {output_path}')
-    gds_img.save(output_path, 'PNG')
+    layout_view.save_image(output_path, horizontal_resolution, vertical_resolution)
```
ubfx, line 94: At least the klayout version from litex-hub seems to be built without Qt bindings, which makes `LayoutView::get_image()` unavailable. This function returns a `QtImage` of the layout which can then be saved. There seems to be no other use of the Qt bindings except for saving a screenshot of the layout, which can now be handled via `LayoutView::save_image()` instead.

Documentation for the two functions:
https://www.klayout.de/doc-qt5/code/class_LayoutView.html#method100
https://www.klayout.de/doc-qt5/code/class_LayoutView.html#method203

gadfort: we have a few other tasks (screenshot) which would be impacted by this.

gadfort (changes requested):
thanks for updating this. I'm concerned we don't have the bandwidth to maintain the conda deployment, if we update this, I think it would need a CI test of some sort to ensure that it's kept reasonably functional. Additionally the environment.yml is not linked with the _tools.json.
I think the CI test and _tools.json<->environment.yml would need to be solved to ensure there is continued value to the conda deployment.

siliconcompiler/tools/klayout/klayout_show.py
ubfx: They should all work with these changes because no task actually uses the Qt object for anything other than saving to a file, which can be done through `save_image()` instead without the need for going through Qt on the Python side.
I'll branch this change off to its own PR and test it some more. Then this PR can focus on the conda/CI changes.

siliconcompiler/tools/openroad/openroad.py
```diff
@@ -34,7 +34,8 @@ def setup_tool(chip, exit=True, clobber=True):
 
     chip.set('tool', tool, 'exe', tool)
     chip.set('tool', tool, 'vswitch', '-version')
-    chip.set('tool', tool, 'version', '>=v2.0-10708', clobber=clobber)
+    # Temporarily disable OpenROAD version check
+    # chip.set('tool', tool, 'version', '>=v2.0-10708', clobber=clobber)
```
ubfx, line 38: This was necessary to get sc working with OpenROAD from conda, because those binaries only have the git hash as the version string (although the commit is from version 2.x). I'll see if I can find the reason for that over at OpenROAD.

ubfx:
> a CI test of some sort to ensure that it's kept reasonably functional

I added a CI test which triggers on pushes / PRs to the `setup/` folder and runs the `tests/tools/test_{openroad, yosys, surelog, klayout}.py` tests with conda tools.

> _tools.json<->environment.yml

How would you implement this? The binary packages available on conda might not always be up to date with the github releases of the tools.

siliconcompiler/tools/openroad/openroad.py
gadfort: I think this is an issue with the openroad compile for conda (if I recall, they are using an old version of git which doesn't support the git describe call used). I think the solution would be to correct it with conda (whomever is maintaining the openroad compile).

.github/workflows/conda_ci.yml
```diff
@@ -0,0 +1,52 @@
...
+    paths: 'setup/**'
+    
+jobs:
+  test-basic-tools:
+    name: Test basic tools
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.11"] # ["3.8", "3.9", "3.10", "3.11", "3.12"]
```
gadfort, line 17: any reason the other versions are not included?

.github/workflows/conda_ci.yml
```diff
@@ -0,0 +1,52 @@
...
+          
+          ${{ steps.setup-python.outputs.python-path }} -m venv clean_env
+          source clean_env/bin/activate
+          python3 --version
+          
+          cd $GITHUB_WORKSPACE
+          python3 -m pip install --upgrade pip
+          python3 -m pip install .[test]
+          pytest tests/tools/test_openroad.py tests/tools/test_klayout.py tests/tools/test_surelog.py tests/tools/test_yosys.py
+
```
gadfort, line 51: the tests that most frequently break are the show/screenshot tests in klayout so they should be included as well.

setup/conda/environment.yml
```diff
@@ -7,8 +7,10 @@ dependencies:
...
-  - openroad
-  - magic
+  - surelog>=1.51
+  - openroad>=2.0-10708
+  - magic>=8.3.196
   - netgen
-  - yosys
+  - yosys>=0.33+21
+  - klayout>=0.28.0
+  - fmt
```
gadfort, line 16: why fmt?

.github/workflows/conda_ci.yml
gadfort: ideally this should be able run all tests with these/supported tools

setup/conda/environment.yml
ubfx: It seems like on ubuntu (including the github runners), OpenROAD is missing libfmt at runtime. I'll try to look into this as well on the conda recipe side but currently we need this to run the tests.

https://github.com/hdl/conda-eda/issues/360

.github/workflows/conda_ci.yml
ubfx: Just for quicker testing, I enabled them back.

.github/workflows/conda_ci.yml
ubfx: I added the show test and the heartbeat test (which should cover all of the basic tools). However, we can't run all tests on this basic conda setup, because we don't include the simulators, magic, nextpnr, etc in the environment.

siliconcompiler/tools/openroad/openroad.py
ubfx: I'm trying to get the OpenROAD build fixed over at hdl/conda-eda, then this should hopefully be taken care of (along with the fmt issue below)

ubfx:
@gadfort We resolved the main roadblock for this, OpenROAD on LiteX-Hub being old and not having the correct `git describe` version string (https://github.com/hdl/conda-eda/pull/366 https://github.com/hdl/conda-eda/pull/374). The fmt requirement is still there, but I have a PR open for this (https://github.com/hdl/conda-eda/pull/375) and it seems like less of a problem to temporarily include fmt in our conda setup.

The testing setup now works out of the box on the github runners: https://github.com/ubfx/siliconcompiler-1/actions/runs/8124560562

.github/workflows/conda_ci.yml
```diff
@@ -0,0 +1,60 @@
...
+          conda activate sc_env
+          
+          ${{ steps.setup-python.outputs.python-path }} -m venv clean_env
+          source clean_env/bin/activate
+          python3 --version
+          
+          cd $GITHUB_WORKSPACE
+          python3 -m pip install --upgrade pip
+          python3 -m pip install .[test]
+          pytest -n auto -k "not sim" tests/tools/test_openroad.py tests/tools/test_klayout.py tests/tools/test_surelog.py tests/tools/test_yosys.py tests/core/test_tool_tasks.py tests/flows/test_show.py tests/apps/test_sc_show.py tests/examples/test_heartbeat.py
```
gadfort, line 60: sim is not a pytest marker we use, what is it for?

gadfort (approved):
Looks good. Just need to address the "not sim" thing in the CI test.
Once / if the fmt and ruby get correctly added to the conda setups, we should make sure they get removed from here.
Thanks for sorting that out.

.github/workflows/conda_ci.yml
ubfx: The `test_heartbeat.v` contains the "test_sim" test which can't be run because we haven't added icarus to the conda setup. The `-k "not sim"` should filter that test out by name

gadfort: ahh, I see I missed the `-k` vs `-m`.

PR #2156, opened 2024-01-17
Fix "benchmark" example

ubfx:
I wasn't able to run the "benchmark" example, it seems like some of the schema keypaths are from an older version of siliconcompiler.

### DIFF
--- a/examples/benchmark/benchmark.py
+++ b/examples/benchmark/benchmark.py
@@ -2,57 +2,53 @@ import os
 import time
 import siliconcompiler
 
+
 # Setting up the experiment
 def main():
     rootdir = os.path.dirname(__file__)
-    parent = os.path.abspath(rootdir + '/..')
+    parent = os.path.abspath(rootdir + "/..")
 
-    examples = [{'heartbeat': 1000},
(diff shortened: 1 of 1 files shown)


examples/benchmark/benchmark.py
```diff
@@ -1,52 +1,47 @@
...
-from siliconcompiler.flows import asicflow
 
 # Setting up the experiment
-
-
 def main():
     rootdir = os.path.dirname(__file__)
 
     examples = [{'heartbeat': 1000},
-                {'picorv32': 1000},
```
ubfx, line 13: The `picorv32.v` is not delivered with siliconcompiler, maybe it's better to leave it out so that the example runs straight after cloning.

gadfort: I'll take a look at this, there is a method to get the picorv32 source files via the package source handling we introduced a few months ago.

ubfx: I added the package mechanism as it used in `examples/picorv32`

gadfort (approved):
Thanks

PR #2158, opened 2024-01-18
Remove dependency on klayout Qt bindings for saving screenshots

ubfx:
This allows using klayout built without Qt bindings with siliconcompiler. The Qt bindings were only used to grab an image of a `LayoutView` and then save it to a .png. This can be achieved by calling `save_image()` directly which combines those two steps in a single call without the need of handling an intermediate Qt object.

Tested this with the normal klayout export task as well as the gcd_screenshot example.

### DIFF
--- a/siliconcompiler/tools/klayout/show.py
+++ b/siliconcompiler/tools/klayout/show.py
@@ -88,7 +88,8 @@ def pre_process(chip):
     index = chip.get('arg', 'index')
     tool, task = chip._get_tool_task(step, index)
 
-    if chip.valid('tool', tool, 'task', task, 'var', 'show_filepath'):
+    if chip.valid('tool', tool, 'task', task, 'var', 'show_filepath') and \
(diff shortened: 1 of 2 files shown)


siliconcompiler/tools/klayout/show.py
```diff
@@ -88,7 +88,8 @@ def pre_process(chip):
     index = chip.get('arg', 'index')
     tool, task = chip._get_tool_task(step, index)
 
-    if chip.valid('tool', tool, 'task', task, 'var', 'show_filepath'):
+    if chip.valid('tool', tool, 'task', task, 'var', 'show_filepath') and \
+       chip.get('tool', tool, 'task', task, 'var', 'show_filepath', step=step, index=index):
```
ubfx, line 92: Needed to add this line because the `gcd_scerenshot` example would crash otherwise, because `show_filepath` is empty.

gadfort: thanks.

gadfort (approved):

PR #2171, opened 2024-01-28
Removing pip installs from surelog setup

aolofsson:
- pip cmake package install fails, not needed for machine install
- mutidict for venv seems misplaced, looking at commit log it was to fix something in venv, which doesn't belong here?

Install now works correctly on a clean ubuntu 22.04 machine


gadfort:
@aolofsson cmake is needed since the Ubuntu 20.04 version of cmake is too old and this provides an easy way to get that a working version. `orderedmultidict` mirrors the surelog requirements (likely once we move to Ubuntu 22.04 for the tools build we can probably remove these two).

gadfort:
These changes have been incorporated into #2104

PR #2184, opened 2024-02-07
Grossmann/vpr constraints

petergrossmann21:
In this PR we add two capabilities to the VPR placement step in the FPGA flow:

1.  you can now specify an XML placement constraints file via `chip.set('input', 'constraint', 'pins', <filename>)`
2. You can now also specify FPGA placement constraints to VPR through the schema using `chip.set('constraint', 'component', '<port_name>', 'placement', <x,y,block_num> )`  The place.py module in the VPR tool driver now will aggregate these constraints and auto-generate an XML placement constraint file, which VPR will pick up if the user does not take advantage of #1.

To support #2, a support file is added to the VPR tool driver for handling the XML generation.

A couple of opens that might need resolution as part of this PR:

1. The method by which the auto-generated file is detected may not be correct for SC best practices; suggestions welcome for a preferred approach.
2. It's still not clear to me whether I should have 'pins' or 'placement' as the keyword for the constraints file for item #1 above, as in practice use this just for pins, but VPR allows you to put any placeable block in the contraints file
3. VPR allows specification of a bounding box of locations for block placement, rather than just a single point location.  Currently this is not supported, as it's not applicable to constraining pins.  To support it correctly, the existing tuple that is supplied for component constraints would no longer be adequate; there would need to be support to specify a second pair of (X,Y) components so that the extent of the bounding box can be specified.


siliconcompiler/tools/vpr/place.py
```diff
@@ -24,6 +26,39 @@ def setup(chip, clobber=True):
...
+
+def pre_process(chip):
+    ''' Tool specific function to run before step execution
+    '''
+
+    design = chip.top()
+    step = chip.get('arg', 'step')
+    index = chip.get('arg', 'index')
+
+    if 'pins' in chip.getkeys('input', 'constraint'):
```
gadfort, line 42: use `.valid('input', 'constraint', 'pins')`

siliconcompiler/tools/vpr/place.py
```diff
@@ -24,6 +26,39 @@ def setup(chip, clobber=True):
...
+        chip.logger.info("Using pin constraints file instead of pin constraints from chip.set")
+    else:
+        all_component_constraints = chip.getkeys('constraint', 'component')
+        all_place_constraints = {}
+        if (len(all_component_constraints) > 0):
```
gadfort, line 47: this check isn't needed, if all_component_constraint is empty then the for loop will no nothing and you can just check the dict before calling generate_vpr_constraints_xml

siliconcompiler/tools/vpr/place.py
```diff
@@ -24,6 +26,39 @@ def setup(chip, clobber=True):
...
+            for component in all_component_constraints:
+                place_constraint = chip.get('constraint', 'component', component, 'placement',
+                                            step=step, index=index)
+                chip.logger.info(f'Place constraint for {component} at {place_constraint}')
+                all_place_constraints[component] = place_constraint
+
+            constraints_xml = generate_vpr_constraints_xml(all_place_constraints)
+            xml_file = f'inputs/{design}_constraints.xml'
```
gadfort, line 55: use yosys convention for file naming: sc_constraints.xml (the design name is not needed).
also, use the set the path in the schema like: https://github.com/siliconcompiler/siliconcompiler/blob/7516ab2a6108b788aac41a0aa50423132058e4ea/siliconcompiler/tools/yosys/syn_asic.py#L500

siliconcompiler/tools/vpr/vpr.py
```diff
@@ -86,27 +89,37 @@ def runtime_options(chip, tool='vpr'):
         chip.error(f"Specified task {task} doesn't map to a VPR operation", fatal=True)
 
     if 'sdc' in chip.getkeys('input', 'constraint'):
-        sdc_file = f"--sdc_file {chip.get('input', 'constraint', 'sdc', step=step, index=index)}"
-        options.append(sdc_file)
+        sdc_file = find_single_file(chip, 'input', 'constraint', 'sdc',
+                                    file_not_found_msg="SDC file not found")
```
gadfort, line 93: why this helper function? find_files will error out if the files are missing and you should ensure the input/constraint/sdc is in the require list if its set so the preprocess will check if it's there.

siliconcompiler/tools/vpr/vpr.py
```diff
@@ -86,27 +89,37 @@ def runtime_options(chip, tool='vpr'):
...
-        rr_graphs = chip.find_files('fpga', part_name, 'file', 'graphfile')
-
-    else:
-        rr_graphs = []
+    # Give input constraint file priority over individually-specified
+    # constraints that our place step pre-processing may have generated
+    # from chip.get('constraint', 'placement', ...) settings:
+    if 'pins' in chip.getkeys('input', 'constraint'):
+        pin_constraint_file = find_single_file(chip, 'input', 'constraint', 'pins',
+                                               file_not_found_msg="VPR constraints file not found")
```
gadfort, line 105: same

siliconcompiler/tools/vpr/vpr.py
```diff
@@ -47,6 +49,7 @@ def runtime_options(chip, tool='vpr'):
 
     topmodule = chip.top()
     blif = f"inputs/{topmodule}.blif"
+    auto_constraints = f'inputs/{topmodule}_constraints.xml'
```
gadfort, line 52: get from schema

siliconcompiler/tools/vpr/vpr.py
```diff
@@ -127,6 +140,28 @@ def runtime_options(chip, tool='vpr'):
...
+################################
+# Wrapper around find files to
+# help with error checking that
+# only a single file is found
+################################
+
+def find_single_file(chip, *keypath, file_not_found_msg="File not found"):
+
+    if chip.valid(*keypath) and chip.get(*keypath):
+        file_list = chip.find_files(*keypath)
```
gadfort, line 152: need call find_files with step/index.

siliconcompiler/tools/vpr/xml_constraints_gen.py
```diff
@@ -0,0 +1,107 @@
...
+    subtile = int(pin_region[2])
+    return x_low, x_high, y_low, y_high, subtile
+
+
+def generate_partition_name(pin):
+
+    partition_name = pin
+    partition_name = partition_name.replace('[', '_')
+    partition_name = partition_name.replace(']', '')
+    partition_name = "part_" + partition_name
```
gadfort, line 70: where does this come from?

siliconcompiler/tools/vpr/xml_constraints_gen.py
```diff
@@ -0,0 +1,107 @@
...
+def write_vpr_constraints_xml_file(constraints: ET.Element, filename: str):
+
+    m_encoding = 'UTF-8'
+
+    dom = xml.dom.minidom.parseString(ET.tostring(constraints))
+    xml_string = dom.toprettyxml()
+    part1, part2 = xml_string.split('?>')
+
+    with open(filename, 'w') as xfile:
+        xfile.write(part1 + 'encoding=\"{}\"?>\n'.format(m_encoding) + part2)
```
gadfort, line 106: why is this needed? can't you just set the encoding tag attribute in the root element?

siliconcompiler/tools/vpr/xml_constraints_gen.py
```diff
@@ -0,0 +1,107 @@
...
+        xfile.close()
```
gadfort, line 107: using `with` you do not need to `.close()`

siliconcompiler/tools/vpr/xml_constraints_gen.py
```diff
@@ -0,0 +1,107 @@
...
+
+def generate_partition_list_xml(pin_map):
+
+    partition_list = ET.Element("partition_list")
+
+    # ***ASSUMPTION:  pin map is a dictionary of block names
+    #                 and tuples, where the tuples specify the
+    #                 (X,Y,subtile) locations that each block
+    #                 is constrained to
+    for pin in pin_map:
```
gadfort, line 25: use `for pin, region in pin_map.items()`

siliconcompiler/tools/vpr/xml_constraints_gen.py
```diff
@@ -0,0 +1,107 @@
+import xml.etree.ElementTree as ET
```
gadfort, line 1: maybe rename this file as _xml_constraint.py to help indicate it's not a task but a utility file

gadfort (changes requested):

siliconcompiler/tools/vpr/vpr.py
petergrossmann21: The main purpose of the helper is to hide/reuse the ugliness of checking that exactly one file has been passed to these arguments instead of a whole list of files, since VPR only allows one file to be passed to each of these arguments.

siliconcompiler/tools/vpr/xml_constraints_gen.py
petergrossmann21: Just a personal convention; could have been anything.

siliconcompiler/tools/vpr/xml_constraints_gen.py
petergrossmann21: This is an old formula I found somewhere online back in summer 2022 -- wasn't smart enough to write down where it came from.  After messing around a bit, here's what I found:
1.  ET doesn't actually use <xml ...> as its root element, the dom wraps this around the ET top level.
2. VPR doesn't seem to care if the encoding is specified, and will happily ingest the file if I simplify the xfile.write() call and omit the encoding spec altogether.

Simplifying the code based on these observations yields:

  ```
    dom = xml.dom.minidom.parseString(ET.tostring(constraints))
    xml_string = dom.toprettyxml()

    with open(filename, 'w') as xfile:
        xfile.write(str(xml_string))
```

petergrossmann21:
@gadfort Only outstanding items left now are the find single file helper and the partition_name setup; see comments for each.

siliconcompiler/tools/vpr/vpr.py
gadfort: @petergrossmann21 this still applies

gadfort (changes requested):
Please fix the call to find files

gadfort (approved):

PR #2196, opened 2024-02-14
Update tclint and fix expression issues

nmoroze:
I just released a new version of `tclint` that parses expressions. It caught a few small things, one bug and a couple other things that are more of a style issue. 

Expression style checking is coming next, but wanted to open this now to get the bug fixed!

### DIFF
--- a/siliconcompiler/tools/openroad/scripts/sc_screenshot.tcl
+++ b/siliconcompiler/tools/openroad/scripts/sc_screenshot.tcl
@@ -14,7 +14,7 @@ sc_save_image "screenshot" "outputs/${sc_design}.png" $sc_resolution
 
 gui::restore_display_controls
 
-if { [dict exist $sc_cfg tool $sc_tool task $sc_task {var} include_report_images] &&
+if { [dict exists $sc_cfg tool $sc_tool task $sc_task {var} include_report_images] &&
      [lindex [dict get $sc_cfg tool $sc_tool task $sc_task {var} include_report_images] 0]
      == "true"} {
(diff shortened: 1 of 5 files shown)


siliconcompiler/tools/openroad/scripts/sc_screenshot.tcl
```diff
@@ -14,7 +14,7 @@ sc_save_image "screenshot" "outputs/${sc_design}.png" $sc_resolution
 
 gui::restore_display_controls
 
-if { [dict exist $sc_cfg tool $sc_tool task $sc_task {var} include_report_images] &&
+if { [dict exists $sc_cfg tool $sc_tool task $sc_task {var} include_report_images] &&
```
nmoroze, line 17: Turns out this was okay since `dict` will allow partial subcommands, but `dict exists` is technically the full command.

siliconcompiler/tools/yosys/sc_syn.tcl
```diff
@@ -61,7 +61,7 @@ if { [file exists "inputs/$sc_design.v"] } {
 yosys chparam -list
 if {[dict exists $sc_cfg option param]} {
     dict for {key value} [dict get $sc_cfg option param] {
-        if !{[string is integer $value]} {
+        if {![string is integer $value]} {
```
nmoroze, line 64: This also works as intended in this case, but in the general case I think `!{[cmd...]}` can behave funny

gadfort (approved):
Awesome, thanks.

PR #2224, opened 2024-03-08
Import data sources of an imported module

ubfx:
I noticed that since the recent changes to the data source importing, the `tests/test_paths.py` in lambdapdk fails.

I think this might be the reason: We're trying to get the 'package' information from the subgroup ('pdk', 'fpga', ...) instead of at the top level of the schema.


gadfort (approved):
Thanks, nice catch.

gadfort:
@ubfx it appears there were no CI testing for this so I added a test in #2225

PR #2258, opened 2024-03-22
Fix Conda CI Tests and environment

ubfx:
This fixes the Conda CI Test by forcing the environment setup to use python>=3.8, which apparently also solves the dependency problem with the old ruby version.


gadfort (approved):
Thanks for fixing this

gadfort:
https://github.com/siliconcompiler/siliconcompiler/actions/runs/8390444866/job/22978676757

PR #2264, opened 2024-03-24
Gracefully handle encoding errors in logfile when in quiet mode

ubfx:
The error handlers for invalid encoding were missing here which can lead to a crash when in quiet mode and log contains invalid characters.


gadfort:
Is there a more "universal" fix to this? it seems like all the `open()` are at risk including in all the tool drivers

ubfx:
I didn't find a way to set the default handler. I think the only other way would be trying to override the `strict` handler although that also seems unclean.

gadfort:
#2267 see if something like this works, it might be useful to have a testcase of the issue in the CI if you can provide one

gadfort:
Replaced with #2267 2267

PR #2275, opened 2024-03-25
Grossmann/fpga flow systemverilog test

petergrossmann21:
This adds a highly minimal check to the sv2v front end for the FPGA flow.  Additional work could be done in the future to expand the range of SystemVerilog syntax that we wish to have test coverage for.


gadfort:
Already getting exercised in ASIC flow

PR #2277, opened 2024-03-25
Add test for error handlers in codecs

ubfx:
This test runs a step producing output/logfile with invalid UTF-8. This tests whether the codecs error handling works correctly in a spawned subprocess.


gadfort (changes requested):
Looks good, just need to fix the linting failures.

gadfort (approved):
Lgtm

PR #2281, opened 2024-03-27
Add hook to do clock routing for non-ideal clock architectures

petergrossmann21:


siliconcompiler/tools/vpr/vpr.py
```diff
@@ -119,6 +119,16 @@ def runtime_options(chip, tool='vpr'):
     options.append('--sweep_constant_primary_outputs off')
     options.append('--sweep_dangling_blocks off')
 
+    # Academic FPGA architectures (such as those included as examples
+    # in Silicon Compiler) model clocks as ideal nets for simplicity;
+    # however, most FPGA chips require clocks to be routed using
+    # VPR's clock routing algorithm; enable that algorithm unless the
+    # part specifies an "ideal_clock" variable
+    if (chip.valid('fpga', part_name, 'var', 'ideal_clock')):
```
gadfort, line 127: just checking if valid, just indicates it' been set, but not what the value is, you need to affirm the value was 'true' or 'false'

petergrossmann21: Now checks if 'true' and enables clocking_method route otherwise.

gadfort (approved):

PR #2295, opened 2024-03-30
Grossmann/constraints gen

petergrossmann21:
Here is the first cut at constraints generation.  The format for mapping port names to placement locations is kept as minimalist and simple as possible.  An example is provided for the example_arch_X014Y014 architecture, but is not yet documented.

A couple of testing-related considerations:

* not implemented, but feasible if we wish, is a validation step where we read in the placement file and compare it against constraints to verify that the constraints were honored.  Haven't done this yet as a bit of elbow grease is required to reduce to practice.
* This PR also catches a minor testing gap in verifying that users can specify the XML file themselves.  This test requires committing an example XML file to the repo (or replacing the XML I've committed with a generator for testing).  The presence of the XML also offers us the option to use it as a golden file for the generator.


siliconcompiler/tools/vpr/place.py
```diff
@@ -38,7 +41,25 @@ def pre_process(chip):
     step = chip.get('arg', 'step')
     index = chip.get('arg', 'index')
 
-    if not chip.valid('input', 'constraint', 'pins', default_valid=True):
+    part_name = chip.get('fpga', 'partname')
+
+    if 'pinmap' in chip.getkeys('input', 'constraint'):
```
gadfort, line 46: `if chip.valid('input', 'constraint', 'pinmap') and chip.get('input', 'constraint', 'pinmap', step=step, index=index):`

siliconcompiler/tools/vpr/place.py
```diff
@@ -38,7 +41,25 @@ def pre_process(chip):
...
+                                        file_not_found_msg="constraints map not found")
+
+        constraints_map = load_constraints_map(map_file)
+        json_constraints = load_json_constraints(constraint_file)
+        all_place_constraints = map_constraints(json_constraints, constraints_map)
+
+        constraints_xml = generate_vpr_constraints_xml(all_place_constraints)
+        write_vpr_constraints_xml_file(constraints_xml, vpr.auto_constraints())
+
+    elif not chip.valid('input', 'constraint', 'pins', default_valid=True):
```
gadfort, line 62: why default_valid?

tests/flows/test_fpga_xml_constraints.py
```diff
@@ -0,0 +1,49 @@
...
+    # 2. Define source files
+    v_src = os.path.join(flow_root, 'designs', benchmark_name, f'{benchmark_name}.v')
+    chip.input(v_src)
+
+    # 3. Set placement constraints
+    xml_constraints = os.path.join(flow_root,
+                                   'designs',
+                                   benchmark_name,
+                                   'constraints',
+                                   f'pin_constraints_{arch_name}.xml')
```
gadfort, line 33: since this xml file is only used in the tests, please move it to the tests folder `/tests/flows/data/`

siliconcompiler/tools/vpr/_json_constraint.py
```diff
@@ -0,0 +1,98 @@
...
+    return constraints_map
+
+
+def map_constraints(json_generic_constraints,
+                    constraints_map):
+
+    design_constraints = {}
+
+    # If no constraints map is provided pass the constraints directly
+    if (len(constraints_map) == 0):
```
gadfort, line 62: `if constraints_map:`

siliconcompiler/tools/vpr/_json_constraint.py
```diff
@@ -0,0 +1,98 @@
...
+
+            if (design_pin_assignment in constraints_map):
+                # Convert the dictionary entries in the constraints map into a tuple:
+                design_pin_constraint_assignment = (
+                    constraints_map[design_pin_assignment]['x'],
+                    constraints_map[design_pin_assignment]['y'],
+                    constraints_map[design_pin_assignment]['subtile'])
+
+            else:
+                design_pin_constraint_assignment = (0, 0, 0)
```
gadfort, line 90: what does this default mean?

siliconcompiler/tools/vpr/_json_constraint.py
```diff
@@ -0,0 +1,98 @@
...
+    if (constraints_map_file):
+        constraints_map = load_constraints_map(constraints_map_file)
+    else:
+        constraints_map = {}
+
+    mapped_constraints = map_constraints(json_generic_constraints,
+                                         constraints_map)
+
+    constraints_xml = generate_vpr_constraints_xml(mapped_constraints)
+    write_vpr_constraints_xml_file(constraints_xml, constraints_file_out)
```
gadfort, line 35: probably not needed since this file is not meant to be run outside of the pre_processing step

siliconcompiler/tools/vpr/_json_constraint.py
```diff
@@ -0,0 +1,98 @@
...
+def load_constraints_map(constraints_map_file):
+
+    constraints_map = {}
+    with (open(constraints_map_file, "r")) as constraints_map_data:
+        constraints_map = json.loads(constraints_map_data.read())
+
+    return constraints_map
+
+
+def map_constraints(json_generic_constraints,
```
gadfort, line 56: is there any error checking available? ie. if I set a pin to a non-existent location?

gadfort (changes requested):

siliconcompiler/tools/vpr/place.py
petergrossmann21: I tried the proposed change prior to what was implemented and it didn't work; the choice on line 46 matches methodology we greenlit for passing XML pin constraints [here](https://github.com/siliconcompiler/siliconcompiler/blob/2560559abded23d75a5e3f38be395ee993cf9ef9/siliconcompiler/tools/vpr/vpr.py#L155).

I don't have strong opinions about which is "right", but I would prefer to be consistent and to understand what is preventing your proposed change from producing the desired behavior.

siliconcompiler/tools/vpr/place.py
petergrossmann21: Unclear at the moment; everything on that line except 'el' is from a previous PR.  Reviewing SC docs makes my choice to use that parameter setting suspect.  Will try deleting and retesting and revisit only if tests fail.

siliconcompiler/tools/vpr/_json_constraint.py
petergrossmann21: It means that the pin constraint will be set to block (0,0) element #0.  In (nearly) all VPR based architectures, this location does not exist, so making this setting should result in placement failure.  That said, this is not a completely safe default for trying to trigger things to break, nor is triggering placement failure in this way necessarily the best error handling strategy.  Am open to other strategies for how to handle this case and better trap errors for the user.

siliconcompiler/tools/vpr/_json_constraint.py
petergrossmann21: I used it for testing purposes during development; would only recommend keeping if we feel that such a test mode is useful in some future enterprise.

siliconcompiler/tools/vpr/_json_constraint.py
petergrossmann21: For checking that locations are valid, pretty robust error checking should be possible if we add a VPR XML file parser.  Need to decide a) if you want to sign up for supporting said parser; b) on what timetable to do the development.  More rudimentary error checking is possible if FPGA array size is supplied, as you can do out of bounds checking.  If we check for in-bounds errors in assignment without XML support the checking is likely to be quite brittle.  

Also possible to add error checking for pin name correctness by adding a BLIF parser; then you could report whether the JSON file has a bad pin name.  Similarly, and without any extra parsers, we can check that the user has supplied port names that are in the constraints map JSON file for mapping to the FPGA architecture.

tests/flows/test_fpga_xml_constraints.py
petergrossmann21: Done

siliconcompiler/tools/vpr/place.py
petergrossmann21: Removing default_valid=True causes the following failures; all tests pass otherwise.  Based on which pass (the three constraints tests), I believe we added default_valid=True to allow the flow to keep working when no constraints are given?

```
FAILED test_dspblackbox_fpgaflow.py::test_dspblackbox_fpgaflow - siliconcompiler._common.SiliconCompilerError: These final steps could not be reached: ['route']
FAILED test_dspextract_fpgaflow.py::test_dspextract_fpgaflow - siliconcompiler._common.SiliconCompilerError: These final steps could not be reached: ['route']
FAILED test_dspmap_fpgaflow.py::test_dspmap_fpgaflow - siliconcompiler._common.SiliconCompilerError: These final steps could not be reached: ['route']
FAILED test_flopmap_fpgaflow.py::test_flopmap_fpgaflow - siliconcompiler._common.SiliconCompilerError: These final steps could not be reached: ['route']
FAILED test_fpgaflow_apr.py::test_fpgaflow_apr - siliconcompiler._common.SiliconCompilerError: These final steps could not be reached: ['route']
FAILED test_fpgaflow.py::test_fpgaflow - siliconcompiler._common.SiliconCompilerError: These final steps could not be reached: ['bitstream']
FAILED test_matrix_multiply_fpgaflow.py::test_matrix_multiply_fpgaflow - siliconcompiler._common.SiliconCompilerError: These final steps could not be reached: ['route']
FAILED test_mem_to_flops_fpgaflow.py::test_mem_to_flops_fpgaflow - siliconcompiler._common.SiliconCompilerError: These final steps could not be reached: ['route']
FAILED test_vhdl_fpgaflow.py::test_fpgaflow - siliconcompiler._common.SiliconCompilerError: These final steps could not be reached: ['bitstream']
```

siliconcompiler/tools/vpr/_json_constraint.py
petergrossmann21: Done.

siliconcompiler/tools/vpr/_json_constraint.py
petergrossmann21: Note that this is exactly the error checking I describe as easy to implement below; submitting a side PR with a proposal for how to convert this from questionable default behavior into SC-driven error checking.

siliconcompiler/tools/vpr/_json_constraint.py
gadfort: just move this code down to the `if __name__...` section and remove the `main()` then

siliconcompiler/tools/vpr/place.py
petergrossmann21: I understand this issue now; the default_valid=True was indeed masking a bug in the setup.  Adding a hook in place.py's `pre_process()` function to block constraints file generation when no constraints are provided resolves this.

siliconcompiler/tools/vpr/place.py
```diff
@@ -38,7 +40,38 @@ def pre_process(chip):
...
+
+    # If the user explicitly provides an XML constraints file, give that
+    # priority over other constraints input types:
+    if 'pins' in chip.getkeys('input', 'constraint'):
+        constraint_file = vpr.find_single_file(chip, 'input', 'constraint', 'pins',
+                                               step=step, index=index,
+                                               file_not_found_msg="VPR constraints file not found")
+
+        if (constraint_file is not None):
+            shutil.copy(constraint_file, vpr.auto_constraints())
```
gadfort, line 53: I'm going to approve, but this is bad practice. If the file doesn't need to be moved around, we should always refer to the source material.

PR #2299, opened 2024-04-01
Rudimentary error checking for pin constraints

petergrossmann21:


siliconcompiler/tools/vpr/_json_constraint.py
```diff
@@ -87,11 +88,11 @@ def map_constraints(json_generic_constraints,
                     constraints_map[design_pin_assignment]['subtile'])
 
             else:
-                design_pin_constraint_assignment = (0, 0, 0)
+                errors += 1
```
gadfort, line 91: maybe you can add a call to `chip.logger.error('')` with a message about which pin has issues

gadfort (changes requested):

siliconcompiler/tools/vpr/_json_constraint.py
petergrossmann21: Done.

gadfort (approved):

PR #2304, opened 2024-04-01
Add save graphics commands to VPR place and route

azaidy:


siliconcompiler/tools/vpr/vpr.py
```diff
@@ -194,6 +194,25 @@ def runtime_options(chip, tool='vpr'):
     return options
 
 
+############################
+# Get common graphics files
+############################
+
+def get_common_graphics(chip, graphics_commands=[]):
```
gadfort, line 201: no need to pass in the graphics_commands list, you can just extend the list on the return (also this doens't work correctly)

siliconcompiler/tools/vpr/route.py
```diff
@@ -24,6 +24,55 @@ def setup(chip, clobber=True):
...
+                             "set_routing_util 3; " +
+                             f"save_graphics reports/{design}_route_util_with_formula.png;")
+    graphics_commands.append("set_draw_block_text 0; " +
+                             "set_draw_block_outlines 0; " +
+                             "set_routing_util 4; " +
+                             f"save_graphics reports/{design}_route_util_over_blocks.png;")
+
+    graphics_command_str = ""
+    for command in graphics_commands:
+        graphics_command_str = graphics_command_str + " " + command
```
gadfort, line 67: why not do `graphics_command_str = " ".join(graphics_commands)`

gadfort: if they are all list appends you could just do `"; ".join(graphics_commands)`

siliconcompiler/tools/vpr/vpr.py
azaidy: Fixed.

siliconcompiler/tools/vpr/route.py
azaidy: Fixed.

gadfort (approved):

gadfort (approved):

PR #2335, opened 2024-04-12
Yosys: Introduce parameter allowing to mask out RTL modules

ubfx:
I think this is quite a useful option to have when running only a yosys frontend flow through siliconcompiler. It allows masking out RTL modules, which would later be replaced my macro cells without having to modify the RTL.

### DIFF
--- a/siliconcompiler/tools/yosys/syn_asic.tcl
+++ b/siliconcompiler/tools/yosys/syn_asic.tcl
@@ -163,6 +163,15 @@ foreach bb_file $sc_blackboxes {
 # Synthesis
 ########################################################
 
+# Before working on the design, we mask out any module supplied via 
+# `blackbox_modules`. This allows synthesis of parts of the design without having
+# to modify the input RTL.
+if { [dict exists $sc_cfg tool $sc_tool task $sc_task var blackbox_modules] } {
+    foreach bb [dict get $sc_cfg tool $sc_tool task $sc_task var blackbox_modules] {
+        yosys blackbox $bb
+    }
+}
+
 # Although the `synth` command also runs `hierarchy`, we run it here without the
 # `-check` flag first in order to resolve parameters before looking for missing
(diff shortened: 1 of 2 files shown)


siliconcompiler/tools/yosys/syn_asic.tcl
```diff
@@ -163,6 +163,15 @@ foreach bb_file $sc_blackboxes {
 # Synthesis
 ########################################################
 
+# Before working on the design, we mask out any module supplied via 
+# `blackbox_modules`. This allows synthesis of parts of the design without having
+# to modify the input RTL.
+if { [dict exists $sc_cfg tool $sc_tool task $sc_task var blackbox_modules] } {
+    foreach bb [dict get $sc_cfg tool $sc_tool task $sc_task var blackbox_modules] {
+        yosys blackbox $bb
```
gadfort, line 171: could you add a log message about what modules are getting blackboxed?

gadfort (approved):
Looks good, just need to add a log message to make it easier to debug later if needed.

siliconcompiler/tools/yosys/syn_asic.tcl
```diff
@@ -163,6 +163,16 @@ foreach bb_file $sc_blackboxes {
 # Synthesis
 ########################################################
 
+# Before working on the design, we mask out any module supplied via
+# `blackbox_modules`. This allows synthesis of parts of the design without having
+# to modify the input RTL.
+if { [dict exists $sc_cfg tool $sc_tool task $sc_task var blackbox_modules] } {
+    foreach bb [dict get $sc_cfg tool $sc_tool task $sc_task var blackbox_modules] {
+        puts "Blackboxing module: $bb"
```
gadfort, line 171: please use `yosys log ` to log the message for some reason `puts` doesn't get handled correctly in yosys. (also I noticed puts was used just above this, if you have the time to fix that too that would be great, if not don't worry about it)

PR #2345, opened 2024-04-15
add source link to docs

### LINKED ISSUE #2344: Add [source] tags to Docs linking to github source code
I've been attempting to understand the SC source code and I find it difficult to find the source code related to API calls in the documentation. 

Additionally, because of the often dynamic nature of classes in SC python linters don't work which would let me track function calls back to their source code more easily.

For example if looking at the following function:
[SC Docs Chip.input](https://docs.siliconcompiler.com/en/stable/reference_manual/core_api.html#siliconcompiler.Chip.input)

It would be convenient to have a [source] tag following the function header which would link to the github source code and send me to the relevant lines.

Desired functionality can be seen here in the matplotlib docs:
[Matplotlib Axes Plot Docs](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.plot.html#matplotlib.axes.Axes.plot)

What I've been doing as an alternative is to just use a debugger and dive into functions at runtime but it's not ideal.

gadfort:
Closes #2344

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -17,7 +17,7 @@ import inspect
 import importlib
 
 sc_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
-sys.path.insert(0, os.path.join(sc_root, 'siliconcompiler'))
+sys.path.insert(0, sc_root)
 import siliconcompiler  # noqa E402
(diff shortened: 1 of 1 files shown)


gadfort:
https://siliconcompiler--2345.org.readthedocs.build/en/2345/reference_manual/core_api.html

PR #2349, opened 2024-04-15
Consistently use `Chip._getworkdir()`

ubfx:
This gives users the option to override `_getworkdir()` with a custom scheme and keep everything else working.


gadfort (approved):
LGTM, thanks for fixing this

PR #2359, opened 2024-04-21
Fix liberty files templates issues when using multiple liberty files (format)

gadfort:
Builds on #2354 to make formatting consistent and use regex for matching to match method in rest of script.

### DIFF
--- a/siliconcompiler/tools/yosys/mergeLib.py
+++ b/siliconcompiler/tools/yosys/mergeLib.py
@@ -4,14 +4,28 @@
 import re
 
 
-library_match = re.compile(r"^\s*library\s*\(")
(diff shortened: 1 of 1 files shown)


gadfort:
@sebinho can you check to see if these changes still work for your setup?

sebinho:
> @sebinho can you check to see if these changes still work for your setup?

thanks, I will try with your changes.
The reason I did not use regex is because it is extremely slow, especially on big liberty files like mine (400MB +).

sebinho:
I just tried your changes and it seems to work just like before.
Thanks for merging those changes @gadfort

gadfort:
@sebinho hopefully the regexes are only temporary while the yosys/abc enablement is merged.

PR #2362, opened 2024-04-21
add logic depth to schema and expose openroad enablement for it.

gadfort:


aolofsson:
I like it

Questions:

-can you expand on definition of logic depth definition. Is there an official one? Do buffers/inverters count?
-do any tools besides OR report logic depth?

gadfort:
@aolofsson yes, the proprietary tools report this number. This allows us to be tracking. Buffers and inverters count. I wasn't able to find a "ground truth" definition, but it's more or less the length of the critical path in logic gates.

aolofsson (approved):

PR #2366, opened 2024-04-22
Fix non-working openroad setting for default routing via

sebinho:
This commit is fixing the non-working setting that can be used to change the default vias that are used by OpenROAD router.


gadfort (approved):
Thank you for fixing this

PR #2389, opened 2024-04-29
Add post-synthesis static timing analysis in ASIC flow

sebinho:
The goal of this pull request is to add post synthesis STA in the ASIC flow.

### DIFF
--- a/siliconcompiler/tools/openroad/scripts/sc_metrics.tcl
+++ b/siliconcompiler/tools/openroad/scripts/sc_metrics.tcl
@@ -109,24 +109,26 @@ utl::metric_int "design__nets" [llength [[ord::get_db_block] getNets]]
 utl::metric_int "design__registers" [llength [all_registers]]
 
 # get number of buffers
-set bufs 0
-set invs 0
-foreach inst [get_cells -hierarchical *] {
-  set cell [$inst cell]
-  if { $cell == "NULL" } {
-    continue
+if { $::sc_task_synsta != "syn_sta" } {
+  set bufs 0
+  set invs 0
+  foreach inst [get_cells -hierarchical *] {
+    set cell [$inst cell]
+    if { $cell == "NULL" } {
+      continue
+    }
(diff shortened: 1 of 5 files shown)


siliconcompiler/tools/openroad/scripts/sc_metrics.tcl
```diff
@@ -109,24 +109,26 @@ utl::metric_int "design__nets" [llength [[ord::get_db_block] getNets]]
...
+    }
+    set liberty_cell [$cell liberty_cell]
+    if { $liberty_cell == "NULL" } {
+      continue
+    }
+    if { [$liberty_cell is_buffer] } {
+      incr bufs
+    } elseif { [$liberty_cell is_inverter] } {
+      incr invs
+    }
```
gadfort, line 128: any reason this doesn't work?

sebinho: @gadfort this part about buffers does work. I just though this was not important to report at this stage but you can re-add if you think it should remain there.
It is very easy to test I think.

sebinho: @gadfort  Just wondering if there is any plan to accept this merge request. Could you let me know? Thx

gadfort: This fell off my radar. I'm debating how to best incorporate the task. It feels redundant in the `asicflow` since the floorplanning already generates this information. It might be better to have a separate flow to avoid slowing down the `asicflow` any more.

sebinho: I get your point. I was also debating creating a separate flow as well.
This new step is actually very fast in the complete flow.

gadfort:
Closing in favor of: https://github.com/siliconcompiler/siliconcompiler/pull/2466

PR #2408, opened 2024-05-08
remove lib merging from yosys with new yosys version

gadfort:


gadfort:
@sebinho FYI

sebinho:
Thanks 
Will it be an issue if my different LIB files contain templates with the same name?

gadfort:
@sebinho it shouldn't, if it does we would need to get that fixed in ABC. Please try it out and let me know if you have any issues.

gadfort:
Thats great to hear.

PR #2458, opened 2024-05-29
ensure local runs honor thread resources

gadfort:
Adds:
- option/scheduler/maxnodes to schema to control maximum number of parallel jobs allowed in the local scheduler
- honor requested number of threads against local machines total number of threads to avoid overloading system.


aolofsson (approved):
Nice!

PR #2493, opened 2024-06-10
Add OpenROAD command checking with tclint

nmoroze:
I just released a new version of tclint that adds an OpenROAD plugin for checking its tool-specific commands. This PR fixes an issue it caught and demonstrates how you can add it to CI.

Using the plugin is more involved than just enabling it in the regular job - in order to be responsive to OpenROAD API changes, the plugin requires generating a "command spec" in an environment that has OpenROAD installed. I figured the simplest approach to adding it to the SC CI was to add it as a new job to the tool CI workflow. I kept the regular Tcl lint check as a fast failure path.

Here's an example of a [passing](https://github.com/nmoroze/siliconcompiler/actions/runs/9440700733/job/26000427774) and [failing](https://github.com/nmoroze/siliconcompiler/actions/runs/9440506425/job/25999939939) check in my own fork. 

See this [docs page](https://github.com/nmoroze/tclint/blob/main/docs/plugins.md) for more info on the feature. 

This is fairly fresh so no need to merge if you have any concerns, but I'm curious if you have any feedback on the feature! I'm still thinking about how best to design the UX around enabling plugins and managing command specs.

### DIFF
--- a/.github/workflows/tools_ci.yml
+++ b/.github/workflows/tools_ci.yml
@@ -204,7 +204,7 @@ jobs:
           context: ./sc_tools/
           push: true
           tags: ${{ needs.build_tool_builder.outputs.sc_tools }}
-  
+
   runner_image:
     if: always() && github.event_name == 'release'
     needs: [build_tool_builder, build_sc_tools]
@@ -341,3 +341,34 @@ jobs:
           workflow_file_name: containers.yml
           client_payload: '{"sc-ref": "${{ github.event.pull_request.head.ref || github.sha }}"}'
           wait_interval: 60
+
+  lint_tcl:
+    name: Lint OpenROAD TCL scripts
+    if: always()
+    needs: [build_tool_builder, build_sc_tools]
+    runs-on: ubuntu-latest
+    container:
+      image: ${{ needs.build_tool_builder.outputs.sc_tools }}
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ github.token }}
+    env:
+      COMMAND_SPEC_PATH: tclint-openroad-spec.json
+
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v4
+
+      - name: Install tclint
+        run: |
+          python3 -m pip install tclint -c requirements.txt
+
+      - name: Generate OpenROAD command spec
+        run: |
+          tclint-plugins make-spec openroad --output $COMMAND_SPEC_PATH
+
+      - name: Lint with tclint
+        run: |
+          tclint --commands $COMMAND_SPEC_PATH \
+            siliconcompiler/tools/openroad/scripts \
+            siliconcompiler/tools/opensta/scripts
--- a/siliconcompiler/tools/openroad/scripts/sc_rcx_bench.tcl
+++ b/siliconcompiler/tools/openroad/scripts/sc_rcx_bench.tcl
@@ -11,7 +11,7 @@ set openroad_top_metal_number [[[ord::get_db_tech] findLayer $sc_maxmetal] getRo
 # store it in the database
 bench_wires \
   -len $openroad_bench_length \
-  -met $openroad_top_metal_number \
+  -met_cnt $openroad_top_metal_number \
   -all
 
 # Writes the verilog netlist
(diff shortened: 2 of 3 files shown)


yczeng (approved):

lengstrom (commented):
lgtm

lengstrom (approved):

.github/workflows/tools_ci.yml
```diff
@@ -341,3 +341,34 @@ jobs:
...
+  lint_tcl:
+    name: Lint OpenROAD TCL scripts
+    if: always()
+    needs: [build_tool_builder, build_sc_tools]
+    runs-on: ubuntu-latest
+    container:
+      image: ${{ needs.build_tool_builder.outputs.sc_tools }}
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ github.token }}
```
gadfort, line 354: this should not be needed anymore

.github/workflows/tools_ci.yml
```diff
@@ -341,3 +341,34 @@ jobs:
...
+      - name: Check out Git repository
+        uses: actions/checkout@v4
+
+      - name: Install tclint
+        run: |
+          python3 -m pip install tclint -c requirements.txt
+
+      - name: Generate OpenROAD command spec
+        run: |
+          tclint-plugins make-spec openroad --output $COMMAND_SPEC_PATH
```
gadfort, line 368: It might be nice if this was generated by the openroad build image (that would require the command-spec.json format to be fairly stable though

.github/workflows/tools_ci.yml
```diff
@@ -341,3 +341,34 @@ jobs:
...
+
+      - name: Lint with tclint
+        run: |
+          tclint --commands $COMMAND_SPEC_PATH \
+            siliconcompiler/tools/openroad/scripts \
+            siliconcompiler/tools/opensta/scripts
```
gadfort, line 374: I wonder if there is a way to get this into the linting flow?

siliconcompiler/tools/openroad/scripts/sc_rcx_bench.tcl
```diff
@@ -11,7 +11,7 @@ set openroad_top_metal_number [[[ord::get_db_tech] findLayer $sc_maxmetal] getRo
 # store it in the database
 bench_wires \
   -len $openroad_bench_length \
-  -met $openroad_top_metal_number \
+  -met_cnt $openroad_top_metal_number \
```
gadfort, line 14: thanks

gadfort (changes requested):

gadfort:
@nmoroze my main concern with this is that OpenROAD is not running tclint on all their tools so there is a good chance this will become a source of CI failures. I think it would also be nice to see the tclint repo have a CI flow for the opensource tools this plugin supports (the sc_tools:latest image is available if you want to use a prebuilt one).

nmoroze:
@gadfort Thanks for taking a look! 

>  I think it would also be nice to see the tclint repo have a CI flow for the opensource tools this plugin supports (the sc_tools:latest image is available if you want to use a prebuilt one).

Yeah, that makes sense - I'm planning to add CI for this very soon. I guess since the SC bot constantly tracks OpenROAD's main branch I probably won't be able to catch bugs before they'd end up affecting you. 

> my main concern with this is that OpenROAD is not running tclint on all their tools so there is a good chance this will become a source of CI failures
 
What exactly is your concern here? I think OR not running tclint on its own code shouldn't affect this, although I am worried about the fact their CI for checking doc strings doesn't seem to have full coverage right now. I am wondering if some of the code in tclint could help fix this (https://github.com/The-OpenROAD-Project/OpenROAD/issues/5057#issuecomment-2132042452). 

Maybe the thing to do would be to skip adding this to SC CI for now, and I can add CI within tclint that tracks OpenROAD's latest and ensures it remains stable for a little bit. 

In the meantime I can just leave the change to `sc_rcx_bench.tcl` if you want that. Not sure how strongly you feel about using the full argument names, since that's more of a style thing.

gadfort:
@nmoroze maybe if their CI ran the config as well, it would atleast offer some protection against regressions based on this.  I think this is fine to merge.
As for the CI catching failures in OR (I think that is on them, but I'm mainly interested in making sure you have something exercising the tclint plugin code for future releases.

I think as a linting "rule" it makes sense to use full argument names.

gadfort (approved):

PR #2515, opened 2024-06-14
use capsys instead of capfd

gadfort:


gadfort:
https://github.com/siliconcompiler/siliconcompiler/actions/runs/9519617038

PR #2580, opened 2024-06-27
fix handling of lists in collect

gadfort:


gadfort:
https://github.com/siliconcompiler/siliconcompiler/actions/runs/9697813482

PR #2585, opened 2024-06-28
add yosys attributes to netlist

gadfort:


gadfort:
https://github.com/siliconcompiler/siliconcompiler/actions/runs/9715088752

PR #2586, opened 2024-06-28
ensure enum lists get allowed values added in docs

gadfort:


gadfort:
https://siliconcompiler--2586.org.readthedocs.build/en/2586/reference_manual/schema.html#param-datasheet-proc-datatypes-ref

PR #2589, opened 2024-06-29
check timestamps on files when hash is not available

gadfort:


gadfort:
https://github.com/siliconcompiler/siliconcompiler/actions/runs/9725273960

PR #2591, opened 2024-06-30
Add enablement for users to set VPR's max router iterations

petergrossmann21:
VPR has a command-line option --max_router_iterations (default value == 50) that is used to set how many iterations the router will execute before giving up.  The router will also give up if an estimation algorithm computes that the required number of router iterations will exceed the value of max_router_iterations.  I have found in a number of experiments that the estimator sometimes wildly overestimates the number of router iterations that will be required in practice.  This has motivated setting --max_router_iterations to a high value as a workaround.  

More generally, users may wish to take control of this value to debug their designs or to ask VPR to work harder at finding a solution to route their design.


siliconcompiler/tools/vpr/route.py
```diff
@@ -54,6 +54,12 @@ def runtime_options(chip):
     enable_images = chip.get('tool', tool, 'task', task, 'var', 'enable_images',
                              step=step, index=index)[0]
 
+    route_iterations = chip.get('tool', tool, 'task', task, 'var', 'max_router_iterations',
```
gadfort, line 57: This parameter needs a help string that is set in `setup()`

tests/flows/test_vpr_max_router_iterations.py
```diff
@@ -0,0 +1,44 @@
...
+    chip.set('tool', 'vpr', 'task', 'route', 'var', 'max_router_iterations', '300')
+
+    # 3. Load target
+    chip.load_target(fpgaflow_demo)
+
+    chip.run()
+
+    route_file = chip.find_result('route', step='route')
+
+    assert os.path.exists(route_file)
```
gadfort, line 39: This test doesn't actually check if this parameter is getting set or used. Let's talk tomorrow about how we should structure this test to ensure proper coverage.

tests/flows/test_vpr_max_router_iterations.py
```diff
@@ -0,0 +1,44 @@
+import os
```
gadfort, line 1: You don't need a separate file for each test. We probably need to merge a bunch of these together make it a little cleaner.

gadfort (changes requested):

PR #2602, opened 2024-07-02
fix wheel build

gadfort:


gadfort:
https://github.com/siliconcompiler/siliconcompiler/actions/runs/9766993787

PR #2796, opened 2024-08-29
Format Tcl code using `tclfmt`

nmoroze:
Announcing that `tclint` finally includes a formatter (`tclfmt`) by way of this PR!

The first commit is the result of running the formatter on the code (`tcfmt --in-place .`). It mostly follows `tclint`'s style guide, and uses the same configuration. There are two categories of changes that are responsible for most of the diff that perhaps might be controversial:

- `tclfmt` pushes the braces for multi-line expressions onto different lines. There were a couple reasons for doing this, the main one is that it consistently looks good across differently configured indent levels (the style currently used by this repo can sometimes have a case where the expression of an if statement is indented the same as the body, for example). 
- It breaks `set` alignment. This was just a case of being more opinionated for simplicity, although I'm open to making `tclfmt` respect the `allow-aligned-sets` config if you feel strongly about preserving these.

I wrote up some docs describing usage of the utility plus a description of the style guide with some rationale, so that might be interesting to look through: https://github.com/nmoroze/tclint/blob/v0.4.0/docs/tclfmt.md.

The second commit adds the formatter to CI. Turns out it's much easier to comprehensively enforce a consistent style by diffing against formatted code rather than checking for different style violations piecemeal. Because of this, `tclfmt` is gonna be the canonical way of enforcing style going forward, and I'll eventually remove style checks from `tclint` itself. To preview this future behavior and reduce output noise, `tclint` got a new flag `--no-check-style`. `tclfmt` also doesn't automatically reducing line length yet, so `--no-check-style` still reports line length violations. 

As always, I'm open to any feedback!

### DIFF
--- a/siliconcompiler/tools/openroad/scripts/sc_dfm.tcl
+++ b/siliconcompiler/tools/openroad/scripts/sc_dfm.tcl
@@ -9,8 +9,10 @@ foreach obstruction [[ord::get_db_block] getObstructions] {
 }
 utl::info FLW 1 "Deleted $removed_obs routing obstructions"
 
-if { $openroad_fin_add_fill == "true" && \
-     [sc_cfg_exists pdk $sc_pdk aprtech openroad $sc_stackup $sc_libtype fill] } {
+if {
+  $openroad_fin_add_fill == "true" &&
+  [sc_cfg_exists pdk $sc_pdk aprtech openroad $sc_stackup $sc_libtype fill]
+} {
   set sc_fillrules \
     [lindex [sc_cfg_get pdk $sc_pdk aprtech openroad $sc_stackup $sc_libtype fill] 0]
   density_fill -rules $sc_fillrules
--- a/siliconcompiler/tools/openroad/scripts/sc_procs.tcl
+++ b/siliconcompiler/tools/openroad/scripts/sc_procs.tcl
@@ -19,8 +19,10 @@ proc sc_global_placement_density {} {
   # User specified adjustment
   if { $openroad_gpl_uniform_placement_adjustment > 0.0 } {
     set or_uniform_adjusted_density \
-      [expr { $or_uniform_density + ((1.0 - $or_uniform_density) * \
-              $openroad_gpl_uniform_placement_adjustment) + $or_adjust_density_adder }]
+      [expr {
+        $or_uniform_density + ((1.0 - $or_uniform_density) *
+          $openroad_gpl_uniform_placement_adjustment) + $or_adjust_density_adder
+      }]
     if { $or_uniform_adjusted_density > 1.00 } {
       utl::warn FLW 1 "Adjusted density exceeds 1.00 ([format %0.3f $or_uniform_adjusted_density]),\
         reverting to use ($openroad_gpl_place_density) for global placement"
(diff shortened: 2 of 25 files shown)


siliconcompiler/tools/openroad/scripts/sc_dfm.tcl
```diff
@@ -9,8 +9,10 @@ foreach obstruction [[ord::get_db_block] getObstructions] {
 }
 utl::info FLW 1 "Deleted $removed_obs routing obstructions"
 
-if { $openroad_fin_add_fill == "true" && \
-     [sc_cfg_exists pdk $sc_pdk aprtech openroad $sc_stackup $sc_libtype fill] } {
+if {
+  $openroad_fin_add_fill == "true" &&
+  [sc_cfg_exists pdk $sc_pdk aprtech openroad $sc_stackup $sc_libtype fill]
```
gadfort, line 14: I don't know why, but I find this formatting to be a little harder to read. I'm not sure what the better solution is, but I think it's because it is indent aligned with the code in the body that makes it look like it's part of that code.

siliconcompiler/tools/openroad/scripts/sc_procs.tcl
```diff
@@ -19,8 +19,10 @@ proc sc_global_placement_density {} {
   # User specified adjustment
   if { $openroad_gpl_uniform_placement_adjustment > 0.0 } {
     set or_uniform_adjusted_density \
-      [expr { $or_uniform_density + ((1.0 - $or_uniform_density) * \
-              $openroad_gpl_uniform_placement_adjustment) + $or_adjust_density_adder }]
+      [expr {
+        $or_uniform_density + ((1.0 - $or_uniform_density) *
+          $openroad_gpl_uniform_placement_adjustment) + $or_adjust_density_adder
```
gadfort, line 24: why is this indenting different? it seems like these two lines should have been aligned

siliconcompiler/tools/openroad/scripts/sc_dfm.tcl
nmoroze: Yeah, that makes sense. The one problem is that that someone who, for example, configured a 4-space indent and no spaces in braces would get something even less readable for a similar reason:

```
if {$openroad_fin_add_fill == "true" && \
    [sc_cfg_exists pdk $sc_pdk aprtech openroad $sc_stackup $sc_libtype fill]} {
    set sc_fillrules \
         [lindex [sc_cfg_get pdk $sc_pdk aprtech openroad $sc_stackup $sc_libtype fill] 0]
```

The hope is that the extra line of `} {` helps a bit, but I get that it's not a very substantial separator.

One compromise thought I had was to cue off the existing formatting, where `tclfmt` will do the aligned style like you have if there's no newline before the start of the expression, and will do the current formatting if the user starts the expression on a new line themselves. This complicates the implementation a bit, but I can take another look at supporting it.

siliconcompiler/tools/openroad/scripts/sc_procs.tcl
nmoroze: That's cause there's a parenthesized expression spanning both lines. The formatter does that to try and make parenthesized groupings clear, but if that's more distracting than helpful I can reconsider.

gadfort: I see that now.

siliconcompiler/tools/openroad/scripts/sc_dfm.tcl
gadfort: I see your point. The python flake8 errors when the indent of the if and following sections align, maybe in this case the second line needs to be indented one more time?

gadfort (approved):
Lgtm

nmoroze:
Thanks for merging, sorry I dropped the thread re: expression indentation style.

I can look into alternatives if you'd still prefer a different style.

gadfort:
@nmoroze no worries, I don't feel so strongly about the indents, so I think it's fine. I'll work my way through the repos that would benefit from this and let you know if anything changes or if any issues arise.

PR #2813, opened 2024-09-05
Updating package pin schema

aolofsson:
- Removing net/port parameters doesn't belong here. The datasheet should not describe things intenal to device, slipper slope.

- Adding a 'pin' key for better grouping. Previously I had optimized brevity over clarity (bad choice)

- Adding missing logical pin 'name' mapping to package

### DIFF
--- a/siliconcompiler/schema/schema_cfg.py
+++ b/siliconcompiler/schema/schema_cfg.py
@@ -1259,7 +1259,7 @@ def schema_datasheet(cfg, name='default', mode='default'):
                 switch=f"-datasheet_package_pin_{i} 'name pinnumber <(float,float,float)>'",
                 example=[
                     f"cli: -datasheet_package_pin_{i} 'abcd B1 {v[1]}'",
-                    f"api: chip.set('datasheet', 'package', 'abcd', 'B1', '{i}', {v[1]}"],
+                    f"api: chip.set('datasheet', 'package', 'abcd', 'pin', 'B1', '{i}', {v[1]}"],
                 schelp=f"""Datsheet: {v[0]} specified on a per package and per pin number
                 basis. Values are tuples of (min, nominal, max).""")
 
--- a/siliconcompiler/schema/schema_cfg.py
+++ b/siliconcompiler/schema/schema_cfg.py
@@ -1281,7 +1281,7 @@ def schema_datasheet(cfg, name='default', mode='default'):
             switch="-datasheet_package_pin_name 'name pinnumber <str>'",
             example=[
                 "cli: -datasheet_package_pin_name 'abcd B1 clk'",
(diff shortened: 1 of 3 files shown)


siliconcompiler/schema/schema_cfg.py
```diff
@@ -1232,62 +1234,60 @@ def schema_datasheet(cfg, name='default', mode='default'):
...
-                "cli: -datasheet_package_pinloc 'abcd B1 (0.5,0.5)'",
-                "api: chip.set('datasheet', 'package', 'abcd', 'pinloc', 'B1', (0.5,0.5)"],
-            schelp="""Datsheet: Pin location specified as an (x,y) tuple. Locations
-            specify the center of the pin with respect to the center of the package.
-            """)
+                "cli: -datasheet_package_pin_loc 'abcd B1 (0.5,0.5)'",
+                "api: chip.set('datasheet', 'package', 'abcd', 'pin', 'B1', 'loc', (0.5,0.5)"],
+            schelp="""Datsheet: Package pin location specified as an (x,y) tuple on a per
+            package and per pin number basis. Locations specify the center of the pin with
+            respect to the center of the package.""")
```
gadfort, line 1276: I'm not sure about the center of the package thing, it feels awkward to define things relative to the center at the package level

siliconcompiler/schema/schema_cfg.py
```diff
@@ -1232,62 +1234,60 @@ def schema_datasheet(cfg, name='default', mode='default'):
...
-            switch="-datasheet_package_pinshape 'name name <str>'",
+            switch="-datasheet_package_pin_shape 'name pinnumber <str>'",
             example=[
-                "cli: -datasheet_package_pinshape 'abcd B1 circle'",
-                "api: chip.set('datasheet', 'package', 'abcd', 'pinshape', 'B1', 'circle')"],
-            schelp="""Datasheet: pin shape (rectangle or circle) specified on a per package
-            and per pin basis.""")
+                "cli: -datasheet_package_pin_shape 'abcd B1 circle'",
+                "api: chip.set('datasheet', 'package', 'abcd', 'pin', 'B1', 'shape', 'circle')"],
+            schelp="""Datasheet: package pin shape (rectangle or circle) specified on a per package
```
gadfort, line 1247: remove the shape suggestions from the help since it's captured in the `enum`

gadfort (approved):
Looks fine, if you are planning a follow on PR with the help fixes what should be okay.

siliconcompiler/schema/schema_cfg.py
aolofsson: Package pins and any interactions withh the OSATs always refers to the origin being at the center of the package. In the chip world, the origin is in the lower left.

PR #2815, opened 2024-09-06
Changing the default origin from center to lower left corner

aolofsson:
- Making the default ASIC design, our bread and butter
- @gadfort pointed out that since the default in the ASIC world is lower left corner, using the center coordinate adds extra work/code for the user and in the tool specific driver....non intuitive.
- Adding origin parameter to support packaging and board design


gadfort (approved):
LGTM

PR #2816, opened 2024-09-07
Adding basic schematic/netlist capability to schema

aolofsson:
I know this is going to be controversial, because it's a slippery slope, but there have been a number of times in the last 3 years that we have needed some way of keeping track of the instances in the design. As we move siliconcompiler into new areas, this need is only going to grow. 

We could punt and say that schematic functionality is going to be put inside SC derivative projects, but that is going to increase the overall development effort. The data structure for small structured netlists is simple enough that I think we can afford to add it to the SC schema. This solves the issue with the derivative project I was working on today

Why yes:
- path towards checking use dependencies versus top level netlist
- path to connecting verilog netlist to layout (at least for structure hard designs), like boards, packages.
- path to interactive query of netlist from python within sc  (without heavy external dependency).
- deeper offline post run analysis from sc

For the net, gets a bit awkward, but not much worse than existing file formats. How to support primary pins, pins on cells, and cells with an implid pin (like bump cells) with one single tuple or string?

**Option 1 (current  tuple):**
self.package.add('schematic', 'net', net, ("I0", "Z") # hier pin
self.package.add('schematic', 'net', net, ("", "CLK") # primary pin, mandatory emopty string in first field.
self.package.add('schematic', 'net', net, ("bump", "") # special case bump, mandatory empty string in second field

**Option 2 (use single string and a dot hiearchy character)**
self.package.add('schematic', 'net', net, ("I0.Z") # hier pin
self.package.add('schematic', 'net', net, ("CLK") # primary pin (absence of hierchar), must be in pins section
self.package.add('schematic', 'net', net, ("bump") #must be in components section, special case/short cut. Should have pin.

**Option 3 (variation of option 2, also with new syntax)**
self.package.add('schematic', 'net', net, 'connect', ("I0.Z") # internal component pin
self.package.add('schematic', 'net', net, 'connect', ("CLK") # primary pin (absence of hierchar), must be in pins section
self.package.add('schematic', 'net', net, 'connect', ("bump.*") #specifies that all pins of bump should be connected to net 

**Schematic Rules (consistent with Verilog):**
-there can be no conflict betweem component/instance names and pin/net names
-a net name of the same name is implied for every pimaary I/O pin
-"default" is a reserved key word, cannot be used as a pin/met/component name
-"." is the default hierarchy character for specifying component pins 
-"[]" is the default bus character for aggregating bits for ports/nets by external tool. 

**Resolution After Review with @gadfort**:
- Option 3 
- Change tuples into SC consistent dictionary keys. Tuples are non-standard. Usability issue will be handled through helper functions. Speed should not be an issue on read/write.


siliconcompiler/schema/schema_cfg.py
```diff
@@ -215,6 +215,74 @@ def schema_cfg():
...
+    scparam(cfg, ['schematic', 'buschar'],
+            sctype='str',
+            defvalue='[]',
+            shorthelp="Schematic: bus character",
+            switch="-schematic_buschar <str>",
+            example=["cli: -schematic_buschar '[]'",
+                     "api: chip.set('schematic', 'buschar', '[]')"],
+            schelp="""Specifies the character used to express bus bits. If the
+            bus character is used as part of a name, it must be
+            escaped with a backslash('\').""")
```
gadfort, line 284: do we really want to carry this around? I have yet to see anything other than []

siliconcompiler/schema/schema_cfg.py
```diff
@@ -215,6 +215,74 @@ def schema_cfg():
...
+
+    scparam(cfg, ['schematic', 'pin'],
+            sctype='[(str,str)]',
+            shorthelp="Schematic: pins",
+            switch="-schematic_pin 'name <(str,str)>'",
+            example=["cli: -schematic_pin '(A,input)",
+                     "api: chip.set('schematic', 'pin', ('A','input'))"],
+            schelp="""Primary design I/O pins specified as a list of
+            (NAME,DIRECTION) tuples. Name can be any legal verilog string
+            name while DIRECTION is limited to input, output, and inout.""")
```
gadfort, line 248: why not make this
`schematic`, `pin`, name, `direction`, `enum`? this way the schema can enforce the legal values

siliconcompiler/schema/schema_cfg.py
```diff
@@ -215,6 +215,74 @@ def schema_cfg():
...
+    '''
+
+    scparam(cfg, ['schematic', 'component'],
+            sctype='[(str,str)]',
+            shorthelp="Schematic: components",
+            switch="-schematic_component <(str,str)>",
+            example=["cli: -schematic_component (B0,NAND2X1)",
+                     "api: chip.set('schematic', 'component', ('B0','NAND2X1'))"],
+            schelp="""Components in the design, specified as a list of
+            (INSTANCE,MODEL) tuples.""")
```
gadfort, line 238: `schematic`, `component`, name, `model`, model? that is much more inline with the other schema keys

tests/core/test_schematic.py
```diff
@@ -0,0 +1,35 @@
...
+                                  ('OUT', 'output')])
+
+    # nets
+    chip.add('schematic', 'net', 'IN0', 'I0.A')
+    chip.add('schematic', 'net', 'IN1', 'I0.B')
+    chip.add('schematic', 'net', 'IN2', 'I1.A')
+    chip.add('schematic', 'net', 'IN3', 'I1.B')
+    chip.add('schematic', 'net', 'INTA', ['I0.Z', 'I2.A'])
+    chip.add('schematic', 'net', 'INTB', ['I1.Z', 'I2:B'])
+    chip.add('schematic', 'net', 'OUT', ['I2.Z'])
```
gadfort, line 30: this test doesn't actually check anything. it would only fail is the keys changed, but we have tests that do that already. 
I'm not sure it is worth having unless it does something additional checking.

gadfort (changes requested):

siliconcompiler/schema/schema_cfg.py
aolofsson: Will keep as discussed. Some weird schematic things floating around like <0>, <1> in some netlists.

siliconcompiler/schema/schema_cfg.py
aolofsson: Fixed and committed.

tests/core/test_schematic.py
aolofsson: Added a basic assertion for connecting netlist to pins to components. Caught a bug in test.

siliconcompiler/schema/schema_cfg.py
aolofsson: Fixed and committed.

gadfort (approved):

PR #2914, opened 2024-10-18
Grossmann/add vpr device option

petergrossmann21:
VPR has a ``--device`` option that defaults to the value ``auto``.  The default value is intended to set VPR's default behavior to automatically select a minimum size array for the circuit being mapped, which it normally does by processing the content of an ``<auto_layout>`` section of the XML file.  However, VPR is smart enough to use ``<fixed_layout>`` sections if no ``<auto_layout>`` section is provided and the default ``--device`` setting is in place.

The above behavior is not really appropriate for an SC flow, which is aimed at producing output for a specific architecture size.  As a result, this PR proposes to make setting ``--device`` to a valid value in all cases using a part name variable called ``device_code``.  This will have the following effects

* All FPGA part drivers will need to have an equivalent to the following

``chip.set('fpga', part_name, 'var', 'device_code', <string>)``

* It will still be possible to use VPR in its traditional size-independent manner by setting this variable's value to ``auto`` in the applicable FPGA part driver.
* Writing a part driver for an XML file that has multiple subsections to its ``<layout>`` tag (examples:  both ``<auto_layout>`` and ``<fixed_layout>`` in the same file, multiple ``<fixed_layout>`` sections in the same file) will have a clear "best practice" to define each layout option as a separate part name in the same architecture family's Python module, and simply have the same architecture XML file specified for each part with the correct device code specified for each part name.  In such a regime, providing an rr_graph XML file will not be appropriate unless a separate RR graph is provided for each device code.

The above observations probably beg the question why part name and device code should ever have different values or be independent variables.  For logik testing/validation of revisions of existing architectures it is necessary to introduce a temporary part name to disambiguate the new version from the old version as an assist to the package manager.  While the device code is a fixed string inside an XML file, the part name can be selected independently of any file content.  For this reason I do not recommend forcing them to be identical in all circumstances, even if their having equal values will be a common case.  The existing CI suite in fact has divergence between part names and device codes already due to some historical phenomena, so this case is already tested.

### DIFF
--- a/siliconcompiler/tools/vpr/vpr.py
+++ b/siliconcompiler/tools/vpr/vpr.py
@@ -66,7 +66,7 @@ def setup_tool(chip, clobber=True):
         chip.add('tool', tool, 'task', task, 'require', f'fpga,{part_name},var,{resource}',
                  step=step, index=index)
 
-    chip.add('tool', tool, 'task', task, 'require', f'fpga,{part_name},var,device_code',
+    chip.add('tool', tool, 'task', task, 'require', f'fpga,{part_name},var,vpr_device_code',
              step=step, index=index)
 
 
@@ -79,7 +79,7 @@ def runtime_options(chip):
 
     options = []
 
-    device_code = chip.get('fpga', part_name, 'var', 'device_code')
+    device_code = chip.get('fpga', part_name, 'var', 'vpr_device_code')
 
     options.append(f"--device {device_code[0]}")
 
--- a/siliconcompiler/tools/vpr/vpr.py
+++ b/siliconcompiler/tools/vpr/vpr.py
@@ -66,6 +66,16 @@ def setup_tool(chip, clobber=True):
         chip.add('tool', tool, 'task', task, 'require', f'fpga,{part_name},var,{resource}',
                  step=step, index=index)
 
+    add_tool_requirements(chip)
+
+def add_tool_requirements(chip):
+
+    step = chip.get('arg', 'step')
+    index = chip.get('arg', 'index')
+    tool, task = get_tool_task(chip, step, index)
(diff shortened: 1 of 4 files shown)


siliconcompiler/tools/vpr/vpr.py
```diff
@@ -66,6 +66,9 @@ def setup_tool(chip, clobber=True):
         chip.add('tool', tool, 'task', task, 'require', f'fpga,{part_name},var,{resource}',
                  step=step, index=index)
 
+    chip.add('tool', tool, 'task', task, 'require', f'fpga,{part_name},var,device_code',
+             step=step, index=index)
```
gadfort, line 70: rename `vpr_devide_code`

siliconcompiler/tools/vpr/vpr.py
```diff
@@ -76,6 +79,10 @@ def runtime_options(chip):
 
     options = []
 
+    device_code = chip.get('fpga', part_name, 'var', 'device_code')
+
+    options.append(f"--device {device_code[0]}")
```
gadfort, line 84: make this a helper function so genfasm can also get these arguments

gadfort (changes requested):

siliconcompiler/tools/vpr/vpr.py
petergrossmann21: bitstream.py current gets these by calling `vpr.runtime_options()` -- is there any benefit to restructuring genfasm's tool driver such that there is a genfasm.py `runtime_options()` that does this instead, and then having bitstream.py call genfasm.py's `runtime_options()`?

gadfort (approved):

PR #2967, opened 2024-11-05
allow oversubscribe in slurm CI

gadfort:


gadfort:
Before
```
85.18s call     tests/server/test_slurm_server.py::test_gcd_server
80.29s call     tests/flows/test_slurm_local.py::test_slurm_local_py
```
After
```
78.94s call     tests/server/test_slurm_server.py::test_gcd_server
74.25s call     tests/flows/test_slurm_local.py::test_slurm_local_py
```
Not worth it

PR #2971, opened 2024-11-05
add metrics task to openroad and add cell types to schema

gadfort:
adds:
* macroarea
* stdcellarea
* padcellarea (could be omitted if we want)


aolofsson (approved):
Incredible example of agile development!

PR #3018, opened 2024-11-27
FPGA Synthesis Adjustment

petergrossmann21:
This change consists of moving the macrolib extract pass up in the flow to prevent extract detection from failing.

Ideally I would add an FPGA synthesis test to this PR so that the functionality has test coverage.  The minimum requirements for such a test are:  a yosys techlib with a macro defined for extraction detection; a design that contains the RTL pattern in the macro library; a pytest that executes synthesis on the design with the yosys techlib specified and verifies by inspecting yosys statistics that the output netlist contains an instance of the techlib macro.

There are numerous ways to write such a test, so some guidance on how to do so in a way that fits with existing SC test methodology would be appreciated.


gadfort:
It would be helpful to have a test that incorrectly maps before this change and correctly maps after this change.
To verify we probably want to load the stat.json and assert the correct number of instances (of the macro) are present.

petergrossmann21:
> It would be helpful to have a test that incorrectly maps before this change and correctly maps after this change. To verify we probably want to load the stat.json and assert the correct number of instances (of the macro) are present.

Here's what I know after looking into this for a while:

1.  The issue this attempts to resolve is confined to extraction of macros that contain adders; pure multipliers extract correctly on current main.  Existing examples have DSP blocks that are pure mulitpliers, so they don't expose the issue.  This means that a new example architecture will need to be provided.  logik_demo_mini is suitable for this purpose.  A decision will be needed as to how to integrate this architecture with existing examples (rename it?  Separate FPGA family driver or graft on to existing?)
2. I will need a recipe for loading the stat.json that complies with required/desired coding styles.

gadfort (approved):

PR #3039, opened 2024-12-08
fix upperbound on packaging as streamlit now handles that

gadfort:


HU90m:
@gadfort Why has the version of packaging capped? It looks like streamlit is OK with any version of packaging >= 20: https://github.com/streamlit/streamlit/blob/408cbbdbab855cddc5c572ba6db3d133e399fead/lib/setup.py#L46

context: I'm just looking at packaging SilconCompiler for Nix, so I've been looking at the dependencies.

gadfort:
@HU90m I'm not 100% sure I remember the reasoning here. It looks like they released 25. If you want move the cap to 26 and PR the change I'd be happy to merge if it passes.

HU90m:
Thanks Peter, I've opened https://github.com/siliconcompiler/siliconcompiler/pull/4605

PR #3046, opened 2024-12-09
add python packages to schema

gadfort:
Adds:
- pythonversion and pythonpackage to schema records to allow for better provenance
- add testing to ensure functionality


aolofsson (approved):
A great addition!

PR #3147, opened 2025-01-24
apply colors to log levels

gadfort:
Why:
- adding colors to the logging in the terminal makes it easier to find errors and warnings

Before:
image

New:
image

Choice of colors can be adjusted.


aolofsson (approved):
Wow, such a cool feature!

PR #3296, opened 2025-03-05
add VPR timing extraction and update VPR to 9.0.0

gadfort:


siliconcompiler/tools/openroad/_apr.py
petergrossmann21, line 1: This seems unrelated; was it intended to be part of a different PR?

petergrossmann21 (approved):
LGTM

siliconcompiler/tools/openroad/_apr.py
gadfort: ahh, yes that should be removed

PR #3407, opened 2025-04-04
add a layout class and improve responsiveness

gadfort:
@Ciprian167 I created a PR for us to discuss this in.
I like the log appears be correct, but if I down size the window and then upsize it the log doesn't return to full size (repeatedly doing this causes the log to shrink to almost nothing).

### DIFF
--- a/siliconcompiler/report/dashboard/cli/__init__.py
+++ b/siliconcompiler/report/dashboard/cli/__init__.py
@@ -8,6 +8,7 @@ from collections import deque
 from dataclasses import dataclass, field
 from typing import List, Dict
 
+from rich import box
 from rich.theme import Theme
 from rich.live import Live
 from rich.table import Table
@@ -110,6 +111,7 @@ class Layout:
             after accounting for reserved space, job board, and log sections.
             Returns 0 if the total height is not set.
     """
+
     height: int = 0
     width: int = 0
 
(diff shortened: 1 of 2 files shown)


siliconcompiler/report/dashboard/cli/__init__.py
```diff
@@ -58,6 +58,24 @@ def get_lines(self):
...
+            # Create a list of the current contents (newest last)
+            current_contents = list(self.buffer)
+
+            # If we have more items than new_size, keep only the last 'new_size' items
+            if len(current_contents) > new_size:
+                current_contents = current_contents[-new_size:]
+
+            # Create a new deque with the new maxlen and preserved items
+            self.buffer = deque(current_contents, maxlen=new_size)
+
```
gadfort, line 78: could this be accomplished by just passing in how many lines we want to get from `get_lines()`?
It seems like resizing the buffer would not be needed and just returning the last `N` lines from the buffer would be sufficient

Ciprian167: I have to test it more. That was might first instinct as well, but something didn't work when I tried it.

Ciprian167: So I think you need to have the buffer size the same as displayed log_size, in order to get that effect. The reason being we render many more times than pushing the the buffer. The `_render` can trigger in two places, logger writes something, and when we recieve updates about the jobs. Because of this disparity between the number of pushes into the buffer and renders, it's actually hard to know when to pop thing out of the buffer. So in order to get that scrolling effect, we give the decision of "pop-ing" thing out when we push in the buffer. But for that to work, the buffer needs to have the right size. 

It's not super elegant, and not very efficient. I'll look for other ways.

Ciprian167: Well, you're right. I was worried about losing a few lines, so was focusing on the first lines in the buffer. But if we don't mind losing some, that will work.

Ciprian167:
I think removing the buffer resize fixes problem you were noticing with the shrinking log when resizing the terminal.

Ciprian167:
@gadfort there still some issues with `end_of_run`. I don't see that being called, and because of that it doesn't report the run summary, nor does it restore the terminal.

gadfort:
@Ciprian167 I'll take a look (do you know under what circumstances it's not getting called?), I'm still getting this strange shrinking log issue, so I'm going to try and debug that.

gadfort:
@Ciprian167 I'm going to merge this since it seems to have fixed the log issue

Ciprian167:
> @Ciprian167 I'll take a look (do you know under what circumstances it's not getting called?), I'm still getting this strange shrinking log issue, so I'm going to try and debug that.

I'm running the `gcd` example. I have one error in the `write.gds` step because I don't have klayout installed. I keep it like that so I can see how things look like with an error. Could errors in the execution be preventing `end_of_run`?

PR #3428, opened 2025-04-09
commandline dashboard

gadfort:
Adds a commandline dashboard, which is enabled by `chip.dashboard(type="cli")`.
The plan is to switch the main run over to use this once it's been through it's paces.

Thank you @Ciprian167 for your contribution.


Ciprian167:
Thanks @gadfort. 

I think the header needs to be in place now that we have the error and warnings added in. But I think it works out nicely!

The only thing I would change would be the __JOB_BOARD_BOX to  `box.SIMPLE_HEAD` which would remove a bunch of the lines, but still leave the header nicely underlined. 

`__JOB_BOARD_BOX = box.MINIMAL_DOUBLE_HEAD`
image

`__JOB_BOARD_BOX = box.SIMPLE_HEAD`
image

Not sure how you feel about it, but we could expose that setting to the user. 

But either way, it's great! Very nice work.

gadfort:
> Thanks @gadfort.
> 
> I think the header needs to be in place now that we have the error and warnings added in. But I think it works out nicely!
> 
> The only thing I would change would be the __JOB_BOARD_BOX to `box.SIMPLE_HEAD` which would remove a bunch of the lines, but still leave the header nicely underlined.
> 
> `__JOB_BOARD_BOX = box.MINIMAL_DOUBLE_HEAD` image
> 
> `__JOB_BOARD_BOX = box.SIMPLE_HEAD` image
> 
> Not sure how you feel about it, but we could expose that setting to the user.
> 
> But either way, it's great! Very nice work.

@aolofsson what do you think? I would be okay with suggested change.

aolofsson:
They both look great! I am find ether way,  if I had to pick it would be the first one, but I am biased towards "tables".

PR #3464, opened 2025-04-24
cli dashboard: ensure errors get priority in printing and print final table

### LINKED ISSUE #3463: Display errors on the last CLI dash rendering
In large flowgraphs, the failing nodes are not visible on the last rendering. As an example checkout the output from `gcd.py`. 

Image

That doesn't make it a very nice UX. If something fails, it would be nice to have immediate access to the log, and job name.

As a quick fix, maybe we can give higher priority to errors. 

```python 
<<< priority_node = {0: startnodes.union(error_nodes)}

>>> priority_node = {-1: startnodes.union(error_nodes)}
```

Image

There probably is a better way to handle this. Also, do you think it makes sense that on the last rendering to show the entire error list, outside of the `live display`. That would allow the user to scroll through all the errors without them being truncated by the `live` display.

gadfort:
Closes #3463 

@Ciprian167 try this out and see if it works for you.

There was a bug that caused the priority printing order to overwrite the errors.

I also switched the final render to show the entire table (not 100% sold on that one, yet).

### DIFF
--- a/siliconcompiler/report/dashboard/cli/__init__.py
+++ b/siliconcompiler/report/dashboard/cli/__init__.py
@@ -553,21 +553,21 @@ class CliDashboard(AbstractDashboard):
         finally:
             try:
                 if live:
-                    live.update(self._get_rendable(final=True), refresh=True)
+                    live.update(self._get_rendable(), refresh=True)
                     live.stop()
                 else:
-                    self._console.print(self._get_rendable(final=True))
+                    self._console.print(self._get_rendable())
             finally:
                 # Restore the prompt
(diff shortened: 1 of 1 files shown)


Ciprian167:
Looks good. 

The final rendering doesn't work, but let me rephrase it. I was thinking to show just the errors. For example say that the job board height is 10 and we have 15 failed jobs. It would be nice to see all these failed jobs at the end, but not necesarily all the passed ones as it can get quite big.

gadfort:
@Ciprian167 lets make that a separate issue then, I removed it from this PR.

PR #3478, opened 2025-04-29
cli dashboard support mutliple runs in a single session

gadfort:
TODO:
- [x]  need to rework how the jobs board is sorted to keep related nodes together: https://github.com/siliconcompiler/siliconcompiler/pull/3480
- [x] need to put some better testing in for multiple jobs

### DIFF
--- a/siliconcompiler/report/dashboard/cli/__init__.py
+++ b/siliconcompiler/report/dashboard/cli/__init__.py
@@ -58,13 +58,13 @@ class CliDashboard(AbstractDashboard):
         """
         Stops the dashboard rendering thread and ensures all rendering operations are completed.
         """
-        self.stop()
+        self._dashboard.end_of_run(self._chip)
 
     def stop(self):
         """
         Stops the dashboard rendering thread and ensures all rendering operations are completed.
         """
-        self._dashboard.stop()
+        self._dashboard.end_of_run(self._chip)
 
         # Restore logger
         if self.__logger_console:
--- a/siliconcompiler/report/dashboard/cli/board.py
+++ b/siliconcompiler/report/dashboard/cli/board.py
@@ -3,10 +3,10 @@ import threading
 import logging
 import queue
 import time
+import multiprocessing
 
 from collections import deque
 from dataclasses import dataclass, field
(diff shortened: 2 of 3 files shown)


Ciprian167:
I was starting the jobs using `subprocess`. But I changed that to simple threading and seemed to have fixed the problem. Now both `chip.run`s are being displayed in the same board. 

To make it work with subprocesses, we might have to use a global `multiprocessing.Manager` in BoardSingleton. 

Also, another issue seems to be the `end_of_run` handling. I have two runs take different amount of time. When the first run finished it halts the dashboard printing for the longer one. 

I tried something like this and it works. Not the most elegant. 

```python
    def stop(self):
        """
        Stops the dashboard rendering thread and ensures all rendering operations are completed.
        """
        with self._render_data_lock:
            total = self._render_data.total
            finished = self._render_data.finished

        if finished == 0 or finished != total:
            return

```

gadfort:
> I'm testing using a setup similar to [parallel job execution](https://docs.siliconcompiler.com/en/stable/user_guide/tutorials/parallel.html) and doesn't seem to work.
> 
> I have two jobs running in parallel, and it seems that it creates two separate `boards` one for each chip.
> 
> ```
> <siliconcompiler.report.dashboard.cli.board.Board object at 0x7f187d2702f0>
> <siliconcompiler.core.Chip object at 0x7f1858be7e30>
> 
> <siliconcompiler.report.dashboard.cli.board.Board object at 0x7f18df1c0440>
> <siliconcompiler.core.Chip object at 0x7f1858be7ef0>
> ```
> 
> In this case, I expect the board object to be the same, right?

Within a single python "session" the singleton is a singleton, but in a multiprocessing state it's a different memory so the multiprocessing.Manager handles communications between the two sessions.

I don't think supporting `subprocess` is worth it, since it adds all kinds of overhead with getting different python instances to communicate.

As for end of run, I didn't see that behavior, but the finished around the same time, so maybe I just missed it.

gadfort:
@Ciprian167 I switch it such that the dashboard only stops when all jobs are reported as stopped.

Ciprian167:
That works. Thank you.

PR #3511, opened 2025-05-05
change schema enum to just a type

gadfort:


aolofsson (changes requested):
As discussed, consider removing "enum" keyword, since type can be detected via < >.

aolofsson (approved):
Looks great!

Issue #3506, opened 2025-05-05
Better runtime performance on large flowgraphs


Ciprian167:
I'm seeing quite bad performance on larger graphs. I'm running a typical DV flow with `compile`, `elaborate` and `test` steps, with the test step having around 320 indices. In this large graph a rerun takes around `160s`. I'm not changing anything between runs, so all those `~160s` seems to be spend inside siliconcompiler to check the graph and so on. 

I enabled `cProfile` in the `scheduler/__init__.py:run`. From what I can see there are two functions that consume the most time:
- ` _reset_flow_nodes`: ~50.6s
- `_local_process`: >100s

Going further down the stack, there seems to be two obvious things that stand out. 

1. `flowgraph.py:99(_get_flowgraph_node_outputs)` - takes about `52s` and is called about `7800` times. Which raises the question, can we cache some of these get values? For a quick test, I modified the function to build a local cache, and reduced it to `~3s`. 
2. `schema_obj.py:71(__init__)` - takes about `>91.7s` and called `121` times. Would it be possible to reduce call number to this? 

Here are the cProfile reports:
[reset_flow_nodes.txt](https://github.com/user-attachments/files/20034743/reset_flow_nodes.txt)
[local_process.txt](https://github.com/user-attachments/files/20034744/local_process.txt)

Ciprian167:
The thing I tried in `_get_flowgraph_node_outputs` is:

```python
# >>>
_cached_flowgraph_node_outputs = {}
def _get_flowgraph_node_outputs(chip, flow, node):
    # >>>
    key = (id(chip), flow, node)
    if key in _cached_flowgraph_node_outputs:
        return _cached_flowgraph_node_outputs[key]

    node_outputs = []
    iter_nodes = _get_flowgraph_nodes(chip, flow)
    for iter_node in iter_nodes:
        iter_node_inputs = chip.get('flowgraph', flow, *iter_node, 'input')
        if node in iter_node_inputs:
            node_outputs.append(iter_node)
   # >>>
    _cached_flowgraph_node_outputs[key] = node_outputs

    return node_outputs

```

gadfort:
@Ciprian167 do you think it would be possible to add a version of your setup to `scripts/profile_sc.py`? That will help ensure we have a testcase for profiling this going forward.

[...]

PR #3518, opened 2025-05-07
Large flow profile example

Ciprian167:
Cover parts of #3506.


siliconcompiler/scheduler/__init__.py
```diff
@@ -121,6 +124,8 @@ def run(chip):
     # Merge cfgs from last executed tasks, and write out a final manifest.
     _finalize_run(chip)
 
+    _cache_clear()
+
```
gadfort, line 128: If you move this into _finalize_run, I think that would be sufficient to ensure some of the other callers of `run` will hit this.

gadfort (changes requested):
Looks good to me, just move the call to `_cache_clear` and it should be good.

PR #3560, opened 2025-05-15
[Tools] Update VPR Version

AlexandreSinger:
This updates VPR from the mid-january version (approximately VTR-9) to a May version.

Bringing in recent changes to delay modeling and timing analysis which are useful for the Logik flow.

This VPR commit hash is from May 7th and has passed VPR's regular regression tests and weekly strong tests.


AlexandreSinger:
I have raised a PR on the VTR repository that should resolve this issue:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3046

The problem comes from `install-vpr.sh` which calls VPR's `install_apt_packages.sh` script. This script installs all of the packages VPR thinks the user would need to run VPR in general. A new package was added to this script `clang-format-18` which is only found in Ubuntu 24.04. I have made this an optional package to install since it is only used to optionally format the code. I have added the `install_apt_packages.sh` script to VPR's regression tests to ensure it does not regress in the future.

Once this PR is merged I will update SC to the new version.

gadfort:
vpr build: https://github.com/siliconcompiler/siliconcompiler/actions/runs/15068433068

AlexandreSinger:
After discussing with @petergrossmann21 , we decided to drop support for VTR on Ubuntu20. VTR does not include Ubuntu20 in its CI anymore and is bringing in new C++20 features which will break the Ubuntu20 build in the future.

PR #3570, opened 2025-05-16
[Tools][Tests] Added OpenSTA Tool Driver Test

AlexandreSinger:
The OpenSTA driver did not have a test in the tests directory. Added a basic test which takes the foo design used for the Parmys test and performs timing analysis.

In order to make the test interesting, wrote a simple SDC file for this testcase.


gadfort (approved):
lgtm

Issue #3467, opened 2025-04-25
CLI Dashboard: end of run information


gadfort:
From: https://github.com/siliconcompiler/siliconcompiler/issues/3463

Also, do you think it makes sense that on the last rendering to show the entire error list, outside of the live display. That would allow the user to scroll through all the errors without them being truncated by the live display.

Ciprian167:
@gadfort I've been testing the CLI dashboard on projects that have large number of tests. When there are many tests failing printing everything would get very messy. I think for the `job board` printing the nodes in order would be sufficient. If there is an error, print starting from the error, or print the errors based on the graph order should be enough. Which I think you already do. 

But for the complete report one solution would be to have a run summary in the `builddir`. I think that can be separated from the CLI and can be beneficial for flows that don't use the CLI dash. I think the information is already stored in the manifests, but it would be nice to have it presented in a human redable form. Have like a session summary, with that jobs failed, pointers to the log and so on. Then maybe display a path to this log at the end of the cli dashboard?

Would something like this make sense?

gadfort:
@Ciprian167 sorry, got distracted and missed this. I think it could be interesting to have an end of run report, it might make sense to wait a bit since we are working towards cleaning up the APIs and this would be easier to add afterwards.

gadfort:
@Ciprian167 I've been thinking about this recently. I have a basic thing on this branch if you want to take a look:
https://github.com/siliconcompiler/siliconcompiler/tree/eor-summary

I would be interested to know what you think would be helpful to add / remove.
I would love to find a solution to highlight some metrics on some nodes up top, but thats going to require a little more infrastructure  to support.

Ciprian167:
@gadfort Sorry. I missed this. I'll take a look.

Ciprian167:
It's looking good. Thanks for taking the time to look at this. 

I'm wondering if it makes more sense to the `SiliconCompiler End-Of-Run Summary` in the main log, then dump the metrics in a file. It can even be some kind of CSV format. For large jobs, especially when running in CI, it's useful to immediately see what failed, straight from the runner logs. Then if everything is passing, look at metrics in depth. 

```
=============================================================================================
SiliconCompiler End-of-Run Summary
=============================================================================================

Overall Status : PASSED
Design         : heartbeat
Job Name       : job0
Flow           : dvflow-verilator
Filesets       : testbench.verilator, rtl
[...]

PR #3601, opened 2025-05-22
Improve cli rendering for multijob scenarios

Ciprian167:
The main update here is to keep the rendering loop open, and move to a common live display.

### DIFF
--- a/siliconcompiler/report/dashboard/cli/board.py
+++ b/siliconcompiler/report/dashboard/cli/board.py
@@ -22,6 +22,8 @@ from siliconcompiler import SiliconCompilerError, NodeStatus
 from siliconcompiler.utils.logging import SCColorLoggerFormatter
 from siliconcompiler.flowgraph import RuntimeFlowgraph
 
+import atexit
+
 
 class LogBufferHandler(logging.Handler):
     def __init__(self, sync_queue, n=50, event=None):
@@ -253,7 +255,7 @@ class Board(metaclass=BoardSingleton):
             auto_refresh=True,
         )
 
-        self._registered_runs = 0
+        atexit.register(self._stop_on_exit)
 
         self._active = self._console.is_terminal
         if not self._active:
@@ -288,10 +290,14 @@ class Board(metaclass=BoardSingleton):
 
         self._metrics = ("warnings", "errors")
 
+    def _stop_on_exit(self):
+        self.stop()
+
     def open_dashboard(self):
         """Starts the dashboard rendering thread if it is not already running."""
 
-        self._registered_runs += 1
+        if not self._active:
+            return
 
         if not self.is_running():
             self._update_render_data(None)
@@ -336,7 +342,6 @@ class Board(metaclass=BoardSingleton):
             return
 
         self._update_render_data(chip, complete=True)
-        self.stop()
 
     def stop(self):
         """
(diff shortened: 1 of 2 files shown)


siliconcompiler/report/dashboard/cli/board.py
```diff
@@ -282,14 +291,13 @@ def _init_singleton(self):
     def open_dashboard(self):
         """Starts the dashboard rendering thread if it is not already running."""
 
-        if not self._active:
-            return
```
gadfort, line 286: this test is needed when running in a non-tty terminal. You can test this by setting your environmental variable to TTY_COMPATIBLE=0

siliconcompiler/report/dashboard/cli/board.py
```diff
@@ -337,6 +345,11 @@ def stop(self):
         if not self.is_running():
             return
 
+        if self._registered_runs:
+            self._registered_runs -= 1
+            if self._registered_runs:
+                return
```
gadfort, line 351: does accesses to `_registered_runs` need to be gated by a lock?

gadfort (changes requested):
When I run the gcd.py example with
```

    chip = siliconcompiler.Chip('gcd')
    chip.dashboard(type='cli')
    chip.input(os.path.join(root, "gcd.v"))
    chip.input(os.path.join(root, "gcd.sdc"))
    chip.set('option', 'quiet', True)
    chip.set('option', 'track', True)
    chip.set('option', 'hash', True)
    chip.set('option', 'nodisplay', True)
    chip.set('constraint', 'outline', [(0, 0), (100.13, 100.8)])
    chip.set('constraint', 'corearea', [(10.07, 11.2), (90.25, 91)])
    chip.use(freepdk45_demo)
    chip.run()
    chip._dash.stop()
    chip.summary()
````

The summary is printed before the dashboard (off screen), which seems to indicate that someone is not happening for shutdown

Ciprian167:
> The summary is printed before the dashboard (off screen), which seems to indicate that someone is not happening for shutdown

You're right. I haven't tested that. That's because I removed the `live.stop`.  I was trying for fix a weird scenario where the live is stopped and started many times. That's valid in multijob runs, particularily when the run doesn't have anything new  and they finish very fast. The easiest way to recreate this is to modify the blinky.py and just add multiple `chip.run()` call back to back. That will result in something like:

image

gadfort:
@Ciprian167 maybe we can keep the live "active" until a call to `stop()` happens (we would just need to remove the call to `stop` in `end_of_run` so the display stays on and once we call `stop` we can stop the live display

Ciprian167:
That is esentially what we need to do. But I'm not sure what would be the best way. To cover all cases, we probably need to either let the user call the stop function, or maybe let the dashboard stop itself at exit. Or maybe both. I think this might affect the summary display, but might work with a sequence of `chip.run -> chip.dashboard.stop() -> chip.summary()`. 

How do you think we should handle the stop? Where should it be?

gadfort:
I think `chip.run() -> chip.dashboard.stop() -> chip.summary()` should yield the same behavior as now (ie. dashboard stops and printing of the summary is fine).
If the user does not call `stop()` then I think it's fine to let it run (this leaves the issue of the logger output a TBD) and then exit when the python session ends.

siliconcompiler/report/dashboard/cli/board.py
Ciprian167: `_registered_runs` can be removed if we leave the live display running the entire time.

siliconcompiler/report/dashboard/cli/board.py
Ciprian167: I'll put it back.

Ciprian167:
Seems to be working well. We don't have a dasboard object. The user will have to call `chip._dash.stop`. Do you want me to add a `dash` property to expose the dashboard object?

gadfort:
@Ciprian167 does it make sense to remove the final render and leave it the progressbar instead with the log, calling `.run()` multiple times after the first run finishes the logger output is not displayed anymore, which feels like an issue (we can handle that in a separate PR if you want)

Ciprian167:
> @Ciprian167 does it make sense to remove the final render and leave it the progressbar instead with the log, calling `.run()` multiple times after the first run finishes the logger output is not displayed anymore, which feels like an issue (we can handle that in a separate PR if you want)

Yes. We need to rethink that. I was thinking about a similar thing while I was running stuff. I'll be off grid for a while. But I can do it when I'm back in a separate PR.

gadfort (approved):

gadfort:
FYI https://github.com/siliconcompiler/siliconcompiler/issues/3467

PR #3652, opened 2025-06-02
record microseconds for start and end times to improve reporting times

gadfort:
Adds the clarification of timezone used for record to help


aolofsson (approved):
Makes sense.

PR #3660, opened 2025-06-03
[Tools][VPR] Added Option to Generate Post Implementation Netlist

AlexandreSinger:
Similar to the "max_router_iterations" option in the VTR flow, created an option that can be passed into the VPR tool to have the route stage generate a post-implementation netlist. This option will put the generated files into the output directory in a form that can be accepted into the OpenSTA tool.


siliconcompiler/tools/vpr/route.py
```diff
@@ -122,3 +159,16 @@ def post_process(chip):
...
+    step = chip.get('arg', 'step')
+    index = chip.get('arg', 'index')
+    tool, task = get_tool_task(chip, step, index)
+    gen_post_implementation_netlist = chip.get('tool', tool, 'task', task, 'var',
+                                               'gen_post_implementation_netlist',
+                                               step=step, index=index)[0]
+    if gen_post_implementation_netlist == 'true':
+        shutil.copy2(f'{design}_post_synthesis.v', f'outputs/{design}.vg')
+        shutil.copy2(f'{design}_post_synthesis.sdc', f'outputs/{design}.sdc')
+        shutil.copy2(f'{design}_post_synthesis.sdf', f'outputs/{design}.typical.sdf')
```
gadfort: move this file, no reason to have them lying around

siliconcompiler/tools/vpr/route.py
```diff
@@ -44,6 +44,31 @@ def setup(chip, clobber=True):
...
+    gen_post_implementation_netlist = chip.get('tool', tool, 'task', task, 'var',
+                                               'gen_post_implementation_netlist',
+                                               step=step, index=index)[0]
+    if gen_post_implementation_netlist == 'true':
+        chip.add('tool', tool, 'task', task, 'output', design + '.vg',
+                 step=step, index=index)
+        chip.add('tool', tool, 'task', task, 'output', design + '.sdc',
+                 step=step, index=index)
+        chip.add('tool', tool, 'task', task, 'output', design + '.typical.sdf',
+                 step=step, index=index)
```
gadfort: might be nice to control the corner name with a variable

siliconcompiler/tools/vpr/route.py
```diff
@@ -44,6 +44,31 @@ def setup(chip, clobber=True):
     chip.add('tool', tool, 'task', task, 'output', design + '.net', step=step, index=index)
     chip.add('tool', tool, 'task', task, 'output', design + '.place', step=step, index=index)
 
+    # Add a parameter for generating a post-implementation netlist for external
+    # static timing analysis.
+    chip.set('tool', tool, 'task', task, 'var', 'gen_post_implementation_netlist',
+             'false', step=step, index=index, clobber=False)
```
gadfort: `'false'` -> `False`

tests/flows/test_fpgaflow.py
```diff
@@ -522,3 +522,49 @@ def test_fpga_syn_extract(top_module,
         assert expected_macro_count == 1, \
             f'Expected one instance of {expected_macro},' \
             ' got {expected_macro_count} instances'
+
+
+@pytest.mark.quick
```
gadfort: remove this mark, it only applies to the eda tests

tests/flows/test_fpgaflow.py
```diff
@@ -522,3 +522,49 @@ def test_fpga_syn_extract(top_module,
...
+    fpga.set('fpga', part_name, 'var', 'channelwidth', 50)
+
+    chip.use(fpga)
+    chip.set('fpga', 'partname', 'faux')
+
+    # 3. Load flow
+    chip.use(fpgaflow, fpgaflow_type='vpr')
+    chip.set('option', 'flow', 'fpgaflow')
+
+    chip.set('tool', 'vpr', 'task', 'route', 'var', 'gen_post_implementation_netlist', 'true')
```
gadfort: `'true'` -> `True`
while the value is stored as a string, that will be changed soon-ish so might as well get the setting side of things right

tests/flows/test_fpgaflow.py
```diff
@@ -522,3 +522,49 @@ def test_fpga_syn_extract(top_module,
...
+        for step, index in layer_nodes:
+            SchedulerNode(chip, step, index).setup()
+
+    assert 'true' == \
+        chip.get('tool', 'vpr', 'task', 'route', 'var', 'gen_post_implementation_netlist',
+                 step='route', index='0')[0]
+
+    chip.set('arg', 'step', 'route')
+    chip.set('arg', 'index', '0')
+    assert '--gen_post_synthesis_netlist' in route.runtime_options(chip)
```
gadfort: See if you can use
```
node = SchedulerNode(chip, step, index)
node.init_state(assign_runtime=True)
assert ... in node.task.get_runtime_arguments()
```

tests/flows/test_fpgaflow.py
```diff
@@ -522,3 +522,49 @@ def test_fpga_syn_extract(top_module,
         assert expected_macro_count == 1, \
             f'Expected one instance of {expected_macro},' \
             ' got {expected_macro_count} instances'
+
+
+@pytest.mark.quick
+def test_vpr_gen_post_implementation_netlist():
```
gadfort, line 527: probably need a test for this behavior too with VPR

AlexandreSinger: I agree. I tried adding one, but the FPGA architectures provided in the demo folder of logik lib do not have delay annotations as far as I can tell: https://github.com/siliconcompiler/logiklib/tree/main/logiklib/demo

We could use the Z1000 architecture, but I thought it may be a bit large for a CI test. I could add a CI test using this architecture if you think it would be ok.

VPR requires at least delay annotations of the setup and hold constraints of flip flops in order to perform timing analysis, which is required to generate the timing annotations of the post implementation netlist. I also noticed that we do not have a test that provides an SDC file to the VPR tool (which turns on timing analysis). I think if we can add a small architecture to logiklib with delay annotations we can add these tests.

tests/flows/test_fpgaflow.py
AlexandreSinger: That worked! I also prefer this over what I wrote prior.

tests/flows/test_fpgaflow.py
gadfort: The Z1000 is fairly small, if the design running on it is small enough it should be fast enough to run it (either as a `quick` or not)

gadfort: @AlexandreSinger lets merge this, and if you can add a timing test for VPR in a different PR that would be great.

PR #3681, opened 2025-06-06
Implementing a new SC Design class

aolofsson:
This PR implements a new SC Design class that extends the work on the scalable and object oriented approach that @gadfort has already started with the new schema. This work is not yet deployed, but will eventually get integrated in the full blown SC.

**Context:**
1. When SC started in 2021, I had a certain working methodology in mind (flists, -y etc) that had worked for me. SC started out with the simple goal of enabling rtl2ds flows but has grown up to be a a general purpose flow manager. We have outgrown our initial architecture.
2. A general purpose flow manager such as the current SC needs more capture flexibilty (more like make, cmake, fusesoc, bender,...). We have chosen to lean into the 'fileset' concept used by tools like Vivado and fusesoc.
3. This new design class raises the level of abstraction for design/ip reuse, by capturing designs via design objects rather than via hard coded setup() functions attached to chip objects.
4. The manifest/schema concept is maintained, but accessing is now handled via design specific setter/getters for ease of use/error checking
5. The new design object will supports reading and writing of different design setup formats (flist ...) (wip). 
6. Design dependency hiearchy is supported (wip)

**Notes:**
-We are abandoning support for -v,  -y, since these are not generally supported across the EDA ecosystema and was causing implementation issues.


siliconcompiler/design.py
```diff
@@ -0,0 +1,535 @@
...
+        Args:
+           fileset (str): Fileset name.
+           value (str): Topmodule name.
+
+        Returns:
+           str: Topmodule name
+
+        Notes:
+        - first character must be letter or underscore
+        - remaining characters can be letters, digits, or underscores
```
gadfort, line 33: why particular reason for this restriction?

siliconcompiler/design.py
```diff
@@ -0,0 +1,535 @@
...
+
+        """
+
+        # topmodule safety check
+        if (value is not None) and isinstance(value, str):
+            if not re.match(r'^[_a-zA-Z]\w*$', value):
+                raise ValueError(f"{value} is not a legal topmodule string")
+
```
gadfort, line 41: needs an else to generate an error if the value is None or not a str:
```
else:
 raise ValueError("....")
```

siliconcompiler/design.py
```diff
@@ -0,0 +1,535 @@
...
+        return self._set(fileset, 'idir', value, typelist=[str, list])
+
+    def get_idir(self, fileset: str) -> List[str]:
+        """Returns include directories for a fileset.
+
+        Args:
+           fileset (str): Fileset name.
+
+        Returns:
+           list[str]: Topmodule name
```
gadfort: include directories

siliconcompiler/design.py
```diff
@@ -0,0 +1,535 @@
...
+            fileset (str): Fileset name.
+            name (str): Parameter name.
+            value (str): Parameter value.
+        """
+
+        if not isinstance(fileset, str):
+            raise TypeError("fileset key must be a string")
+
+        if not isinstance(value, str) or value is None:
+            raise TypeError("param value must be a string")
```
gadfort: this wont hurt, but would be caught by the schemas key error checking.

siliconcompiler/design.py
```diff
@@ -0,0 +1,535 @@
...
+
+        Args:
+           name (str): Parameter name.
+           fileset (str): Fileset name.
+
+        Returns:
+            str: Parameter value
+        """
+        if not isinstance(fileset, str):
+            raise RuntimeError("fileset value must be a string")
```
gadfort: This error needs to be consistent (it's  TypeError above and ValueError below), ValueError is probably correct for the fileset

siliconcompiler/design.py
```diff
@@ -0,0 +1,535 @@
...
+        return self.get('fileset', fileset, 'param', name)
+
+    ###############################################
+    def add_file(self, fileset: str, filename: str, filetype=None, package=None):
```
gadfort: package isn't handled in this function.

siliconcompiler/design.py
```diff
@@ -0,0 +1,535 @@
...
+        .lef      → (input, lef)
+        .def      → (input, def)
+        ...       → etc.
+
+        Args:
+            fileset (str): Logical group to associate the file with.
+            filename (Path or list[Path]): File path or list of paths to add.
+            filetype (str, optional): Type of the file (e.g., 'verilog', 'sdc').
+                If not provided, it is inferred from the file extension.
+
```
gadfort, line 276: package missing from args

siliconcompiler/design.py
```diff
@@ -0,0 +1,535 @@
...
+        if filename is None:
+            raise ValueError("read_fileset() filename cannot be None")
+
+        if not fileformat:
+            formats = {}
+            formats['f'] = 'flist'
+            fileformat = formats[Path(filename).suffix.strip('.')]
+
+        if fileformat == "flist":
+            raise RuntimeError("read_fileset is not implemented yet")
```
gadfort: no change needed, but `NotImplementedError` is the more appropriate error type here.

siliconcompiler/design.py
```diff
@@ -0,0 +1,535 @@
...
+                legalval = True
+        if not legalval:
+            raise ValueError("value type must be str or List")
+
+        # None is illegal for all setters
+        if value is None:
+            raise ValueError(f"None is an illegal {option} value")
+
+        self.set('fileset', fileset, option, value)
+        return self.get('fileset', fileset, option)
```
gadfort: any reason to not just return the result of self.set() directly?

siliconcompiler/design.py
gadfort: do we want a way to clobber the files? ie. reset the list, which calls set instead of add?

siliconcompiler/design.py
```diff
@@ -0,0 +1,535 @@
...
+
+        if fileformat == "flist":
+            raise RuntimeError("read_fileset is not implemented yet")
+        else:
+            raise ValueError(f"{fileformat} is not supported")
+
+    ################################################
+    # Helper Functions
+    ################################################
+    def _set(self, fileset, option, value, typelist=None):
```
gadfort: this should be `__set` to avoid exposing this to users (this is purely an internal function)

siliconcompiler/design.py
```diff
@@ -0,0 +1,535 @@
...
+            raise ValueError("value type must be str or List")
+
+        # None is illegal for all setters
+        if value is None:
+            raise ValueError(f"None is an illegal {option} value")
+
+        self.set('fileset', fileset, option, value)
+        return self.get('fileset', fileset, option)
+
+    def _get(self, fileset, option):
```
gadfort: this should be `__get`

siliconcompiler/design.py
```diff
@@ -0,0 +1,535 @@
...
+
+
+###########################################################################
+class DesignSchema(NamedSchema):
+
+    def __init__(self, name: str):
+        NamedSchema.__init__(self, name=name)
+        schema_design(self)
+        # TODO: move this into use schema?
+        self.__dependency = {}
```
gadfort: I think you can safely remove this since we'll implement this separately and inherit it

gadfort (changes requested):
whats the reason for the change in the arguments list?
I was thinking about how to handle the "default" fileset and one possibility would be a context manager to handle it, like:
```
with design.with_fileset("rtl"):
    design.add_file("top.v")
```
This avoids the state information since the context manager would reset the default fileset when we exit the `with`. However, this would not be all that practical to add with the current function arguments, since the fileset is a positional argument instead of a keyword argument.

siliconcompiler/design.py
aolofsson: This is the verilog restriction, if we don't want to filter we can remove. Seems like a reasonable limitation.

siliconcompiler/design.py
aolofsson: What would be the use case for clobber? A design should be self contained. Overriding by a different clase or downstram in the tool driver seems like a dangerous idea?

siliconcompiler/design.py
aolofsson: I forgot that you return the value, will do

aolofsson:
> whats the reason for the change in the arguments list? I was thinking about how to handle the "default" fileset and one possibility would be a context manager to handle it, like:
> 
> ```
> with design.with_fileset("rtl"):
>     design.add_file("top.v")
> ```
> 
> This avoids the state information since the context manager would reset the default fileset when we exit the `with`. However, this would not be all that practical to add with the current function arguments, since the fileset is a positional argument instead of a keyword argument.

Are you suggesting hard coding filesets like 'rtl' inside the class?

The reason for the change was to avoid confusion around order. I made a mistake a couple of times, could be just me or it's a sign that it's more non intuitive to have the fileset as an option in the set but not in the get.

set_topmodule(fileset, value)
get_topmodule(fileset)

vs

set_topmodule(value, fileset)
get_topmodule(fileset)

siliconcompiler/design.py
aolofsson: handled in __set() for all setters

siliconcompiler/design.py
aolofsson: done

siliconcompiler/design.py
aolofsson: Don't we have to wait intil the use schema is complete to implement?

siliconcompiler/design.py
aolofsson: done

siliconcompiler/design.py
aolofsson: done

siliconcompiler/design.py
aolofsson: done

siliconcompiler/design.py
aolofsson: done

siliconcompiler/design.py
aolofsson: done

siliconcompiler/design.py
aolofsson: so in that case, it should be removed right?

siliconcompiler/design.py
aolofsson: so value error for all?

siliconcompiler/design.py
gadfort: yes, this can be removed

siliconcompiler/design.py
gadfort: yes

siliconcompiler/design.py
gadfort: I don't think a downstream tool would be using it, but if a user is working on this it might be helpful to be able to "undo" that they did if it was wrong (more of an interactive use case, I would think clobber should be false by default).

siliconcompiler/design.py
aolofsson: Seems like None is not being flagged by the underluing schema. Example

```python
import siliconcompiler
chip = siliconcompiler.Chip("test")
chip.set('option', 'param', 'N', None)
print(chip.get('option', 'param', 'N'))
```

gadfort: my mistake, I thought it was the "N" you were checking, not the value, the value needs this check, so nevermind.

siliconcompiler/design.py
gadfort: is there any concern this will not be applicable to HLS and any other languages?

gadfort:
> > whats the reason for the change in the arguments list? I was thinking about how to handle the "default" fileset and one possibility would be a context manager to handle it, like:
> > ```
> > with design.with_fileset("rtl"):
> >     design.add_file("top.v")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This avoids the state information since the context manager would reset the default fileset when we exit the `with`. However, this would not be all that practical to add with the current function arguments, since the fileset is a positional argument instead of a keyword argument.
> 
> Are you suggesting hard coding filesets like 'rtl' inside the class?
> 
> The reason for the change was to avoid confusion around order. I made a mistake a couple of times, could be just me or it's a sign that it's more non intuitive to have the fileset as an option in the set but not in the get.
> 
> set_topmodule(fileset, value) get_topmodule(fileset)
> 
> vs
> 
> set_topmodule(value, fileset) get_topmodule(fileset)

I'm not suggesting that any specific fileset get hardcoded, just that we could implement a context manager (https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) so we can temporarily set the fileset within a given context.
```
 with design.with_fileset("rtl"):
     design.add_file("top.v")
     design.add_file("impl.v")
     design.set_top("top")
 design.set_top("hlstop", fileset="hls")
 with design.with_fileset("testbench.0"):
     design.add_file("tb0.v")
     design.set_top("tb")
 with design.with_fileset("testbench.1"):
     design.add_file("tb1.v")
     design.set_top("tb")
```

I think it would be helpful to be able to implement something like this

siliconcompiler/design.py
aolofsson: Haskell seems to be outlier, what a surprise:-). Not sure it's credible to have an apostrophe in a top level function name, even in Haskell?

**Python:**
Valid Characters: Function names may contain letters (a-z, A-Z), digits (0-9), and underscores (_). Start with a Letter or Underscore: Function names must begin with a letter (a-z, A-Z) or an underscore (_).

**C++:**
Valid C++ Identifiers Must:
Start with a letter (A–Z, a–z) or an underscore _
Cannot start with a digit.
Only contain letters, digits (0–9), and underscores
No spaces, punctuation, or special characters like !, @, #, etc.

**VHDL**
Valid VHDL Identifiers Must:
Start with a letter (A–Z, a–z)
Contain only letters, digits (0–9), and underscores (_)
Not end with an underscore
Not contain two or more underscores in a row

**Haskell**
Must start with a lowercase letter: a–z
May contain:
Letters: a–z, A–Z
Digits: 0–9 (but not at the beginning)
Underscores _
Apostrophes '

**Scala**
Must start with a letter (a–z or A–Z) or an underscore _
Can contain:
Letters
Digits (0–9)
Underscores _

gadfort: I don't know of any users using haskell, so seems safe to keep as is

aolofsson:
> > > whats the reason for the change in the arguments list? I was thinking about how to handle the "default" fileset and one possibility would be a context manager to handle it, like:
> > > ```
> > > with design.with_fileset("rtl"):
> > >     design.add_file("top.v")
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This avoids the state information since the context manager would reset the default fileset when we exit the `with`. However, this would not be all that practical to add with the current function arguments, since the fileset is a positional argument instead of a keyword argument.
> > 
> > 
> > Are you suggesting hard coding filesets like 'rtl' inside the class?
> > The reason for the change was to avoid confusion around order. I made a mistake a couple of times, could be just me or it's a sign that it's more non intuitive to have the fileset as an option in the set but not in the get.
> > set_topmodule(fileset, value) get_topmodule(fileset)
> > vs
> > set_topmodule(value, fileset) get_topmodule(fileset)
> 
> I'm not suggesting that any specific fileset get hardcoded, just that we could implement a context manager (https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) so we can temporarily set the fileset within a given context.
> 
> ```
>  with design.with_fileset("rtl"):
>      design.add_file("top.v")
>      design.add_file("impl.v")
>      design.set_top("top")
>  design.set_top("hlstop", fileset="hls")
>  with design.with_fileset("testbench.0"):
>      design.add_file("tb0.v")
>      design.set_top("tb")
>  with design.with_fileset("testbench.1"):
>      design.add_file("tb1.v")
>      design.set_top("tb")
> ```
> 
> I think it would be helpful to be able to implement something like this

Nice solutation. Confirmed! I will change the  argument order on set/add to make fileset optional and second argument.

siliconcompiler/design.py
aolofsson: sold on adding clobber.

aolofsson: done

PR #3682, opened 2025-06-06
add use schema class to augment child classes

gadfort:
Adds `.use` to classes that implement this along with accessors for the dependency data:
* `.get_dependency()`
* `.get_all_dependencies()`
* `.write_dependencygraph()`
* `.remove_dependency()`

+ callback to NamedSchema with `._reset()` to enable removing state information

### DIFF
--- a/docs/reference_manual/schema_api.rst
+++ b/docs/reference_manual/schema_api.rst
@@ -104,6 +104,9 @@ Full API
     :show-inheritance:
     :inherited-members:
 
+.. autoclass:: UseSchema
+    :members:
+
 .. autoclass:: PackageSchema
     :members:
     :show-inheritance:
--- a/siliconcompiler/schema/namedschema.py
+++ b/siliconcompiler/schema/namedschema.py
@@ -26,6 +26,12 @@ class NamedSchema(BaseSchema):
         '''
         return self.__name
 
+    def _reset(self) -> None:
+        """
+        Resets the state of the object
+        """
+        pass
+
     @classmethod
     def from_manifest(cls, name, filepath=None, cfg=None):
         '''
--- a/siliconcompiler/schema/useschema.py
+++ b/siliconcompiler/schema/useschema.py
@@ -1,0 +1,176 @@
+import os.path
+
+from .namedschema import NamedSchema
+
+
+class UseSchema:
+    '''
+    Schema extention to add :meth:`.use` capability to a schema section.
+    '''
+
+    def __init__(self):
+        self.__imported = {}
+
+    def use(self, obj: NamedSchema, clobber: bool = True) -> bool:
+        """
+        Adds a module as a dependency.
+
+        Returns:
+            True if object was imported, otherwise false.
+
+        Args:
+            obj (:class:`NamedSchema`): Dependency to add
+        """
+
+        if not isinstance(obj, NamedSchema):
+            raise TypeError(f"Cannot use an object of type: {type(obj)}")
+
+        if not clobber and obj.name() in self.__imported:
+            return False
+
+        self.__imported[obj.name()] = obj
+        obj._reset()
+
+        return True
+
+    def write_dependencygraph(self, filename: str,
+                              fontcolor: str = '#000000',
+                              background: str = 'transparent',
+                              fontsize: str = '14',
+                              border: bool = True, landscape: bool = False) -> None:
+        r'''
+        Renders and saves the dependency graph to a file.
+
+        Args:
+            filename (filepath): Output filepath
+            fontcolor (str): Node font RGB color hex value
+            background (str): Background color
+            fontsize (str): Node text font size
+            border (bool): Enables node border if True
+            landscape (bool): Renders graph in landscape layout if True
+
+        Examples:
+            >>> use.write_dependencygraph('mydump.png')
+            Renders the object dependency graph and writes the result to a png file.
+        '''
+        import graphviz
+
+        filepath = os.path.abspath(filename)
+        fileroot, ext = os.path.splitext(filepath)
+        fileformat = ext[1:]
+
+        # controlling border width
+        if border:
+            penwidth = '1'
+        else:
+            penwidth = '0'
+
+        # controlling graph direction
+        if landscape:
+            rankdir = 'LR'
+        else:
+            rankdir = 'TB'
+
+        dot = graphviz.Digraph(format=fileformat)
+        dot.graph_attr['rankdir'] = rankdir
+        dot.attr(bgcolor=background)
+
+        def make_label(dep):
+            return f"lib-{dep.name()}"
+
+        nodes = {
+            self.name(): {
+                "text": self.name(),
+                "shape": "box",
+                "color": background,
+                "connects_to": set([make_label(subdep) for subdep in self.__imported.values()])
+            }
+        }
+        for dep in self.get_all_dependencies():
+            nodes[make_label(dep)] = {
+                "text": dep.name(),
+                "shape": "oval",
+                "color": background,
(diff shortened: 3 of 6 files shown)


siliconcompiler/schema/useschema.py
```diff
@@ -0,0 +1,212 @@
...
+            'used',
+            Parameter(
+                '[str]',
+                scope=Scope.GLOBAL,
+                shorthelp="List of used dependencies",
+                help="List of used modules this design depends on."))
+
+        self.__used = {}
+
+    def use(self, obj: NamedSchema, clobber: bool = True) -> bool:
```
aolofsson: add help for clobber arg

siliconcompiler/schema/useschema.py
```diff
@@ -0,0 +1,212 @@
...
+        super().__init__()
+
+        schema = EditableSchema(self)
+
+        schema.insert(
+            'used',
+            Parameter(
+                '[str]',
+                scope=Scope.GLOBAL,
+                shorthelp="List of used dependencies",
```
aolofsson: does module, mean python module in this help context.

siliconcompiler/schema/useschema.py
```diff
@@ -0,0 +1,212 @@
...
+        self.__used = {}
+
+    def use(self, obj: NamedSchema, clobber: bool = True) -> bool:
+        """
+        Adds a module to this object.
+
+        Returns:
+            True if object was imported, otherwise false.
+
+        Args:
```
aolofsson: args usually go before returns in docstring, right?

siliconcompiler/schema/useschema.py
aolofsson: use case for clobber?

siliconcompiler/schema/namedschema.py
```diff
@@ -26,6 +26,12 @@ def name(self):
         '''
         return self.__name
 
+    def _reset(self) -> None:
+        """
+        Resets the state of the object
+        """
```
aolofsson, line 32: what is this a place holder for?

siliconcompiler/schema/useschema.py
```diff
@@ -0,0 +1,212 @@
...
+                # nothing to iterate over
+                continue
+
+            for obj_dep in obj.__get_all_used(seen):
+                deps.append(obj_dep)
+                seen.add(obj_dep.name())
+
+        return deps
+
+    def get_all_used(self, hierarchy: bool = True) -> list:
```
aolofsson: pros and cons of having multiple functions here get_use + get_all_used vs get_use with arguments?

siliconcompiler/schema/useschema.py
```diff
@@ -0,0 +1,212 @@
...
+        if name not in self.__used:
+            raise KeyError(f"{name} is not an imported module")
+
+        return self.__used[name]
+
+    def __get_all_used(self, seen: set) -> list:
+        '''
+        Loop through the use tree and generate a flat list
+        '''
+        deps = []
```
aolofsson: Some languages use deps as the stanadard abbrenvation for dependencies. I think get_deps() would be pretty clear. Given that use() and get_used() aren't really an obvious set/get pair it might be something to consider.

siliconcompiler/schema/useschema.py
```diff
@@ -0,0 +1,212 @@
...
+        Args:
+            hierarchy (bool): if True, will return all modules including children
+                otherwise only this objects modules are returned.
+        '''
+
+        if hierarchy:
+            return self.__get_all_used(set())
+        return list(self.__used.values())
+
+    def remove_use(self, name: str) -> bool:
```
aolofsson: use case for remove_use()

siliconcompiler/schema/useschema.py
```diff
@@ -0,0 +1,212 @@
...
+    Schema extension to add :meth:`.use` capability to a schema section.
+    '''
+
+    def __init__(self):
+        super().__init__()
+
+        schema = EditableSchema(self)
+
+        schema.insert(
+            'used',
```
aolofsson: consider "deps"

siliconcompiler/schema/useschema.py
```diff
@@ -0,0 +1,212 @@
...
+        if self.__used:
+            return
+
+        for module in self.get("used"):
+            if module not in module_map:
+                raise ValueError(f"{module} not available in map")
+            self.__used[module] = module_map[module]
+
+            if isinstance(self.__used[module], UseSchema):
+                self.__used[module]._populate_used(module_map)
```
aolofsson: where does the register_source/package_root belong? In BaseSchema?

siliconcompiler/schema/useschema.py
gadfort: it's the name of the `NamedSchema`.

siliconcompiler/schema/useschema.py
gadfort: makes sense

siliconcompiler/schema/namedschema.py
gadfort: this was as a callback that can be implemented by the child classes if they have state information that needs to be cleared during import

siliconcompiler/schema/useschema.py
gadfort: they could be merged, with some additional error checking.

siliconcompiler/schema/useschema.py
gadfort: I started with something like that, but ultimately felt that keeping the "use" word was helpful to indicate that these two functions are related.

siliconcompiler/schema/useschema.py
gadfort: during swap_libraries we are likely going to need to remove and replace specific imports (lambdalib -> sky130_lambdalib) for example. This provides a method to do so.

siliconcompiler/schema/useschema.py
gadfort: I'm working on that on a separate branch at the moment, it will not be in this PR.

siliconcompiler/schema/useschema.py
gadfort: probably not needed if `remove_used` is available.

siliconcompiler/schema/useschema.py
aolofsson: Would be good to clarify help. The word module could mean a verilog design module or Python module. 

"List of used modules this design depends on." --> "List of named object dependencies included via use()."

Also, the UseSchema is general purpose right, not jsut for design objects?

gadfort: that is correct, it's not design specific

docs/reference_manual/schema_api.rst
```diff
@@ -104,6 +104,9 @@ Full API
     :show-inheritance:
     :inherited-members:
 
+.. autoclass:: UseSchema
```
aolofsson: outdated reference

siliconcompiler/dependencyschema.py
```diff
@@ -0,0 +1,220 @@
...
+    Schema extension to add :meth:`.add_dep` capability to a schema section.
+    '''
+
+    def __init__(self):
+        super().__init__()
+
+        schema = EditableSchema(self)
+
+        schema.insert(
+            'dependencies',
```
aolofsson: I think we should use the short form 'deps' here in the schema parameter. We use abbrevs in many places and it would be consistent with everything in this module.

aolofsson (approved):
LGTM!

PR #3696, opened 2025-06-12
add find_files and find_source to dependencyschema

gadfort:
Add:
- `file_files` and `find_source` to `DependencySchema`
- `register_source` and associated parameters

`find_files` is a wrapper around the BaseSchema `find_files` which handles adding the package source information for proper path resolution.

### DIFF
--- a/siliconcompiler/dependencyschema.py
+++ b/siliconcompiler/dependencyschema.py
@@ -1,9 +1,12 @@
 import os.path
 
-from .schema.baseschema import BaseSchema
-from .schema.editableschema import EditableSchema
-from .schema.parameter import Parameter, Scope
-from .schema.namedschema import NamedSchema
+from siliconcompiler.schema.baseschema import BaseSchema
+from siliconcompiler.schema.editableschema import EditableSchema
+from siliconcompiler.schema.parameter import Parameter, Scope
+from siliconcompiler.schema.namedschema import NamedSchema
+from siliconcompiler.schema.utils import trim
+
+from siliconcompiler.package import Resolver
 
 
 class DependencySchema(BaseSchema):
@@ -14,8 +17,9 @@ class DependencySchema(BaseSchema):
     def __init__(self):
         super().__init__()
 
-        schema = EditableSchema(self)
+        self.__deps = {}
 
+        schema = EditableSchema(self)
         schema.insert(
             "deps",
             Parameter(
@@ -25,7 +29,38 @@ class DependencySchema(BaseSchema):
                 shorthelp="List of dependencies",
                 help="List of named object dependencies included via add_dep()."))
 
-        self.__deps = {}
+        schema.insert(
+            'source', 'default', 'path',
+            Parameter(
+                'str',
+                scope=Scope.GLOBAL,
+                shorthelp="Package: data source path",
+                example=[
+                    "api: chip.set('source', "
+                    "'freepdk45_data', 'path', 'ssh://git@github.com/siliconcompiler/freepdk45/')"],
+                help=trim("""
+                Package data source path, allowed paths:
+
+                * /path/on/network/drive
+                * file:///path/on/network/drive
+                * git+https://github.com/xyz/xyz
+                * git://github.com/xyz/xyz
+                * git+ssh://github.com/xyz/xyz
+                * ssh://github.com/xyz/xyz
+                * https://github.com/xyz/xyz/archive
+                * https://zeroasic.com/xyz.tar.gz
+                * python://siliconcompiler
+                """)))
+
+        schema.insert(
+            'source', 'default', 'ref',
+            Parameter(
+                'str',
+                scope=Scope.GLOBAL,
+                shorthelp="Package: data source reference",
+                example=[
+                    "api: chip.set('source', 'freepdk45_data', 'ref', '07ec4aa')"],
+                help="Package data source reference"))
 
     def _from_dict(self, manifest, keypath, version=None):
         self.set("deps", False, field="lock")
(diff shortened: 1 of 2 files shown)


siliconcompiler/dependencyschema.py
```diff
@@ -25,7 +29,38 @@ def __init__(self):
                 shorthelp="List of dependencies",
                 help="List of named object dependencies included via add_dep()."))
 
-        self.__deps = {}
+        schema.insert(
+            'source', 'default', 'path',
```
aolofsson: I would rename 'package', 'default', 'root'  OR  'package', 'default', 'home'

Help "Package data source path" shows that package would be appropriate. Not all things pointed to will truly be source so 'source'. This is a packagaging action, so packageis more intuitive imo.

siliconcompiler/dependencyschema.py
```diff
@@ -25,7 +29,38 @@ def __init__(self):
...
+                * git+https://github.com/xyz/xyz
+                * git://github.com/xyz/xyz
+                * git+ssh://github.com/xyz/xyz
+                * ssh://github.com/xyz/xyz
+                * https://github.com/xyz/xyz/archive
+                * https://zeroasic.com/xyz.tar.gz
+                * python://siliconcompiler
+                """)))
+
+        schema.insert(
```
aolofsson, line 58: Not clear from keys or example what this does. Looking at the example, this looks like a commit hash? 'ref' is a very anonymous word.

How about 'tag'? With an expanded help explanation that the meaning is context dependent on the package root and could refer to commit hash, tag, label, release, or version market,..

siliconcompiler/dependencyschema.py
```diff
@@ -218,3 +253,137 @@ def _populate_deps(self, module_map: dict):
 
             if isinstance(self.__deps[module], DependencySchema):
                 self.__deps[module]._populate_deps(module_map)
+
+    def register_source(self, name: str, path: str, ref: str = None):
```
aolofsson: Recommend renaming 'package_root' or 'package_home'

siliconcompiler/dependencyschema.py
```diff
@@ -218,3 +253,137 @@ def _populate_deps(self, module_map: dict):
...
+        """
+        Registers a package by its name with the source path and reference
+
+        Args:
+            name (str): Package name
+            path (str): Path to the sources, can be file, git url, archive url
```
aolofsson: path (str): Path to the package root, can be file, git url, archive url

aolofsson: btw, can it actually point to a single file, or is this a filepath/dir?

siliconcompiler/dependencyschema.py
```diff
@@ -218,3 +253,137 @@ def _populate_deps(self, module_map: dict):
...
+        """
+
+        if os.path.isfile(path):
+            path = os.path.dirname(os.path.abspath(path))
+
+        self.set("source", name, "path", path)
+        if ref:
+            self.set("source", name, "ref", ref)
+
+    def find_source(self, name: str, runnable=None):
```
aolofsson: Rename to 'find_root'

siliconcompiler/dependencyschema.py
```diff
@@ -1,9 +1,12 @@
 import os.path
 
-from .schema.baseschema import BaseSchema
-from .schema.editableschema import EditableSchema
-from .schema.parameter import Parameter, Scope
-from .schema.namedschema import NamedSchema
+from siliconcompiler.schema.baseschema import BaseSchema
+from siliconcompiler.schema.editableschema import EditableSchema
+from siliconcompiler.schema.parameter import Parameter, Scope
```
aolofsson, line 5: Is Dependency the right place for package schema information? Seems like it might be.  What is the up to date diagram showing which schema go in which class?

aolofsson (changes requested):

siliconcompiler/dependencyschema.py
gadfort: A file cannot be a "root" so that is wrong in the help, it's a directory, or ...

siliconcompiler/dependencyschema.py
gadfort: I think root is better

siliconcompiler/dependencyschema.py
gadfort: tag works

siliconcompiler/dependencyschema.py
gadfort: how about `find_package`?

aolofsson: works for me

siliconcompiler/dependencyschema.py
```diff
@@ -218,3 +261,134 @@ def _populate_deps(self, module_map: dict):
 
             if isinstance(self.__deps[module], DependencySchema):
                 self.__deps[module]._populate_deps(module_map)
+
+    def register_package(self, name: str, root: str, tag: str = None):
```
aolofsson, line 265: I like it!

aolofsson (approved):
Still not 100% sure it's a good idea for the lower level details like step and index to bleed through to this level for the find/check methods instead of leaving them in the baseschema, but I don't have an alternative to propose.

gadfort:
> Still not 100% sure it's a good idea for the lower level details like step and index to bleed through to this level for the find/check methods instead of leaving them in the baseschema, but I don't have an alternative to propose.

I think the solution will be to remove the step index access once a little more of the runnable/project class is ready and we can do it without breaking the tool drivers.

PR #3697, opened 2025-06-12
add `with_fileset` to design

gadfort:
Adds:
```
with design.with_fileset("something"):
    ...
```
context manager to allow setting a default fielset within the context of the `with`.

Notes:
- it was added to both setters and getters. Seemed like parity for the getters provided a nice parity to the setters.

### DIFF
--- a/siliconcompiler/design.py
+++ b/siliconcompiler/design.py
@@ -512,7 +512,7 @@ class DesignSchema(NamedSchema, DependencySchema):
         return self.get('fileset', fileset, option)
 
     @contextlib.contextmanager
-    def with_fileset(self, fileset: str):
+    def active_fileset(self, fileset: str):
         """
         Use this context to temporarily set a design fileset.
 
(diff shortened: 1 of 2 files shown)


aolofsson (changes requested):
Not sure about the with ...with pattern?

with d.with_fileset("rtl"):

How about?
with d.active_fileset("rtl")

gadfort:
> Not sure about the with ...with pattern?
> 
> with d.with_fileset("rtl"):
> 
> How about? with d.active_fileset("rtl")

I agree the `with`-`with` is a little awkward.
It would be great to pick a taxonomy that works beyond fileset, the next obvious `with` would be package, so would we be happy with `active_package`?

aolofsson:
> > Not sure about the with ...with pattern?
> > with d.with_fileset("rtl"):
> > How about? with d.active_fileset("rtl")
> 
> I agree the `with`-`with` is a little awkward. It would be great to pick a taxonomy that works beyond fileset, the next obvious `with` would be package, so would we be happy with `active_package`?

I think active_package, active_fileset, active_? works

aolofsson (approved):
LGTM

PR #3710, opened 2025-06-16
implement read_flist

gadfort:
Adds support for readflist (carried over from core.py)  and make write_flist use find_files to generate full filepaths

No api changes, except for the change in write_flist to full paths.


aolofsson (approved):
lgtm

PR #3726, opened 2025-06-20
[Scheduler] Do Not Forward non-Required Outputs

AlexandreSinger:
When a scheduler node is skipped, it currently forwards all of its inputs to the output directory. If the tool was not expecting these files in the output directory, this can raise an issue.

To fix this issue, when a node is being skipped, we now only forward the input files which were required to be output files.


gadfort (approved):
Lgtm, will wait to merge until this run finishes: https://github.com/siliconcompiler/siliconcompiler/actions/runs/15788275318/job/44509277831

PR #3724, opened 2025-06-20
Separate step index

Ciprian167:
Targeting #3713 

I divided all step and index pairs found by the following regex:
- `logger.*step\}\{.*index`
- `f'.*step\}\{.*index`
- `f".*step\}\{.*index`
- .*step.*\}\{

PR #3731, opened 2025-06-23
[Tools][VPR] Post-Implementation Netlist File Generation

AlexandreSinger:
The VPR tool driver decides whether to use timing analysis or not based on if the user provides an SDC file or not. This is a problem for adding an OpenSTA timing node after routing since a post-implementation netlist will not be generated if timing is off (timing information is currently needed to generate this file).

Resolved this issue by having the VPR tool driver always generate the post-implementation netlist files, but just makes them blank if timing analysis is off. The OpenSTA tool driver can then skip itself if it recognizes that blank files were passed into it. This allows the timing analysis tool to be integrated cleanly into the FPGA flow.


siliconcompiler/tools/vpr/vpr.py
```diff
@@ -311,6 +312,13 @@ def auto_constraints():
     return 'inputs/sc_constraints.xml'
 
 
+def use_timing_analysis(chip):
+    '''
+    Return true if the given chip should use timing analysis in the VPR flow.
+    '''
+    return chip.valid('input', 'constraint', 'sdc')
```
gadfort: this should be `chip.valid(...) and chip.get(..., step=, index=)` (a key may be valid, but still have no values set)

siliconcompiler/tools/opensta/timing.py
```diff
@@ -84,6 +85,28 @@ def setup(chip):
...
+
+    # If a verilog and sdc file are both provided as input to this tool, but they
+    # are both empty, skip this tool step.
+    design = chip.top()
+    if (f'{design}.vg' in input_provides(chip, step, index)) and \
+            (f'{design}.sdc' in input_provides(chip, step, index)):
+        if os.path.getsize(f'inputs/{design}.vg') == 0 and \
+                os.path.getsize(f'inputs/{design}.sdc') == 0:
+            chip.set('record', 'status', NodeStatus.SKIPPED, step=step, index=index)
+            chip.logger.warning(f'{step}{index} will be skipped since no timing '
```
gadfort: this should be `{step}/{index}` after #3724

AlexandreSinger:
@gadfort Thank you for your feedback. I have resolved the issues and added a testcase to ensure that the OpenSTA tool is skipped if the inputs are blank as expected.

I was running into issues with the strict mode for the pytest suite since it was catching bugs that I did not introduce. It was an easy fix, but just took some debugging (input files in the OpenSTA tool driver used set instead of add which was clobbering the input files in strict mode but not in regular mode). Should be good now.

PR #3734, opened 2025-06-24
Fixing dependency api examples

aolofsson:


siliconcompiler/dependencyschema.py
```diff
@@ -29,18 +29,18 @@ def __init__(self):
                 '[str]',
                 scope=Scope.GLOBAL,
                 lock=True,
-                shorthelp="List of dependencies",
+                shorthelp="List of package dependencies",
```
gadfort, line 32: These are not really package dependencies, I'm not sure that the right phrase is, but it's basically a design, library, flow, pdk, etc that this object is dependent on.

aolofsson: This might be a point of deeper confusion. We do refer to 'package', 'default' 'root', --> "Package root path, this points the location where the package data can be retrieved or accessed". 

With add_dep(), it refers to adding a "module".  In design at least, all designs are derived from the DependencySchema. Wouldn't library, flow, pdk also derive from the DependencySchema and thus be packages?In my head they were the same thing?

gadfort:
Replaced with https://github.com/siliconcompiler/siliconcompiler/pull/3807

PR #3739, opened 2025-06-27
initial synth_fpga support

gadfort:


tdbesson:
@gadfort FYI, 'synth_fpga" is not case sensitive for partnames at command line option -partname so you can write either z1000 or Z1000. I can do the same if it is given through the config file.
If the partname is not there and unknown,  'synth_fpga' will error out nicely by doing a recap of the allowed partnames (and will not fail any more by accessing a wrong path).

PR #3741, opened 2025-06-27
[VPR] Updated the VTR Version Parsing

AlexandreSinger:
VTR updates its versions infrequently enough to make it so the minor versions cannot only be used to check the version of VTR.

Updated the version parsing for VTR to include the number of commits since the last release as another release segment.


AlexandreSinger:
Version in SC right now:
image

image

PR #3764, opened 2025-07-03
[Tools][VTR] Updated the VTR Version on SC

AlexandreSinger:
Updated the vpr, genfasm, and yosys-parmys versions to the most recent version of VTR.

This new version of VTR comes with improvements to post-implementation timing analysis, small build-time improvements, and cleanups of warnings in the LTO build.

### DIFF
--- a/siliconcompiler/toolscripts/_tools.json
+++ b/siliconcompiler/toolscripts/_tools.json
@@ -76,7 +76,7 @@
   },
   "vpr": {
     "git-url": "https://github.com/verilog-to-routing/vtr-verilog-to-routing.git",
-    "git-commit": "c606a21d9ab5db53fb382a7dd79a0703c887108c",
+    "git-commit": "29cc58dd2d242df245b38eac90e6b5d473ecb726",
     "auto-update": false
   },
   "icepack": {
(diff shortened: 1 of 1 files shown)


AlexandreSinger:
Waiting on the tool builds to finish and then I think this is ready to merge. I will message when they complete.

siliconcompiler/toolscripts/rhel9/install-vpr.sh
```diff
@@ -25,9 +25,6 @@ git submodule update --init --recursive
 sudo dnf config-manager --set-enabled devel || true
 ./install_dnf_packages.sh
 
-sudo yum install -y tbb-devel
-sudo dnf config-manager --set-disabled devel || true
```
gadfort, line 29: need to keep the  `--set-disabled` since we enable it above

gadfort (changes requested):

siliconcompiler/toolscripts/rhel9/install-vpr.sh
AlexandreSinger: Ah thank you! I thought it also said enabled. I will add that back. I was looking at this to see if I can fix this issue:
image

I have since found the issue. It is on VTR's side. I will need to push a fix to VTR master and then we can point to that fixed commit.
image

I have recently updated the includes for VTR and it looks like using the "oneapi" prefix is allowed on some systems but not others...

Easy fix. But it blocks this PR.

AlexandreSinger:
VPR CI Build ran successfully:
https://github.com/siliconcompiler/siliconcompiler/actions/runs/16060765762/job/45326017017

It looks like the Daily CI tests are being run on this PR. Once those are merged in, I think this should be good to go!

@gadfort I have added back that line that I have removed. Please let me know if you have any further comments.

gadfort (approved):
Lgtm

PR #3802, opened 2025-07-11
add depfileset to design filesets

gadfort:
Why:
- design filesets need a way to express the filesets from dependencies they rely on
- also adds concept of alias'ing, which mirrors our current `swap_libraries` call which allows us to change lamdbalib libraries to tech specific lambdalibs.

Adds:
- `*_dependency_fileset` to allow modifying the fileset
- `get_fileset_mapping`, handles resolving the filesets and performing aliasing.

Notes:
- `*_dependency_fileset` should probably be changed to `*_dep_fileset` to be consistent

### DIFF
--- a/siliconcompiler/design.py
+++ b/siliconcompiler/design.py
@@ -252,13 +252,13 @@ class DesignSchema(NamedSchema, DependencySchema):
             raise ValueError("fileset key must be a string")
         return self.get('fileset', fileset, 'param', name)
 
-    def add_depfileset(self, dep: Union[NamedSchema, str], dep_fileset: str, fileset: str = None):
+    def add_depfileset(self, dep: Union[NamedSchema, str], depfileset: str, fileset: str = None):
         """
         Record a reference to an imported dependency's fileset.
 
         Args:
-           dep_name (:class:`NamedSchema` or str): Dependency name or object.
-           dep_fileset (str): Dependency fileset
+           dep (:class:`NamedSchema` or str): Dependency name or object.
+           depfileset (str): Dependency fileset
            fileset (str): Fileset name.
 
         """
@@ -276,7 +276,7 @@ class DesignSchema(NamedSchema, DependencySchema):
         else:
             raise TypeError("dep is not a valid type")
 
-        return self.add("fileset", fileset, "depfileset", (dep_name, dep_fileset))
+        return self.add("fileset", fileset, "depfileset", (dep_name, depfileset))
 
     def get_depfileset(self, fileset: str = None):
         """


siliconcompiler/design.py
```diff
@@ -252,6 +252,41 @@ def get_param(self,
             raise ValueError("fileset key must be a string")
         return self.get('fileset', fileset, 'param', name)
 
+    def add_dep_fileset(self, name: str, dep_fileset: str, fileset: str = None):
```
aolofsson: I recommend the following call. Too many references to the word "fileset" in this context.:

```python 
def add_dep(self, name: str, dep: str, fileset: str = None):
```

siliconcompiler/design.py
```diff
@@ -252,6 +252,41 @@ def get_param(self,
...
+
+        """
+        if fileset is None:
+            fileset = self._get_active("fileset")
+
+        if not isinstance(fileset, str):
+            raise ValueError("fileset key must be a string")
+        return self.add("fileset", fileset, "depfileset", (name, dep_fileset))
+
+    def get_dep_fileset(self, fileset: str = None):
```
aolofsson: Recommend:

```python
def get_dep(self, fileset: str = None):
```

siliconcompiler/design.py
```diff
@@ -624,6 +659,55 @@ def active_fileset(self, fileset: str):
         with self.active(fileset=fileset):
             yield
 
+    def get_fileset_mapping(self,
```
aolofsson: Does this need to be a public method?

siliconcompiler/design.py
```diff
@@ -749,3 +833,13 @@ def schema_design(schema):
             data literals. The types of parameters and values supported is tightly
             coupled to tools being used. For example, in Verilog only integer
             literals (64'h4, 2'b0, 4) and strings are supported.""")))
+
+    schema.insert(
+        'fileset', fileset, 'depfileset',
```
aolofsson, line 846: Too many filesets...recommend:
'fileset', fileset, 'dep'

aolofsson (changes requested):

siliconcompiler/design.py
gadfort: we already have a `get_dep`, this mirrors the pattern for all the other getters

siliconcompiler/design.py
gadfort: I don't see a reason to not make it public

siliconcompiler/design.py
aolofsson: Conflicts with dependency schema. How about?

```python
def add_depfileset(self, depname: str, depfileset: str, fileset: str = None):
```

Same style for get_depfilese()

siliconcompiler/design.py
```diff
@@ -624,6 +668,55 @@ def active_fileset(self, fileset: str):
         with self.active(fileset=fileset):
             yield
 
+    def get_fileset_mapping(self,
```
aolofsson: How about shortening the method name to be get_fileset(), this would be consistent  with read_fileset and write_fileset in that they resolve the whole tree.

aolofsson (changes requested):

aolofsson (approved):
Small change, the "dep_fileset" argument should probably be "depfileset" to be consistent with param and other things.

PR #3817, opened 2025-07-14
return abspaths for get_* on paths

gadfort:
Changes design schema classes to return absolute paths instead just the values


aolofsson (approved):
LGTM

PR #3820, opened 2025-07-15
Reducing confusion around datadir in design API

aolofsson:
- Directory suggests path. This name change makes it clear that we are passing around a reference rather than an explicity path.


gadfort (approved):
make sense, I 'll make the schema internally consistent with this.

PR #3912, opened 2025-07-31
keep cli dashboard output the same on exit

gadfort:
@Ciprian167 FYI, this removes the end of run summary (number of passes and failed nodes) and just leaves the final render the same. I find that loosing the log output is quite frustrating. Let me know if you think something else would be better.


Ciprian167:
I was wainting to see your intention with the "final" report. But yes, that final render has to go away. It's quite annoying in multi-run jobs.

I have two questions maybe:
- Is there any use in displaying the log in multi-job runs?
- Do you usually want to check the log when everything passed?

gadfort:
> I was wainting to see your intention with the "final" report. But yes, that final render has to go away. It's quite annoying in multi-run jobs.
> 
> I have two questions maybe:
> 
> * Is there any use in displaying the log in multi-job runs?

It's something to consider removing, the logs can get somewhat confusing, but I'm not sure yet if removing it would be best (luckily it should be easy to do later)

> * Do you usually want to check the log when everything passed?

Yes, sometimes even is something is passing you might want to check if there is something in the logs to be aware of.

Ciprian167:
I'll use it like this for a while and see how it goes. Thanks for updating it.

PR #3971, opened 2025-08-08
Move dash.stop at main thread exit

Ciprian167:
@gadfort  there is a slight issue with the dashboard. Apparently the `atexit` won't work if it's not in the main thread. That would leave dashbord open, causing a mess of the terminal. Moving this into the `core.py` right after we open the dashboard solves the issue.

### DIFF
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1,6 +1,5 @@
 # Copyright 2020 Silicon Compiler Authors. All Rights Reserved.
 
-import atexit
 import copy
 import tarfile
 import os
@@ -2691,7 +2690,6 @@ class Chip:
             self._dash = WebDashboard(self, port=port, graph_chips=graph_chips)
         elif type == DashboardType.CLI:
             self._dash = CliDashboard(self)
-            atexit.register(self._dash.stop)
             wait = False
 
         self._dash.open_dashboard()
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -2691,9 +2691,9 @@ class Chip:
             self._dash = WebDashboard(self, port=port, graph_chips=graph_chips)
         elif type == DashboardType.CLI:
             self._dash = CliDashboard(self)
+            atexit.register(self._dash.stop)
             wait = False
 
-        atexit.register(self._dash.stop)
         self._dash.open_dashboard()
 
         if wait:
--- a/siliconcompiler/core.py
+++ b/siliconcompiler/core.py
@@ -1,5 +1,6 @@
 # Copyright 2020 Silicon Compiler Authors. All Rights Reserved.
 
+import atexit
 import copy
 import tarfile
 import os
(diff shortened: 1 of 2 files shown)


siliconcompiler/core.py
```diff
@@ -2690,6 +2691,7 @@ def dashboard(self, wait=True, port=None, graph_chips=None, type=DashboardType.W
             self._dash = WebDashboard(self, port=port, graph_chips=graph_chips)
         elif type == DashboardType.CLI:
             self._dash = CliDashboard(self)
+            atexit.register(self._dash.stop)
```
gadfort: could this be in the initializer for the CLIDashboard? that way it would get handled automatically and less prone to be forgotten?

gadfort (commented):
That is the issue this presents? I was making some changes in the last few days to the dashboard to try and prepare for it to be on by default (makes me wonder if this was something I accidentally caused)

siliconcompiler/report/dashboard/cli/board.py
```diff
@@ -343,8 +342,6 @@ def _init_singleton(self):
             auto_refresh=True
         )
 
-        atexit.register(self._stop_on_exit)
```
gadfort, line 346: without this, what ensures the live screen is safely terminated?

Ciprian167: That would be the atexit from the core.py

siliconcompiler/core.py
Ciprian167: Moving it into BoardSingleton seems to work fine. Thanks.

Ciprian167:
@gadfort  There is another issue with the dashboard I need to debug.

If in `post_process` the tools registers more `errors` or `warnings` using `metrics`  those will not be displayed on the dashboard. Have you noticed that?

Ciprian167:
> @gadfort There is another issue with the dashboard I need to debug.
> 
> If in `post_process` the tools registers more `errors` or `warnings` using `metrics` those will not be displayed on the dashboard. Have you noticed that?

Never mind. It was my fault. Seems to be OK.

gadfort:
> > @gadfort There is another issue with the dashboard I need to debug.
> > If in `post_process` the tools registers more `errors` or `warnings` using `metrics` those will not be displayed on the dashboard. Have you noticed that?
> 
> Never mind. It was my fault. Seems to be OK.

One issue I've seen is on the last node, it sometimes finishes and closes the dashboard before the data is updated in the board so it looks like it is still running.

siliconcompiler/report/dashboard/cli/board.py
```diff
@@ -292,6 +292,8 @@ def __call__(cls, *args, **kwargs):
             if cls not in cls._instances:
                 cls._instances[cls] = super(BoardSingleton, cls).__call__(*args, **kwargs)
                 cls._instances[cls]._init_singleton()
+                atexit.register(cls._instances[cls]._stop_on_exit)
+
```
gadfort, line 296: this seems to be functionally identical to the original code? (just moves where in the code that this is called, but its the same thread that would call it.)

Ciprian167: You're right. Now I understand what you mean by moving in the init.

PR #3976, opened 2025-08-09
move atexit to CLIDashboard and ensure stop is called on exception.

gadfort:
Possible substitute for: #3971

@Ciprian167 does this work for you. In my testing this seems to restore my terminal.


Ciprian167:
Looks good. Thanks. Seems to be working in both single-job and multi-job runs.

gadfort:
> Looks good. Thanks. Seems to be working in both single-job and multi-job runs.

Thanks for checking it out

PR #4062, opened 2025-08-29
[OpenSTA] Updated OpenSTA Tool to Work With FPGA

AlexandreSinger:
After the schema updates, the OpenSTA tool no longer worked with FPGAs. Updated the tool driver to allow FPGAs to be passed as input.


siliconcompiler/tools/opensta/__init__.py
```diff
@@ -11,6 +11,10 @@
 
 from siliconcompiler import TaskSchema
 
+from siliconcompiler import FPGASchema
+
+from siliconcompiler.schema.utils import trim
```
gadfort: trim not needed

siliconcompiler/tools/opensta/timing.py
```diff
@@ -25,9 +25,16 @@ def __init__(self):
                            defvalue="tools/_common/sdc/sc_constraints.sdc",
                            package="siliconcompiler")
 
+        self.add_parameter("device_type",
+                           "<asic,fpga>",
+                           "The type of device that the tool is targeting")
```
gadfort: I think instead of this (which is not really user settable anyways. you can overwrite:
`get_tcl_variables` and add a variable (`opensta_timing_mode` = <asic,fpga>)

siliconcompiler/tools/opensta/timing.py
```diff
@@ -25,9 +25,16 @@ def __init__(self):
...
+
     def set_timing_mode(self, mode: str, step: str = None, index: str = None):
         return self.set("var", "timing_mode", mode, step=step, index=index)
 
+    def _set_device_type(self, device_type: str, step: str = None, index: str = None):
+        return self.set("var", "device_type", device_type, step=step, index=index)
+
```
gadfort: rm

siliconcompiler/tools/vpr/__init__.py
```diff
@@ -66,6 +66,12 @@ def __init__(self):
         self.define_tool_parameter("vpr", "clock_model", "<ideal,route,dedicated_network>",
                                    "The clock modeling strategy to be used.")
 
+        self.define_tool_parameter("vpr",
```
gadfort, line 69: mark this a required in the setup method.

gadfort (changes requested):

siliconcompiler/tools/vpr/__init__.py
AlexandreSinger: I am finding this a little tricky to do. This parameter is defined in the VPRFPGA class, which does not have a setup method as far as I can tell. When I try to put it in the setup method of the VPR task, I need to know the library that the VPRFPGA is a part of in order to make the parameter required.

Is there an example of how to do this somewhere else in SC?

gadfort: Ohh, I see, is this a thing the FPGA should control? It seems like this should be a parameter for the task driver

AlexandreSinger: I believe for now this should be something that the FPGA should control. The reason is that the best lookahead someone should use is usually determined by the FPGA architecture. Most architectures would find "map" to be the best lookahead; however, in some cases architecture decisions lead to using the "classic" lookahead.

IMO the ideal scenario is that the FPGA would provide a hard-coded lookahead which has been pre-computed. So to me this is an FPGA parameter.

PR #4158, opened 2025-09-22
open filelist after collecting file content

### LINKED ISSUE #4154: Don't write file on failing write_fileset()
### Feature Description

The current implementation writes out a .f file even if one of the paths fails to resolve. It's a bit confusing to see a file there that is incomplete/empty.  It should be possible to store all fileset lines into an array in memory first and then only dump the array to a file at the end of the write_fileset() when success is guaranteed.

gadfort:
Closes #4154

### DIFF
--- a/siliconcompiler/design.py
+++ b/siliconcompiler/design.py
@@ -1,3 +1,4 @@
+import io
 import re
 
 import os.path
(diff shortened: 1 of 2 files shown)


aolofsson (approved):
nice!

PR #4168, opened 2025-09-22
expand error message for add_depfileset

### LINKED ISSUE #4166: Error reporting for add_depfileset
### Feature Description

It happens that I forget to add parenthes. ex below:

  with self.active_fileset('rtl'):
            self.add_depfileset(MemAgent)

The error message says the type is wrong, since this may be quite a common mistake it would be nice to report back that a class and not an object was found (ie. type(MemAgent) vs type(MemAgent())

           self.add_dep(dep, clobber=True)
        else:
>           raise TypeError("dep is not a valid type")
E           TypeError: dep is not a valid type

/home/aolofsson/work/zeroasic/siliconcompiler/siliconcompiler/design.py:371: TypeError

gadfort:
Closes #4166

### DIFF
--- a/siliconcompiler/design.py
+++ b/siliconcompiler/design.py
@@ -368,7 +368,7 @@ class Design(LibrarySchema, DependencySchema):
             if dep is not self:
                 self.add_dep(dep, clobber=True)
         else:
-            raise TypeError(f"dep is not a valid type: {dep} ({type(dep)})")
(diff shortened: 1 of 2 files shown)


gadfort:
I'll look around to see if there is anything else I can use, although I can also just remove the `type()` since the returned string of indicates this is a class or int.

PR #4174, opened 2025-09-23
remove load_target in favor of just calling a target directly

gadfort:
This removes the `.load_target` and moves to just calling the target method directly.  Removes a level of indirection that is not needed.

TBD: Fix the method names of the targets to not call `.setup()` (this needs to be done as a separate PR)


gadfort:
> I have a hard time getting comfortable with this. We are basically passing in the project wholesale into a function setting an arbitrary? number of settings and then continuing with the program. Can we use inheritence and get rid of the concept of "target" all together?
> 
> ex:
> 
> ```python
> if target=="sky130":
>    p = DemoSky130(design)
> elif target=="gf180":
>    p = DemoGf180(design)
> p.add_fileset(["rtl", "sdc"])
> ...
> p.run()
> ...
> ```

I don't think there is anything that would preclude someone from doing that, but it's not clear to me that it would make it easier to use. Fundamentally, the targets have always been "just a function that does stuff to the chip", encapsulating that behavior in `.use` just hid that fact a little. This PR just exposes that it's just a function.
To me, the right solution would be to remove the `load_target` stuff and then set aside some time to think through how a "target" or "some other name" is properly implemented and exposed to users.

aolofsson (approved):
Agreed, targets is "just code", so does not deserve it's own method.

PR #4175, opened 2025-09-24
Adding comments option to write_fileset()

aolofsson:
- With a large number of files, the comments doubled the size of the file, making it much harder to read.
- May want to consuder changing the polarity to "debug" at some point

### DIFF
--- a/tests/test_design.py
+++ b/tests/test_design.py
@@ -617,19 +617,14 @@ def test_write_fileset_nocomments(datadir):
         d.add_file('heartbeat_tb.v', fileset)
         d.add_define('VERILATOR', fileset)
 
-    d.write_fileset(filename="heartbeat.f", fileset=['rtl', 'tb'], comments=True)
+    d.write_fileset(filename="heartbeat.f", fileset=['rtl', 'tb'])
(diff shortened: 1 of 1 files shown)


gadfort (changes requested):
I think it would be fine to switch the polarity of this flag to default to `False`, but this requires a test to ensure it works as expected.

tests/test_design.py
```diff
@@ -585,7 +585,39 @@ def test_write_fileset(datadir):
...
+        d.add_file(['heartbeat.v', 'increment.v'], fileset)
+        d.add_define('ASIC', fileset)
+        d.add_idir('.', fileset)
+        d.set_topmodule('heartbeat', fileset)
+
+        fileset = 'tb'
+        d.add_file('heartbeat_tb.v', fileset)
+        d.add_define('VERILATOR', fileset)
+
+    d.write_fileset(filename="heartbeat.f", fileset=['rtl', 'tb'], comments=True)
```
gadfort: did you mean for this to be `comments=False`?

gadfort (approved):

PR #4336, opened 2025-10-19
Fix VCD trace path generated for #define by Verilator CompileTask (and support for FST traces)

rtrevinnoc:
When executing [heartbeat/make.py](https://github.com/siliconcompiler/siliconcompiler/blob/main/examples/heartbeat/make.py) the VCD trace file is generated inside a `reports` folder (inside the `outputs`) folder. This is also not what is expected by the line `445` in this example, which contains:
```
vcd = project.find_result(step='simulate', index='0',
                              directory="reports",
                              filename="lithography_accelerator.fst")
```
And thus GTKWave is not automatically opened.

Also, if instead of a VCD trace, a FST trace is produced by changing `project.set('tool', 'verilator', 'task', 'compile', 'var', 'trace_type', 'fst')` then there is currently no tool configured for opening such extension. Given that GTKWave is capable of doing that, I have also added said extension to the supported list.

## Summary by CodeRabbit

* **New Features**
  * Waveform viewer now supports FST files for simulation analysis in addition to existing formats
* **Improvements**
  * Updated trace file path configuration for build tools

### DIFF
--- a/siliconcompiler/tools/verilator/compile.py
+++ b/siliconcompiler/tools/verilator/compile.py
@@ -110,9 +110,9 @@ class CompileTask(VerilatorTask):
             options.append(trace_opt)
 
             # add siliconcompiler specific defines
-            c_flags.append('-DSILICONCOMPILER_TRACE_DIR="../reports"')
+            c_flags.append("-DSILICONCOMPILER_TRACE_DIR=\"reports\"")
             c_flags.append(
-                f'-DSILICONCOMPILER_TRACE_FILE="../reports/{self.design_topmodule}.{ext}"')
+                f"-DSILICONCOMPILER_TRACE_FILE=\"reports/{self.design_topmodule}.{ext}\"")
 
         if c_includes:
             c_flags.extend([f'-I{include}' for include in c_includes])
--- a/examples/heartbeat/make.py
+++ b/examples/heartbeat/make.py
@@ -10,6 +10,7 @@ from siliconcompiler.flows.dvflow import DVFlow
 from siliconcompiler.flows.fpgaflow import FPGAXilinxFlow
 
 from siliconcompiler.targets import asic_target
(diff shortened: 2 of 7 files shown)


siliconcompiler/tools/verilator/compile.py
```diff
@@ -110,9 +110,9 @@ def runtime_options(self):
             options.append(trace_opt)
 
             # add siliconcompiler specific defines
-            c_flags.append("-DSILICONCOMPILER_TRACE_DIR=\\\"reports\\\"")
+            c_flags.append('-DSILICONCOMPILER_TRACE_DIR="../reports"')
```
gadfort: this doesn't work for me, if I add `trace=True` to the verilator build and run the heartbeat example, the vcd file ends up in `build/heartbeat/job0/simulate/reports/` instead of `build/heartbeat/job0/simulate/0/reports/`

So you manually run the `vexe` file or are you simulating it like it is in the heartbeat example? if the later, I wonder why you have different behavior from me.

siliconcompiler/tools/gtkwave/show.py
```diff
@@ -5,13 +5,13 @@
...
     '''
-    Show a VCD file.
+    Show a VCD or FST file.
     '''
     def tool(self):
         return "gtkwave"
 
     def get_supported_show_extentions(self):
-        return ["vcd"]
+        return ["vcd", "fst"]
```
gadfort, line 14: thanks

siliconcompiler/tools/verilator/compile.py
rtrevinnoc: You are right! I don't know exactly what was happening, but I was using verilator v5.040 and running the simulation from the `make.py` script. When using the original params in the `c_flags` array I was getting the file produced inside the `outputs` folder, but now it works correctly.

gadfort (approved):
Lgtm. Thanks

PR #4353, opened 2025-10-24
add check_filepaths capability dependency to allow checking dependencies

gadfort:
## Summary by CodeRabbit

* **New Features**
  * Added recursive file-path validation for designs and their dependencies.

* **Bug Fixes**
  * Improved invalid-path error messages to include parameter names when available for clearer diagnostics.

* **Tests**
  * Added unit tests covering path validation across self and nested dependencies, including success, failure, and partial-failure cases.


gadfort:
@gigeresk FYI this should allow you to check the filepaths without using a Project as a go between

Issue #4433, opened 2025-11-14
Collecting files functionality is incomplete


Ciprian167:
Based on #4422 

When running remote, sc tries to collect all files used during build inside `sc_collected_files` . This functionality seems incomplete as it skips some files. There are two issues:
- Files outside of the Design object, don't seem to be collected.
- The collecting mechanism collects file only on the input node. Nodes in the middle of the graph that accept additional files will be skipped.

The issue can be reproduced with: https://github.com/Ciprian167/siliconcompiler/blob/add-verible-lint/tests/tools/test_verible.py

In this scenario `waivers` and `rules` are not collected.

Ciprian167:
Debug pointers: https://github.com/siliconcompiler/siliconcompiler/issues/4422#issuecomment-3527791123

Ciprian167:
@gadfort regarding this [comment](https://github.com/siliconcompiler/siliconcompiler/issues/4422#issuecomment-3527791123)

As far as I understand, the `client.py` works only when setting `[option, remote, True]`, which enables the `client-server` interaction. To my understanding that's a separate execution flow, compared to Slurm. As far as I can tell, it's impossible update [Client.__run_preprocess](https://github.com/siliconcompiler/siliconcompiler/blob/05021299d716f0afa5198867626fb262a58289e4/siliconcompiler/remote/client.py#L570) so that it detects the `required` keys. When running in `client-server` I think the tool `setup` is actually being run on the server by the [Scheduler.__configure_run_setup](https://github.com/siliconcompiler/siliconcompiler/blob/05021299d716f0afa5198867626fb262a58289e4/siliconcompiler/scheduler/scheduler.py#L643).

To cover `slurm, docker` and any other future remote schedulers, how do you feel about having something added here:
https://github.com/siliconcompiler/siliconcompiler/blob/05021299d716f0afa5198867626fb262a58289e4/siliconcompiler/scheduler/scheduler.py#L276

Something that checks the `[option,scheduler,name]`. If it's not local, attempt to set the copy attribute on all required files.

gadfort:
@Ciprian167 The `Client` code was just for reference. When the remote is executed, the setup is run remotely so we don't know until it's on the server if the files are needed.
I think the correct place to run the collection should be (delay until all the run checks are complete to avoid doing a bunch of copying and then needing to remove it if there is an error):
[...]

PR #4442, opened 2025-11-17
Update scheduler to set 'copy' field for all files

Ciprian167:
Targets #4433 

## Summary by CodeRabbit

* **New Features**
  * Scheduler now detects task-required files/dirs and conditions a remote collection step when needed.
  * Slurm runs mark required artifacts for copying so remote executions gather per-task files and directories.

* **Tests**
  * Added tests verifying collection occurs in Slurm mode and is skipped for local scheduler runs.

### DIFF
--- a/siliconcompiler/scheduler/scheduler.py
+++ b/siliconcompiler/scheduler/scheduler.py
@@ -29,6 +29,7 @@ from siliconcompiler.utils.logging import SCLoggerFormatter
 from siliconcompiler.utils.multiprocessing import MPManager
 from siliconcompiler.scheduler import send_messages, SCRuntimeError
 from siliconcompiler.utils.paths import collectiondir, jobdir, workdir
+from siliconcompiler.utils.curation import collect
 
 if TYPE_CHECKING:
     from siliconcompiler.project import Project
@@ -289,8 +290,9 @@ class Scheduler:
             if not self.__check_flowgraph_io():
                 raise SCRuntimeError("Flowgraph file IO constrains errors")
 
-            if not self.mark_copy():
-                raise SCRuntimeError("Failed to setup 'copy' field on files")
+            # Collect files for remote runs
+            if self.__check_collect_files():
+                collect(self.project)
 
             self.run_core()
 
@@ -973,24 +975,17 @@ class Scheduler:
 
         return not error
 
-    def mark_copy(self) -> bool:
+    def __check_collect_files(self) -> bool:
         """
-        Iterates through all tasks in the scheduler, retrieves the required
-        keys for each task, and marks those paths as "copy" in the project's
-        data structure.
+        Iterates through all tasks in the scheduler, and checks if the there
+        are files or directories that need to be collected
 
         Returns:
-            bool: True if the operation is successful, False otherwise.
+            bool: True if there is something to be collected, False otherwise.
         """
+        do_collect = False
+        for task in self.__tasks.values():
+            if task.mark_copy():
+                do_collect = True
 
-        try:
-            for _, task in self.__tasks.items():
-                keys = task.get_required_path_keys()
-
-                for key in keys:
-                    self.project.set(*key, True, field='copy')
-        except Exception as e:
-            self.__logger.error(f"Error while marking paths for copy: {e}")
-            return False
-
-        return True
+        return do_collect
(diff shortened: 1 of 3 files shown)


siliconcompiler/scheduler/schedulernode.py
```diff
@@ -1370,3 +1370,20 @@ def arcname(path):
             for logfile in self.__logs.values():
                 if os.path.isfile(logfile):
                     tar.add(logfile, arcname=arcname(logfile))
+
+    def get_required_path_keys(self) -> Set[Tuple[str, ...]]:
+        """ This functions walks thourh the require keys and returns the
+        keys that are of type path (file/dir).
+        """
+        # keys = self.__task.get('require')
```
Ciprian167: @gadfort this didn't work for some reason, and I had to reference the project. It looks like `self.__task` is a incomplete copy of the task schema. If that's ever correct to say.

gadfort: thats seems odd, if I comment it "back" will the tests you added fail in some way?

siliconcompiler/scheduler/scheduler.py
```diff
@@ -969,3 +972,25 @@ def __check_tool_versions(self) -> bool:
...
+        Returns:
+            bool: True if the operation is successful, False otherwise.
+        """
+
+        try:
+            for _, task in self.__tasks.items():
+                keys = task.get_required_path_keys()
+
+                for key in keys:
+                    self.project.set(*key, True, field='copy')
```
gadfort: probably just move this to the scheduler node too.
This is only needed in the slurm scheduler, the others can skip this for now.

siliconcompiler/scheduler/scheduler.py
```diff
@@ -969,3 +972,25 @@ def __check_tool_versions(self) -> bool:
...
+        except Exception as e:
+            self.__logger.error(f"Error while marking paths for copy: {e}")
+            return False
```
gadfort: what errors might occur?

siliconcompiler/scheduler/schedulernode.py
gadfort: I see. In the task driver to keep the step and index use to a minimum, they are auto inferred.
However, this only works in particular contexts (during setup, pre_processing, post_processing, run), outside of these it is needed:
```
keys = self.__task.get('require', step=self.step, index=self.index)
```

Ciprian167: Understood. Thanks

Ciprian167:
@gadfort I missed understood your original suggestion, hence the confusing code. Sorry. Maybe this it closer to what you had in mind.

siliconcompiler/scheduler/slurm.py
```diff
@@ -161,6 +142,13 @@ def assert_slurm() -> None:
         if shutil.which('sinfo') is None:
             raise RuntimeError('slurm is not available or installed on this machine')
 
+    def mark_copy(self) -> bool:
+        do_collect = False
```
Ciprian167, line 146: @gadfort by default we mark everything as copy. In more complex graphs, with nodes that take inputs from other nodes, do we need to mark those files as `copy`, or should we ignore them since we assume they will be already in the build directory. 

And also the collect now is done one time, after the setup, right before the run. At that point most of the output will not exist, which makes me think this strategy will not work for files handed between flowgraph nodes, and will probably fail.

Ciprian167: Nevermind, the `input` files are separate from `require`. All good.

siliconcompiler/scheduler/scheduler.py
Ciprian167: I'm guessing none. I got confused. :disappointed:

gadfort: No worries, I was just wondering if you had encountered something

gadfort (approved):
Looks good to me, I'll need to verify a few things on my end to ensure the remote runner "okay" with this change. Afterwards I'll merge it.

gadfort:
@Ciprian167 The remote runner will end up calling collect again, which deletes the files that were uploaded, I'll need to fix that before merging.
Fixed now. Remote runner is passing its tests

PR #4459, opened 2025-11-18
Add drain period for manifest to propagate through nfs

### LINKED ISSUE #4423: Reporting looks wrong when enabling slurm scheduler
Found on: https://github.com/siliconcompiler/siliconcompiler/commit/51e1b59e153bb671b172ec077d5b7c56982a69d5

It looks like something is not complete in the slurm scheduler. I have a simple verible lint task that's passing on slurm, but the reporting shows it failed.

```
| INFO     | Checking manifest before running.
| INFO     | Setting up node verible/0 with verible/lint
| ERROR    | Run failed: Could not run final steps: verible
| ERROR    | Job log: build/custom_design/custom_design_lint/job.log
```
I compared `SlurmSchedulerNode.run` and the `SchedulerNode.execute`. I noticed that the slurm scheduler doesn't set the record, status properly. I tried to force it to SUCCESS, but still no luck.

Has the slurm scheduler been used? If not I can carry it over, and I could appreciate some pointers.

Ciprian167:
Closes #4423 

@gadfort This should wait for 2 seconds max, which should cover a really really slow network. In best case it won't wait at all. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Slurm scheduler robustness: when job completion metadata is briefly delayed on shared storage, the scheduler now waits and retries for a short window with brief pauses before reporting an error, reducing spurious failures from transient filesystem latency.

### DIFF
--- a/siliconcompiler/scheduler/slurm.py
+++ b/siliconcompiler/scheduler/slurm.py
@@ -244,7 +244,7 @@ class SlurmSchedulerNode(SchedulerNode):
             if os.path.exists(log_file):
                 self.logger.error(f"Node log file: {log_file}")
             self.halt()
-        
+
         # Wait for manifest to propagate through network filesystem
         start = time.time()
         elapsed = 0


siliconcompiler/scheduler/slurm.py
```diff
@@ -240,3 +241,13 @@ def run(self):
             if os.path.exists(log_file):
                 self.logger.error(f"Node log file: {log_file}")
             self.halt()
+        
+        # Wait for manifest to propagate through network filesystem
+        start = time.time()
+        elapsed = 0
+        while not os.path.exists(self.get_manifest()) and elapsed <= 2:
+            elapsed = time.time() - start
+            time.sleep(0.5)
```
gadfort: You will need to call: listdir or scandir to trigger the NFS cache from getting updated (this is what we do in several places (including right above this):
https://github.com/Ciprian167/siliconcompiler/blob/e8053f7feec6287a28ad4c4ceb81fe4c839c2cb2/siliconcompiler/scheduler/slurm.py#L224

Could you move the magic number to class const members like `_MAX_FS_DELAY = 2` and `_FS_DWELL = 0.5` you can probably lower the sleep to 0.1 or 0.2 since this would only cause 10 to 20 IO calls.

siliconcompiler/scheduler/slurm.py
```diff
@@ -240,3 +241,13 @@ def run(self):
...
+        if not os.path.exists(self.get_manifest()):
+            self.logger.error(f"Manifest was not created on time: {self.get_manifest()}")
```
gadfort: Is calling halt appropriate? The task scheduler will fail this and halt will trigger a write of the manifest potentially overwriting the one remotely.

siliconcompiler/scheduler/slurm.py
Ciprian167: OK. I know you mentioned calling `listdir` before, but didn't understood the reason. Thanks for clarifying.

siliconcompiler/scheduler/slurm.py
Ciprian167: I didn't think too much about the halt implications. I wanted to have an early error indication. But I guess the error is sufficient.

Ciprian167:
@gadfort  I'm seeing another issue related to Slurm, but not necessarily related to this change. I have a job with multiple indexes. All indexes seems to be passing fine on slurm, confirmed by the manifest and the slurm dashboard, but could be that something goes wrong with the manifest replay.   

Have you seen something with this behavior?

gadfort:
@Ciprian167 
1. you have a conflict to resolve.
2. What is the issue you are seeing. Are they getting marked as failed when they finished on slurm? Maybe raise that as a separate issue if appropriate

Ciprian167:
I see something strange going on only for `steps` with multiple `indices`. I will debug some more and open a separate ticket.

gadfort (approved):
LGTM

gadfort:
> I see something strange going on only for `steps` with multiple `indices`. I will debug some more and open a separate ticket.

Thanks

Ciprian167:
@gadfort I think there was something wrong with our infrastructure. Even the 2 s timeout was not enough. I did some testing and it took around 60s for the cache to be updated. I suspect we were using on older version of NFS, or it was configured wrong. But they updated the infra, and I don't see this behavior anymore, not even for `steps` with multiple `indices`.

gadfort:
@Ciprian167 we had to disable caching on the NFS. 60s is a long time, but good to hear its working again. (If at any point you want to take a look at our slurm setup documentation and suggest improvements, I'd be happy to incorporate them)

PR #4467, opened 2025-11-19
move scheduler initialization

gadfort:
## Summary by CodeRabbit

* **Refactor**
  * Consolidated and deduplicated scheduler initialization so all scheduler init callbacks run earlier in the run workflow.
  * Removed prior per-node preprocessing step; per-node init callbacks are no longer invoked before launching tasks.


gadfort:
@coderabbitai review

PR #4533, opened 2025-12-09
Use [scheduler,options] in slurm

Ciprian167:
@gadfort I noticed `[option, scheduler, options]` were omitted for slurm. This should take care of it. 

## Summary by CodeRabbit

* **New Features**
  * Slurm scheduler now forwards user-defined scheduler options to srun, allowing additional Slurm flags/settings when launching jobs.
  * Queue/partition selection now respects configured scheduler queue values so jobs use the intended partition when set.
  * Deferred start times are now applied to job launches (adds a scheduled/begin time when configured).

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/siliconcompiler/scheduler/slurm.py
+++ b/siliconcompiler/scheduler/slurm.py
@@ -209,8 +209,7 @@ class SlurmSchedulerNode(SchedulerNode):
         self._init_run_logger()
 
         # Determine which cluster parititon to use.
-        partition = self.project.get('option', 'scheduler', 'queue',
-                                     step=self.step, index=self.index)
+        partition = self.project.option.scheduler.get_queue(step=self.step, index=self.index)
         if not partition:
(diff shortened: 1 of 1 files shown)


siliconcompiler/scheduler/slurm.py
```diff
@@ -259,6 +259,13 @@ def run(self):
         if defer_time:
             schedule_cmd.extend(['--begin', defer_time])
 
+        # Forward additional options
+        options = self.project.get('option', 'scheduler', 'options',
+                                      step=self.step, index=self.index)
```
gadfort: I think this should be accessed via:
```
self.project.option.scheduler.get_options(step=...)
```

Ciprian167: Make sense. Still getting the hang of it.

PR #4546, opened 2025-12-15
Remove exclusive option from slurm

Ciprian167:
@gadfort can we remove `--exclusive` from the default slurm arguments? 

This can create some issues for certain setups. I'm no expert in HPC or slurm, but I think there are two different setups. 

1. Many nodes with few resources 
2. Few nodes with many resources

I think the `--exclusive` make a lot of sense in the context of the first setup. But for the second one, it only causes issues and leads to waste of resources. I guess users can use the `scheduler, options` to set something as exclusive. 

Is it possible to remove it?

## Summary by CodeRabbit

* **Changes**
  * Updated job scheduling behavior to allow shared resource allocation instead of exclusive resource allocation, enabling more efficient utilization of compute resources.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/siliconcompiler/scheduler/slurm.py
+++ b/siliconcompiler/scheduler/slurm.py
@@ -245,7 +245,6 @@ class SlurmSchedulerNode(SchedulerNode):
                  os.stat(script_file).st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
 
         schedule_cmd = ['srun',
-                        '--exclusive',
                         '--partition', partition,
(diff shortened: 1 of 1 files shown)


gadfort:
@Ciprian167 makes sense that it should be handled by the options flag, I will just need to trigger a check on an internal thing to verify I don't need to make some fixes before merging.

gadfort (approved):
@Ciprian167 thanks.

Ciprian167:
Thank you.

PR #4911, opened 2026-04-22
Change filelist generation to be depth first

HU90m:
Tools which care about file order, often want to have parsed dependent files first.

picorv32 slang lint is now failing due to `MissingTimeScale` error, due to the first modules (`la_spram_impl` and `la_spram`) not having timescale set but later ones having it set. I wasn't sure of the best way to go about fixing this. I couldn't find a flag to get slang to ignore timescales.

*Worked out a fix for the picorv32 slang lint issue. Popped it on top as a second commit. Happy to squash if it looks good.*

The existing tests check for file order well enough to find regressions. Would you like the [ALU example](https://github.com/HU90m/siliconcompiler/tree/alu-example/examples/alu) as part of this PR regardless?

## Summary by CodeRabbit

* **Refactor**
  * Improved fileset resolution for consistent ordering, de-duplication, and cycle prevention.
  * Alias handling expanded to accept single or multiple targets and now enforces runtime validation.
  * Public APIs updated to accept the enhanced alias format.

* **Tests**
  * Updated unit tests and golden expectations to match corrected fileset ordering and alias behavior.

* **Examples**
  * Example project updated to pull dependent filesets transitively for memory/top-module wiring.


gadfort:
@HU90m The fix to the lint example is fine (and more correct).
At first glance it looks good (a little alarmed that more things didn't break, but okay maybe thats a good thing).
I'll need to run some checks internally to make sure this doesn't cause any issues, if we find something it might delay when I can merge this since we are in the middle of a tapeout.

gadfort:
@HU90m I haven't forgotten this, I still have not had a chance to verify this against some of the internal setups we have.

HU90m:
Thanks for keeping me in the loop. Looking forward to hearing how it goes

gadfort (approved):

PR #5000, opened 2026-06-05
Add support for verilog testbenches with Verilator

### LINKED ISSUE #3632: Allow (System)Verilog testbenches with Verilator
### Feature Description

For quite some time now, Verilator has supported V/SV testbenches rather than exclusively C/C++/SystemC TBs. Allow siliconcompiler to accept SV testbenches as inputs to Verilator.

I think this should be a pretty simple fix, perhaps requiring only a small change to the Verilator tool scripts. Will look into it myself and see if I can provide a patch.

infinitymdm:
Fixes #3632. Enables the use of verilog testbenches with verilator.

## Summary by CodeRabbit

* **New Features**
  * Simulation: choose testbench type (C++ or Verilog) when running sims; numeric parameter overrides apply to the selected testbench. Verilator runs enable tracing for all testbench types and adjust main-wrapper/timing behavior based on testbench selection.
* **Tests**
  * Verilator simulation coverage split into separate tests for Verilog and C++ testbenches.

### DIFF
--- a/siliconcompiler/tools/verilator/compile.py
+++ b/siliconcompiler/tools/verilator/compile.py
@@ -183,6 +183,8 @@ class CompileTask(VerilatorTask):
                 self.add_required_key(lib, "fileset", fileset, "file", "c")
                 added_key = True
         if not added_key and not self.get("var", "main"):
+            self.logger.warning("No c files found. c files are required for verilator compilation "
+                                "unless the \"main\" option is enabled.")
             self.add_required_key(self.project.design, "fileset",
                                   self.project.get("option", "fileset")[0], "file", "c")
 
--- a/siliconcompiler/tools/verilator/compile.py
+++ b/siliconcompiler/tools/verilator/compile.py
@@ -254,9 +254,15 @@ class CompileTask(VerilatorTask):
             options.append(trace_opt)
 
             # add siliconcompiler specific defines
-            c_flags.append("-DSILICONCOMPILER_TRACE_DIR=\"reports\"")
-            c_flags.append(
-                f"-DSILICONCOMPILER_TRACE_FILE=\"reports/{self.design_topmodule}.{ext}\"")
+            if self.get("var", "main"):
+                options.extend([
+                    "-DSILICONCOMPILER_TRACE_DIR=\"reports\"",
+                    f"-DSILICONCOMPILER_TRACE_FILE=\"reports/{self.design_topmodule}.{ext}\""
+                ])
+            else:
+                c_flags.append("-DSILICONCOMPILER_TRACE_DIR=\"reports\"")
+                c_flags.append(
+                    f"-DSILICONCOMPILER_TRACE_FILE=\"reports/{self.design_topmodule}.{ext}\"")
 
         if c_includes:
             c_flags.extend([f'-I{include}' for include in c_includes])
--- a/siliconcompiler/tools/verilator/compile.py
+++ b/siliconcompiler/tools/verilator/compile.py
@@ -19,6 +19,10 @@ class CompileTask(VerilatorTask):
         self.add_parameter("trace_type", "<vcd,fst>",
                            "specifies type of wave file to create when [trace] is set",
                            defvalue="vcd")
+        self.add_parameter("main", "bool",
+                           "if true, generate a toplevel C++ wrapper and relax the C++ file "
+                           "requirement. See --main in Verilator docs for more info",
+                           defvalue=False)
 
         # TODO Move to design object
         self.add_parameter("cincludes", "[dir]",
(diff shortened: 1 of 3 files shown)


siliconcompiler/tools/verilator/compile.py
```diff
@@ -234,9 +254,15 @@ def runtime_options(self):
...
-                f"-DSILICONCOMPILER_TRACE_FILE=\"reports/{self.design_topmodule}.{ext}\"")
+            if self.get("var", "main"):
+                options.extend([
+                    "-DSILICONCOMPILER_TRACE_DIR=\"reports\"",
+                    f"-DSILICONCOMPILER_TRACE_FILE=\"reports/{self.design_topmodule}.{ext}\""
+                ])
+            else:
+                c_flags.append("-DSILICONCOMPILER_TRACE_DIR=\"reports\"")
+                c_flags.append(
+                    f"-DSILICONCOMPILER_TRACE_FILE=\"reports/{self.design_topmodule}.{ext}\"")
```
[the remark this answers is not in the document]
infinitymdm: I'm not confident this is actually an issue (though the code could probably be cleaned up a bit by deduplicating the flags to a single list stored in a variable). In my testing with the heartbeat example, a wavedump is generated when run with `tool='verilator', tb_type='v'`. 

It wouldn't hurt anything to pass these as CFLAGs in addition to HDL preprocessor defines, but with a pure-verilog testbench they won't be used anywhere. And `--main` should only be set when you have no cc sources anyways. Is there any way this could actually cause a problem?

tests/examples/test_heartbeat.py
```diff
@@ -67,10 +67,21 @@ def test_py_make_sim_icarus():
...
+    assert os.path.isfile('build/heartbeat/job0/heartbeat.pkg.json')
+    assert os.path.isfile('build/heartbeat/job0/simulate/0/reports/heartbeat_tb.vcd')
+
+
+@pytest.mark.eda
+@pytest.mark.quick
+@pytest.mark.timeout(300)
+def test_py_make_sim_verilator_cc():
+    from heartbeat import make
+    make.sim(tb_type='cc')
```
gadfort, line 83: please add some kind of assertion
`assert os.path.isfile('build/heartbeat/job0/heartbeat.pkg.json')`

siliconcompiler/tools/verilator/compile.py
```diff
@@ -165,7 +182,7 @@ def _setup_c_file_requirement(self):
             if lib.has_file(fileset=fileset, filetype="c"):
                 self.add_required_key(lib, "fileset", fileset, "file", "c")
                 added_key = True
-        if not added_key:
+        if not added_key and not self.get("var", "main"):
             self.add_required_key(self.project.design, "fileset",
                                   self.project.get("option", "fileset")[0], "file", "c")
```
gadfort, line 189: would it be possible to say, if the c fileset is not added add add main automatically?

siliconcompiler/tools/verilator/compile.py
```diff
@@ -199,6 +216,9 @@ def runtime_options(self):
         if self.get("var", "initialize_random"):
             options.extend(['--x-assign', 'unique'])
 
+        if self.get("var", "main"):
```
gadfort, line 222: make main a required key

siliconcompiler/tools/verilator/compile.py
```diff
@@ -199,6 +216,9 @@ def runtime_options(self):
...
+            options.extend(['--main', '--timing'])
```
gadfort, line 223: why is timing needed?

gadfort (changes requested):
Overall looks good, just a few comments and need something to assert in the test

siliconcompiler/tools/verilator/compile.py
infinitymdm: `--timing` enables support for verilog timing directives, such as `#2` delays etc. 

EDIT: I don't know why this isn't enabled by default, but that's a question for verilator.

tests/examples/test_heartbeat.py
infinitymdm: The assertions are already present on L85-L86.

siliconcompiler/tools/verilator/compile.py
infinitymdm: I considered doing this, but thought it would be more helpful to have an error message pop up if main isn't explicitly selected. That way if users *are* trying to use c testbenches and they forget to add them to the fileset, they still get the error that's most helpful for their usecase.

tests/examples/test_heartbeat.py
gadfort: I see, I was wondering why you had added some to the earlier test, but that was just my missing it. Thanks

siliconcompiler/tools/verilator/compile.py
infinitymdm: Resolved in the latest commit.

siliconcompiler/tools/verilator/compile.py
gadfort: good point.
Maybe we can add a warning `self.logger.warning()` to indicate that either main is needed or the c files? Mainly to help highlight the `set_verilator_main` option.

infinitymdm: I like that idea. Added in the latest commit. Feel free to critique the wording.

gadfort (approved):

PR #5020, opened 2026-06-10
add tests for asic signoff to ensure it works

gadfort:
@infinitymdm looks like gf180 was not running and is currently generating some DRCs, this will serve to ensure we have testing and update lambdapdk to include the missing report flag.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ dependencies = [
     "PyGithub >= 2.8.0, < 2.10.0",
     "urllib3 >= 1.26.0", # Required for PyGithub
 
-    "lambdapdk >= 0.2.12",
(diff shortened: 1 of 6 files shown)


infinitymdm:
Just FYI, fossi-foundation has been working on open-source LVS for gf180mcu. You can find the details over here: https://github.com/fossi-foundation/globalfoundries-pdk-libs-gf180mcu_fd_pv. They have updated klayout DRC decks and LVS using magic/netgen (which is why I though the signoffflow would work in the first place).

It might be worthwhile to update lambdapdk to point to the fossi-foundation versions of the PDK as well. AFAIK that's the authoritative source anyways.

gadfort:
@infinitymdm thats probably a good idea.

PR #5043, opened 2026-06-28
schematic: Fix reversed and dropped bus ranges in `write_verilog`

jamestiotio:
Bit ranges are stored as (max, min) tuples, but `write_verilog` emitted them reversed as "[min:max]" and guarded the range with `if bits[1]:`. That truthiness check treated any bus whose low bit is 0 as a scalar, silently dropping the width and producing structurally wrong Verilog for every vector port and net.

Reconstruct the range through a single `_format_bitrange()` helper that returns "[max:min]" for vectors and "" for the (0, 0) scalar sentinel. `get_partpins()` already rebuilt ranges correctly, so route it through the same helper to keep the two paths from diverging again.

Add `test_write_verilog_buses` with a golden netlist covering both the "[N:0]" and non-zero low bit cases for pins and nets.

## Summary by CodeRabbit

* **New Features**
  * Improved Verilog output for buses and bit ranges, with consistent formatting for scalar and vector pins and nets.
  * Added support for generating and validating Verilog for designs that include mixed-width interface signals.

* **Tests**
  * Added coverage for bus-style Verilog generation to help ensure the emitted hardware description matches expected output.

### DIFF
--- a/siliconcompiler/schematic.py
+++ b/siliconcompiler/schematic.py
@@ -267,9 +267,7 @@ class Schematic(BaseSchema):
         pinlist = []
         for item in keys:
             t = self.get('schematic', 'part', part.name, 'pin', item, 'bitrange')
-            if t != (0, 0):
-                item = f"{item}[{t[0]}:{t[1]}]"
-            pinlist.append(item)
+            pinlist.append(f"{item}{self._format_bitrange(t)}")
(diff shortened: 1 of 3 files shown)


gadfort (approved):
Thanks. Since this code is not really used at the moment, I'm curious what your use model is?

PR #5054, opened 2026-07-01
switch global to *, to avoid name conflicts

gadfort:
Change global keyword to * to avoid name conflicts with common terms like global.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved backward compatibility when loading older manifests by correctly interpreting legacy global keys.
  - Standardized how global/default node values are serialized and recognized using wildcard keys.

- **Documentation**
  - Updated the reference guide for Manifest Tab filtering behavior and clarified schema glossary wording.

- **Tests**
  - Added version-specific deserialization coverage and expanded round-trip/compatibility assertions.
  - Updated expected schema outputs and reserved-name checks to match the wildcard global sentinel.


aolofsson (approved):
I like it!

PR #5075, opened 2026-07-06
move dashboard to optional section

gadfort:
Based on recent experience, it seems like we can generate a new dashboard fairly cleanly with claude or some other tool. Since the web-dashboard brings in a relatively large number of dependencies and is probably not very heavily used, this moves it to an optional install with the plan being to deprecate in the future.

## Summary

* **New Features**
  * Dashboard-related packages are now installed via a dedicated optional `dashboard` extra, so they’re only pulled in when required.
  * Dashboard startup now provides clearer feedback when the needed dashboard dependency is missing.

* **Documentation**
  * Updated the Dashboard how-to guide and tutorial to instruct installing dashboard dependencies (`pip install siliconcompiler[dashboard]`) before running `sc-dashboard`.

* **Tests / Chores**
  * Updated CI and hosted documentation builds to install with the `dashboard` extra.

### DIFF
--- a/.github/workflows/daily_ci.yml
+++ b/.github/workflows/daily_ci.yml
@@ -108,7 +108,7 @@ jobs:
 
           python3 -m pip install --upgrade pip
 
-          python3 -m pip install .[test,cocotb]
+          python3 -m pip install .[test,cocotb,dashboard]
           python3 -m pip install -r ./examples/requirements.txt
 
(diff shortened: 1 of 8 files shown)


gadfort:
@coderabbitai review

aolofsson (approved):
reasonable

PR #5118, opened 2026-07-14
Fix deprecated OpenROAD RCX commands

infinitymdm:
Previously RC extraction with OpenROAD would generate a number of warnings about the deprecated `-ext_model_file` arg on the `extract_parasitics` command. This PR updates the tcl scripts to use the `set_extraction_rules_file` command instead. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated per-corner parasitic (PEX) extraction used during SPEF generation to apply extraction rules correctly for newer tool versions, while preserving the prior flow for older versions.
  * Improves consistency of SPEF output across supported extraction corners.

* **Tests**
  * Updated the GCD test to expect the current warning count for the `write.views` step.

### DIFF
--- a/siliconcompiler/tools/openroad/scripts/apr/sc_write_data.tcl
+++ b/siliconcompiler/tools/openroad/scripts/apr/sc_write_data.tcl
@@ -69,8 +69,12 @@ if { [sc_cfg_tool_task_get var write_spef] } {
     foreach pexcorner [sc_cfg_tool_task_get var pex_corners] {
         set pex_model [lindex [sc_cfg_get_fileset $sc_pdk $pexfileset openrcx] 0]
         puts "Writing SPEF for $pexcorner"
-        set_extraction_rules_file $pex_model
-        extract_parasitics
+        if { [sc_check_version 26 3 23] } {
+            set_extraction_rules_file $pex_model
+            extract_parasitics
+        } else {
+            extract_parasitics -ext_model_file $pex_model
+        }
         write_spef "outputs/${sc_topmodule}.${pexcorner}.spef"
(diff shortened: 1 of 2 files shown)


siliconcompiler/tools/openroad/scripts/apr/sc_write_data.tcl
```diff
@@ -69,7 +69,8 @@ if { [sc_cfg_tool_task_get var write_spef] } {
     foreach pexcorner [sc_cfg_tool_task_get var pex_corners] {
         set pex_model [lindex [sc_cfg_get_fileset $sc_pdk $pexfileset openrcx] 0]
         puts "Writing SPEF for $pexcorner"
-        extract_parasitics -ext_model_file $pex_model
+        set_extraction_rules_file $pex_model
+        extract_parasitics
```
gadfort: this should be to continue to support older openroad versions.
Looks like this was added in:
```
git describe 
26Q3-23-gb65c274cad
```

```
if { [sc_check_version 26 3 23] } {
    set_extraction_rules_file $pex_model
    extract_parasitics
} else {
     extract_parasitics -ext_model_file $pex_model
}
```

gadfort (changes requested):

gadfort:
@infinitymdm I should also mention that OpenROAD is changing the OpneRCX format to be a single file for all corners instead of separate files. If you are using lambdapdk that will be updated for you, but if you have your own enablement I made some helper scripts to assist in the transition when things are ready.

gadfort (approved):
Thanks for fixing this

PR #5159, opened 2026-07-26
Add support for verilator's `--trace-structs` and `--trace-depth`.

HU90m:
An decision to take note of: I've enabled trace_structs by default. I think this is a good default, so new users don't have to struggle before finding the option.

## Summary by CodeRabbit

* **New Features**
  * Added Verilator trace configuration to optionally include signal structure details (`trace_structs`, enabled by default).
  * Added support for setting Verilator trace depth (`trace_depth`) to limit tracing detail.
  * Existing trace waveform options remain unchanged.
* **Tests**
  * Updated Verilator runtime-argument tests to include the new `--trace-structs` flag.
  * Added regression coverage for FST tracing with `trace_structs` disabled and `trace_depth` set.

### DIFF
--- a/siliconcompiler/tools/verilator/compile.py
+++ b/siliconcompiler/tools/verilator/compile.py
@@ -19,6 +19,12 @@ class CompileTask(VerilatorTask):
         self.add_parameter("trace_type", "<vcd,fst>",
                            "specifies type of wave file to create when [trace] is set",
                            defvalue="vcd")
+        self.add_parameter("trace_structs", "bool",
+                           "Enable tracing structure names",
+                           defvalue=True)
+        self.add_parameter("trace_depth", "int",
+                           "Specify the number of levels deep to enable tracing")
+
         self.add_parameter("main", "bool",
                            "if true, generate a toplevel C++ wrapper and relax the C++ file "
                            "requirement. See --main in Verilator docs for more info",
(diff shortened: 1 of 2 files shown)


gadfort:
@HU90m what does this output / provide (I only use verilator superficially)? what (if any) impact does this have on runtime / compile time? is it negligible?

gadfort (changes requested):
Please add a setter for the new parameters:
```
def set_verilator_trace<something>(self, enable: bool, depth: optional[int] = None, step, index)...
```
or two separate ones if that makes better sense. You can see `set_verilator_mode` as an example.
Would be nice to add tests for the new methods and a runtime_options test that includes the depth.

HU90m:
> @HU90m what does this output / provide (I only use verilator superficially)? what (if any) impact does this have on runtime / compile time? is it negligible?

Yes, I believe it does. Verilator can not flatten structs etc. because it need to trace struct members which has a runtime hit. There's also more C++ generated, so there's a compile time hit two (not sure how significant). The trace files are are larger as a result, so more I/O.

However, there's a pretty big performance hit when just enabling tracing, and I'd imagine most people would prefer a more readable trace over a smaller performance hit due to tracing. It's one of those switches people wish they knew about earlier.

If someone does want faster traced simulations, `--trace-structs` isn't the only thing they'd likely fiddle with. They should probably also be looking at settings like `--trace-depth`, `--trace-max-array`, `--trace-max-width`, `--trace-threads`, etc.

^ This was my thinking in making it default, but I'm not wedded to the decision, happy to change

gadfort:
@HU90m  this only applies when compiling with trace support? correct?

HU90m:
Yes, these new `trace_structs` and `trace_depth` parameters are ignored when the trace parameter isn't set. I'm also pretty sure the arguments would have no effect anyway when tracing isn't enabled.

gadfort (approved):
LGTM, Thanks for adding this.

PR #5254, opened 2026-08-15
RTL is a clocked flow of data

### LINKED ISSUE #5209: Clocked RTL
### Feature Description

On page: https://docs.siliconcompiler.com/en/latest/user_guide/glossary.html

May I propose that the definition of ``RTL`` be changed ``from:``

Register Transfer Level – a description of hardware in terms of **data flow** between registers, written in a language such as Verilog, VHDL, or SystemVerilog. The usual starting point of a compilation.

``to:``

Register Transfer Level – a description of hardware in terms of **a clocked flow of data** between registers, written in a language such as Verilog, VHDL, or SystemVerilog. The usual starting point of a compilation.

It is true that Verilog, VHDL, and SystemVerilog encompass more than ``RTL`` and therefore also allow descriptions of unclocked hardware. However, this is a property of those languages rather than of ``RTL`` itself.

From my own perspective as a new user, correctly understanding the significance of clocking in RTL has been crucial to making progress with SiliconCompiler. The helpful glossary has provided a useful way of checking my understanding of this and other important concepts in ASIC design. I am sure other new users will derive similar benefits from reading the glossary.

May I therefore recommend that the glossary express key concepts such as RTL as simply and accurately as possible? I believe this minor change would be a useful step in that direction.

philiprbrenan:
Clarify the definition of RTL by specifying 'a clocked flow of data'.

## Summary by CodeRabbit

* **Documentation**
  * Updated the RTL glossary entry to clarify that RTL describes clocked data flow between registers.
  * Preserved the existing language examples and supporting explanation.

### DIFF
--- a/docs/user_guide/glossary.rst
+++ b/docs/user_guide/glossary.rst
@@ -249,9 +249,9 @@ output.
        for real interconnect delay rather than estimates.
 
     RTL
-       Register Transfer Level -- a description of hardware in terms of data
-       flow between registers, written in a language such as Verilog, VHDL, or
-       SystemVerilog. The usual starting point of a compilation.
+       Register Transfer Level -- a description of hardware in terms of a
+       clocked flow of data between registers, written in a language such as 
+       Verilog, VHDL, or SystemVerilog. The usual starting point of a compilation.
 
     SDC
        Synopsys Design Constraints -- the de facto standard format for


gadfort (approved):
LGTM. Thanks

PR #5299, opened 2026-08-25
add LECFlow, an RTL-to-gate equivalence check using kepler-formal

jyrj:
Adds LECFlow, which synthesizes an RTL design with Yosys and has Kepler-formal check the resulting netlist back against the elaborated RTL, plus a combinational adder example. 

## Summary by CodeRabbit

* **New Features**
  * Added RTL-to-netlist and sequential equivalence checking through Kepler-Formal.
  * Added support for Kepler, Kepler-SEC, and Yosys equivalence-checking workflows.
  * Added an adder equivalence-checking example.
* **Documentation**
  * Expanded formal verification guidance and documented supported equivalence-checking tasks.
* **Tests**
  * Added coverage for successful and failing equivalence checks, including the example workflow.
  * Added validation that equivalence results report detected design differences.


siliconcompiler/flows/formalflow.py
```diff
@@ -75,7 +80,66 @@ def make_docs(cls):
...
+        elab = ElaborationFlow(language=language)
+        self.graph(elab)
+
+        elab_node = elab.get_exit_nodes()
+        if len(elab_node) != 1:
+            raise ValueError("Elaboration flow must have exactly one exit node.")
+        elab_node = elab_node[0][0]  # Get the node name from the tuple
+
+        self.node("synthesis", syn_asic.ASICSynthesis())
+        self.edge(elab_node, "synthesis")
```
gadfort: the lec flow doensn't need synthesis, it just needs the LEC task

siliconcompiler/tools/keplerformal/sec.py
```diff
@@ -0,0 +1,117 @@
...
+                             "library corners.")
+        scenario = list(scenarios.values())[0]
+        libcorners = scenario.get_libcorner(self.step, self.index)
+        delay_model = self.project.get("asic", "delaymodel")
+        for asiclib in self.project.get("asic", "asiclib"):
+            lib = self.project.get_library(asiclib)
+            for corner in libcorners:
+                if not lib.valid("asic", "libcornerfileset", corner, delay_model):
+                    continue
+                yield lib, corner, delay_model
```
gadfort: we normally do not use `yield` , you can just return a list

siliconcompiler/tools/keplerformal/sec.py
```diff
@@ -0,0 +1,117 @@
...
+    def setup(self) -> None:
+        super().setup()
+
+        self.set_exe("kepler-formal")
+
+        # Log file handled by kepler-formal config
+        self.set_logdestination("stdout", "none")
+        self.set_logdestination("stderr", "none")
+
+        self.add_input_file(self.__rtl_file())
```
gadfort, line 26: does kepler support idirs, params, defines?

gadfort (changes requested):

siliconcompiler/tools/keplerformal/sec.py
jyrj: Done. Only two true generator functions exist in the package, the rest of the yields are @contextmanager. and mine was the only one under tools/. Now returns List[Tuple[Design, str, str]], matching the other -> List[...] helpers there.

siliconcompiler/tools/keplerformal/sec.py
jyrj: Kepler-formal has no include-dir, define or parameter flags of its own, I went through the CLI and YAML surface at the pinned commit and confirmed by running it. It doesn't need them as the flow stands: elaborate expands macros and includes and rewrites parameter defaults, so the RTL reaching lec is self-contained. Feeding raw RTL instead fails with error: unknown macro or compiler directive '\ADDER_WIDTH'`.

jyrj:
Dropping synthesis, agreed.. no other check flow builds the thing it checks. With nothing making the netlist now, should the task point at its fileset with a var like opensta's power_activities, or just read one named netlist like the heartbeat gate-level sim does? @gadfort

PR #5374, opened 2026-09-10
Allow maxnodes to take effect for remote schedulers

### LINKED ISSUE #5363: option.scheduler.maxnodes not considered for slurm
Commit: 992b3abbb1fcfdc774c76a537f0310f825b9bc09

Based on the description of `options.scheduler.maxnodes` I can understand that if `maxnodes` is set the job execution will be constrained to that number of jobs. However for remote schedulers that's being bypassed. 

The logic can be traced in `TaskScheduler.__allow_start`. 

```python

    def __allow_start(self, node: Tuple[str, str]) -> bool:
        # ...
        if not info["node"].is_local:
            # using a different scheduler, so allow
            return True
        # ...

```

Is there a reason why that's being done? As far as I understand, for slurm jobs are still being dispatched one by one using `srun`.  So it will dispatch everything that can run at one time slurm. 

I know that slurm's jobs is to schedule execution, but one thing that slurm doesn't schedule/manage is the EDA license usage. That's usually done through FlexLM servers. If those are configured poorly, one user can consume all licenses. So it would be good to have a way to throttle the job submission even for remote schedulers. 

Would it be possible to make remote schedulers react to `maxnodes`?

Thanks,
Ciprian

ciprianantoci-collab:
Target #5363 

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized scheduling behavior so all nodes follow local running-node, parallelism, and core-availability limits.
  * Prevented non-local scheduling requests from bypassing local resource checks.

### DIFF
--- a/siliconcompiler/scheduler/taskscheduler.py
+++ b/siliconcompiler/scheduler/taskscheduler.py
@@ -641,10 +641,6 @@ class TaskScheduler:
         """
         info = self.__nodes[node]
 
-        if not info["node"].is_local:
-            # using a different scheduler, so allow
-            return True
-
         running_nodes = self.get_running_nodes()
 
         if len(running_nodes) >= self.__max_parallel_run:


gadfort (approved):
LGTM

